Issue #3, opened 2018-03-21
Replica Exchange Monte Carlo


YoshikawaMasashi:
Hi.

I implemented Replica Exchange Monte Carlo method for `edward`, recently.
(https://github.com/blei-lab/edward/pull/865)
I also want to implement this method for this repository.

Can I contribute for this project if I implement it?

Thanks.

jvdillon:
Hi YoshikawaMasashi. We'd love to have your contribution!

Ideally it'd be possible to implement the low-level MCMC algorithm as a tfp.mcmc.TransitionKernel. Then, working with Dustin, you could consider using this low-level code in Edward2.

In a matter of days we'll be adding RandomWalkMetropolis so this may serve as a nice example in addition to HMC. (But feel free to begin your implementation whenever you like.)

junpenglao:
BTW this can be close now.

PR #4, opened 2018-03-25
Implementation of Replica Exchange Monte Carlo

YoshikawaMasashi:
Hi!

I implemented `TransitionKernel` that perform Replica Exchange Monte Carlo Method(#3).
I would be appreciated your comment.


jvdillon:
Awesome! Thanks! I'll take a look and report back within 24hrs.

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,365 @@
...
+  return default_exchange_proposed_fn_
+
+
+class ReplicaExchangeMC(kernel_base.TransitionKernel):
+  """Runs one step of the Replica Exchange Monte Carlo.
+
+  [Replica Exchange Monte Carlo](
+  https://en.wikipedia.org/wiki/Parallel_tempering) is a Markov chain
+  Monte Carlo (MCMC) algorithm that is also known as Parallel Tempering. This
+  algorithm perform multiple sampling with different temperatures in parallel,
```
twiecki: performs

YoshikawaMasashi: Thanks! I fixed.

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,365 @@
...
+def default_exchange_proposed_fn(freq):
+  """Default function for `exchange_proposed_fn` of `kernel`.
+  Depending on the probability of `freq`, decide whether to propose combinations
+  of replica for exchange.
+  When exchanging, create combinations of adjacent replicas from 0 or 1 index.
+  """
+  def default_exchange_proposed_fn_(n_replica, seed=None):
+    seed = distributions_util.gen_new_seed(seed, "default_exchange_proposed_fn")
+    u = tf.random_uniform([], seed=seed)
+    exist = u > freq
```
jvdillon: If `freq` is a non-scalar, I worry that this will do the wrong thing. I think we want to use:

    u = tf.random_uniform(tf.shape(freq), dtype=freq.dtype.base_dtype, seed=seed)

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,365 @@
...
+
+def default_exchange_proposed_fn(freq):
+  """Default function for `exchange_proposed_fn` of `kernel`.
+  Depending on the probability of `freq`, decide whether to propose combinations
+  of replica for exchange.
+  When exchanging, create combinations of adjacent replicas from 0 or 1 index.
+  """
+  def default_exchange_proposed_fn_(n_replica, seed=None):
+    seed = distributions_util.gen_new_seed(seed, "default_exchange_proposed_fn")
+    u = tf.random_uniform([], seed=seed)
```
jvdillon: Since the nature of our work is so complex, we try to use as descriptive names as possible. Perhaps call this `random_uniform`?

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,365 @@
...
+    ])
+
+
+def default_exchange_proposed_fn(freq):
+  """Default function for `exchange_proposed_fn` of `kernel`.
+  Depending on the probability of `freq`, decide whether to propose combinations
+  of replica for exchange.
+  When exchanging, create combinations of adjacent replicas from 0 or 1 index.
+  """
+  def default_exchange_proposed_fn_(n_replica, seed=None):
```
jvdillon: num_replica

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,365 @@
...
+    seed = distributions_util.gen_new_seed(seed, "default_exchange_proposed_fn")
+    u = tf.random_uniform([], seed=seed)
+    exist = u > freq
+
+    seed = distributions_util.gen_new_seed(seed, "default_exchange_proposed_fn")
+    i = tf.random_uniform([], minval=0, maxval=2, dtype=tf.int32, seed=seed)
```
jvdillon: random_range_start

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,365 @@
...
+    a = tf.range(i, n_replica - 1, 2)
```
jvdillon: random_tail_dims?

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,365 @@
...
+    exchange_proposed = tf.transpose(tf.concat(([a], [a + 1]), axis=0))
+    exchange_proposed_n = tf.to_int32(tf.floor(
+        (tf.to_int32(n_replica) - i) / 2))
+
+    exchange_proposed = tf.cond(
+        exist, lambda: tf.to_int32([]), lambda: exchange_proposed)
+    exchange_proposed_n = tf.cond(exist, lambda: 0, lambda: exchange_proposed_n)
```
jvdillon: For computational reasons its usually better if we can avoid graph control flow.  How about:

    tf.where(exist, tf.zeros_like(exchange_proposed_n), exchange_proposed_n)

I dont see an easy way to rewrite the first use of `tf.cond`, but maybe you do?

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,365 @@
...
+  target = tfd.MixtureSameFamily(
+      mixture_distribution=tfd.Categorical(probs=[0.5, 0.5]),
+      components_distribution=tfd.MultivariateNormalDiag(
+          loc=[[-5., -5], [5., 5.]],
+          scale_identity_multiplier=[1., 1.]))
+
+  # Initialize the ReplicaExchangeMC sampler.
+  remc = tfp.mcmc.ReplicaExchangeMC(
+      target_log_prob_fn=target.log_prob,
+      inverse_temperatures=tf.pow(10., tf.linspace(0., -2., 5)),
```
jvdillon: Slightly nicer to use operators here, eg,

    10.**tf.linspace(0., -2., 5)

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,365 @@
...
+      other_results.replica_states[i]) for i in range(remc.n_replica)]
+
+  warmup = tf.group([x_update, replica_update])
+
+  init = tf.global_variables_initializer()
+
+  with tf.Session() as sess:
+    sess.run(init)
+    # Warm up the sampler
+    for _ in range(num_warmup_iter):
```
jvdillon: This example is great as is. However it might be nice to teach users about our sample_chain driver, which is muuuch more efficient that repeated calls.  Eg,

    sampled_states, kernel_results = tfp.mcmc.sample_chain(
        num_results=num_chain_iter,
        init_state=tf.zeros([num_chain_iter, 2]),
        num_burnin_steps=num_warmup_iter)

This approach also saves you the trouble of using `tf.Variables`. This is good because its a simpler example and (hopefully) dramatically more efficient.

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,365 @@
...
+        `current_state` (or `*current_state` if it's a list) and returns its
+        (possibly unnormalized) log-density under the target distribution.
+      inverse_temperatures: sequence of inverse temperatures to perform
+        samplings with each replica.
+      replica_kernel_class: mcmc kernel class for multiple sampling with
+        different temperatures in parallel
+      exchange_proposed_fn: Python callable which take a number of replicas, and
+        return combinations of replicas for exchange.
+      seed: Python integer to seed the random number generator.
+      name: Python `str` name prefixed to Ops created by this function.
```
jvdillon: Add:
      Default value: `None` (i.e., no seed).

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,365 @@
...
+               exchange_proposed_fn=default_exchange_proposed_fn(1.),
+               seed=None, name=None, **kwargs):
+    """Instantiates this object.
+
+    Args:
+      target_log_prob_fn: Python callable which takes an argument like
+        `current_state` (or `*current_state` if it's a list) and returns its
+        (possibly unnormalized) log-density under the target distribution.
+      inverse_temperatures: sequence of inverse temperatures to perform
+        samplings with each replica.
```
jvdillon: Must have statically known `rank` and statically known leading shape, i.e., `inverse_temperatures.shape[0].value is not None`

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,365 @@
...
+      replica_kernel_class: mcmc kernel class for multiple sampling with
+        different temperatures in parallel
+      exchange_proposed_fn: Python callable which take a number of replicas, and
+        return combinations of replicas for exchange.
+      seed: Python integer to seed the random number generator.
+      name: Python `str` name prefixed to Ops created by this function.
+        Default value: `None` (i.e., "remc_kernel").
+      **kwargs: Arguments for `replica_kernel_class`.
+    """
```
jvdillon: raise ValueError if `inverse_temperatures.shape.ndims is None or inverse_temperatures.shape[0].value is None`

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,365 @@
...
+                                   lambda: next_replica_idx + exchange_op,
+                                   lambda: next_replica_idx)
+        return [i + 1, next_replica_idx]
+
+      next_replica_idx = tf.while_loop(
+          cond, body, loop_vars=[i, next_replica_idx])[1]
+
+      next_replica_states = []
+      next_replica_results = []
+      for i in range(self.n_replica):
```
jvdillon: Note: since youre using Python control flow (vs TF) we'll need to raise an exception to ensure user meets this contract. (I left comment above.)

tensorflow_probability/python/mcmc/replica_exchange_mc_test.py
```diff
@@ -0,0 +1,84 @@
...
+      for _ in range(num_warmup_iter):
+        sess.run(warmup)
+      # Collect samples
+      samples = np.zeros([num_chain_iter, 2])
+      replica_samples = np.zeros([num_chain_iter, 5, 2])
+      for i in range(num_chain_iter):
+        _, x_, replica_x_ = sess.run([x_update, x, replica_update])
+        samples[i] = x_
+        replica_samples[i] = replica_x_
+
```
jvdillon: We should check that the results are consistent with the `target_log_prob_fn`.

Lets also try to rework this example to use `tfp.mcmc.sample_chain`.

tensorflow_probability/python/mcmc/replica_exchange_mc_test.py
```diff
@@ -0,0 +1,84 @@
...
+
```
jvdillon: Can you add unit-tests which stress that all the exceptions are appropriately raised?

jvdillon:
Sorry, I didnt mean to approve this just yet!

tensorflow_probability/python/mcmc/replica_exchange_mc.py
YoshikawaMasashi: Thank you for your comment!

I want to limit `freq` to scalar. Because exchanging is done for the states (set of variables) collectively rather than the individual variables.

I would appreciate it if you could reply.

tensorflow_probability/python/mcmc/replica_exchange_mc.py
YoshikawaMasashi: Thanks.
I'll use slicing.
```
exchange_proposed_n = tf.where(exist, tf.zeros_like(exchange_proposed_n),
                                   exchange_proposed_n)
exchange_proposed = exchange_proposed[:exchange_proposed_n]
```

tensorflow_probability/python/mcmc/replica_exchange_mc.py
YoshikawaMasashi: Thanks! I'll add new example of sample_chain.

==============================================
I tried using `sample_chain`, but I got an error.

```python
dtype = np.float32

target = tfd.Normal(loc=dtype(0), scale=dtype(1))

samples, _ = tfp.mcmc.sample_chain(
  num_results=1000,
  current_state=dtype(1),
  kernel=tfp.mcmc.RandomWalkMetropolis(
     target.log_prob,
     seed=42),
  num_burnin_steps=500,
  parallel_iterations=1)  # For determinism.
```

Error is this.
```
TypeError: The two structures don't have the same sequence type. First structure has type <class 'tensorflow_probability.python.mcmc.random_walk_metropolis.UncalibratedRandomWalkResults'>, while second structure has type <class 'list'>.
```

I think this cause is that a kernel result before `one_step` and this after `one_step` are different.

```python
with tf.Session() as sess:
  target = tfd.Normal(loc=dtype(0), scale=dtype(1))
  kernel = tfp.mcmc.RandomWalkMetropolis(
          target.log_prob,
          seed=42)
  
  cs = dtype(1)
  pkr = kernel.bootstrap_results(cs)
  
  print(pkr.accepted_results)
  cs, pkr = kernel.one_step(cs, pkr)
  print(pkr.accepted_results)
```

This code prints this.
```
UncalibratedRandomWalkResults(log_acceptance_correction=<tf.Tensor 'mh_bootstrap_results/rwm_bootstrap_results/zeros_like:0' shape=() dtype=float32>, target_log_prob=<tf.Tensor 'mh_bootstrap_results/rwm_bootstrap_results/Normal/log_prob/sub:0' shape=() dtype=float32>)

[<tf.Tensor 'mh_one_step/choose_inner_results/Select:0' shape=() dtype=float32>, <tf.Tensor 'mh_one_step/choose_inner_results/Select_1:0' shape=() dtype=float32>]
```

I rewrote `is_namedtuple_like` like this. Then I got no error.
```python:tensorflow_probability/python/mcmc/util.py
def is_namedtuple_like(x):
  """Helper which returns `True` if input is `collections.namedtuple`-like."""
  return (isinstance(x, tuple) or
          isinstance(getattr(x, '__dict__', None), collections.Mapping) and
          getattr(x, '_fields', None) is not None)
```

YoshikawaMasashi:
I modified some codes based on comments. However, it is not yet all.

When I used `sample_chain`, I got `FailedPreconditionError`. I think this cause is to use uninitialized `replica_states`. I''ll fix.

tensorflow_probability/python/mcmc/replica_exchange_mc.py
YoshikawaMasashi: When I use `sample_chain`, `exchange_proposed` whose shape is not invariant cause an error. Therefore, I rewrote this function so that `exchange_proposed` has invariant shape.

YoshikawaMasashi:
I implemented to generate `replica_states` as `tf.Variable` in `bootstrap_results` only when `init_state` is `tf.Variable`.

YoshikawaMasashi:
@jvdillon 

Thank you for your quick review!

Based on your comment, I think I corrected everything I can fix.
However I commented on the part that I have some trouble about fixing.

I would appreciate it if you would check and advise.

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,426 @@
...
+    with tf.name_scope(
+        name=mcmc_util.make_name(self.name, 'remc', 'bootstrap_results'),
+        values=[init_state]):
+      replica_results = [self.replica_kernels[i].bootstrap_results(init_state)
+                         for i in range(self.num_replica)]
+
+      init_state_parts = (list(init_state)
+                          if mcmc_util.is_list_like(init_state)
+                          else [init_state])
+      replica_states = [[tf.Variable(s) if isinstance(s, tf.Variable)
```
jvdillon: We try to avoid using `tf.Variable` or `tf.get_variable` in any of our API code.  I dont think you actually need to create a variable here. Ie, the user provided `target_log_prob_fn` should not presume its input is a `tf.Variable` instance.

Typically users would do something like:

    state = tf.get_variable("state", shape=[], dtype=np.float32)
    state = tf.assign(state, hmc.one_step(state, hmc.bootstrap_results(state))[0])

Ie, all handling of variables should happen in user code, not API code.

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -49,22 +49,42 @@ def default_exchange_proposed_fn(freq):
...
   """
-  def default_exchange_proposed_fn_(n_replica, seed=None):
+  def default_exchange_proposed_fn_(num_replica, seed=None):
+    num_replica = tf.to_int32(num_replica)
+
     seed = distributions_util.gen_new_seed(seed, "default_exchange_proposed_fn")
-    u = tf.random_uniform([], seed=seed)
-    exist = u > freq
+    random_uniform = tf.random_uniform([], seed=seed)
+    exist = random_uniform > freq
```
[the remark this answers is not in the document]
YoshikawaMasashi: Yes! I'll fix it.

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -0,0 +1,426 @@
...
+            sampled_state,
+            sampled_results,
+        ] = self.replica_kernels[i].one_step(
+            previous_kernel_results.replica_states[i],
+            previous_kernel_results.replica_results[i])
+        sampled_replica_states.append(sampled_state)
+        sampled_replica_results.append(sampled_results)
+        if not mcmc_util.is_list_like(sampled_state):
+          sampled_state = [sampled_state]
+        sampled_replica_ratios.append(self.target_log_prob_fn(*sampled_state))
```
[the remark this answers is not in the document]
YoshikawaMasashi: Thanks for advice!

I am trying to implement, but the result does not go well.

To recalculate ratios, the code and sampling result is this.
```Python
sampled_replica_ratios = [self.target_log_prob_fn(*srs)
                          if mcmc_util.is_list_like(srs)
                          else self.target_log_prob_fn(*[srs])
                          for i, srs in enumerate(sampled_replica_states)]
sampled_replica_ratios = tf.stack(sampled_replica_ratios, axis=0)
```
image


But to avoid recalculation, the code and sampling result is this.
```Python
sampled_replica_ratios = \
    [srr.target_log_prob if 'target_log_prob' in srr._fields
     else srr.accepted_results.target_log_prob
     for i, srr in enumerate(sampled_replica_results)]
sampled_replica_ratios = \
    tf.stack(sampled_replica_ratios, axis=0) / self.inverse_temperatures
```
image

I think it happens because the error comes out from the original value when dividing after multiplication.

I think that it would be better to recalculate, for accuracy.
How about this?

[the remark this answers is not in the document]
jvdillon: The two should be identical so this suggests something strange is going on.

One thing that stands out to me. I _think_ you want:
     tf.stack(sampled_replica_ratios, axis=-1) / self.inverse_temperatures
Broadcasting pads 1's to the left of a shape, not the right. By stacking on the rightmost dim, this ensures that the inverse_temps are broadcast across chains.

Having said that, I doubt this will fix the problem because I think it would have errored out if this was really the problem. (Although if you happened to have the same number of chains as replicas then this could indeed be the culprit.)

Since I dont have any other ideas if youre unable to track down the bug, I will look into it post submission and see if I can find whats going on.  This is worth fixing because we must assume target_log_prob_fn is expensive to call.  (However, that doesnt mean we need to make this optimization on our first attempt.)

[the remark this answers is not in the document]
YoshikawaMasashi: Thanks for the comments!

I tracked down and fixed the bug.
Although I must exchange inverse_temp of `next_replica_results[i].target_log_prob` of `next_replica_results[i].accepted_results.target_log_prob` after exchange replica results and states, I didn't.
I fixed not to recalculate.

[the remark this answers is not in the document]
jvdillon: Great!  This is a big savings, so Im glad you were able to identify (and correct!) the problem.

YoshikawaMasashi:
I modified some codes based on the comments.
Could you check?

YoshikawaMasashi:
Thanks for your comment and kindness!
I was wrong.
I fixed not to use kwargs.
Could you check again?

jvdillon:
Thanks for your contribution! And again, my sincere apologies for the duration of this review. As you are one of the first major contributors to this repository, this meant your PR got caught up in unusual delays.  In the future I expect the review process to take <= 1 week.

A note about how our merges work:

After we iterate on PR here on Github, we pull the change into our internal version control system and submit the change there. An automated system then generates a git commit and pushes that to the repo here. At that point the change is effectively merged, but perhaps not in the "usual" sense of a Github-only workflow. We close the PR manually once we have submitted the change on our end. It should be visibly reflected in the Github repo in O(minutes).

Thanks again for the contribution!

jvdillon:
Looks like there's a merge problem. Can I get you to rebase?

YoshikawaMasashi:
I rebased!

PR #17, opened 2018-04-12
added decode call to load_insteval

lbollar:
I believe load_insteval() was broken in its current form for python 3. I added a call to decode to convert from bytes before calling splitlines. I tried the method in a python 2 notebook to make sure this didn't break it and it appeared to work fine.


junpenglao:
I think it is much easier to just read the table using Pandas:
```python
url = ('https://raw.github.com/vincentarelbundock/Rdatasets/master/csv/'
         'lme4/InstEval.csv')
data = pd.read_csv(url)
data = data.drop(columns=['Unnamed: 0'])
```

dustinvtran:
Thanks @lbollar. Can you make the change while minimally affecting the Jupyter notebook? (i.e., change that line and leave rest of the cells not run)

@junpenglao: That's a good point. I think for now, we decided to attempt relying on Pandas as least as possible (and ideally drop it).

junpenglao:
I see - but since you already imported pandas here I thought it would be easier to just go pandas all the way.

dustinvtran:
That's a sensible suggestion. In general, the context is TF best practices, which if I understand correctly, is that instead of a "DataFrame" abstraction, you should be using dicts of "Column" abstractions. (For example, the code scales nicely to use tf.feature_column if you want to generalize this to TFRecords, tf.data, and the TF Estimator API.)

At the moment, pandas is used only to visualize the data. We may end up dropping that utility.

lbollar:
Yeah sorry, I knew I would screw up a notebook pull. Didn't see the huge diff. Will submit later tonight.

PR #18, opened 2018-04-13
added decode call to contents

lbollar:
Same as [#17](https://github.com/tensorflow/probability/pull/17), easier to just create new branch.


jvdillon:
Thanks!

jvdillon:
Change merged. Thanks!

A note about how our merges work (just in case there's any confusion):

After we iterate on PR here on Github, we pull the change into our internal version control system and submit the change there. An automated system then generates a git commit and pushes that to the repo here. At that point the change is effectively merged, but perhaps not in the "usual" sense of a Github-only workflow. We close the PR manually once we have submitted the change on our end. It should be visibly reflected in the Github repo in O(seconds).

Thanks again for the contribution!

PR #52, opened 2018-05-25
Use DiCE objective

jakobnicolaus:
Rewriting the monte-carlo expectation to use the DiCE objective rather than the surrogate loss approach. This ensures that any order gradients of the objective produce unbiased estimates of higher order gradients (including zero order) when not using reparameterization.

### DIFF
--- a/tensorflow_probability/python/monte_carlo.py
+++ b/tensorflow_probability/python/monte_carlo.py
@@ -175,7 +175,7 @@ def expectation(f, samples, log_prob=None, use_reparametrization=True,
       # https://arxiv.org/abs/1802.05098
       # DiCE ensures that any order gradients of the objective 
       # are unbiased gradient estimators
-      l_dice = fx * tf.exp(logpx - stop(logpx))  # Add zeros_like(logpx).
+      l_dice = fx * tf.exp(logpx - stop(logpx))
       return tf.reduce_mean(l_dice, axis=axis, keep_dims=keep_dims)
 
 
--- a/tensorflow_probability/python/monte_carlo.py
+++ b/tensorflow_probability/python/monte_carlo.py
@@ -174,7 +174,16 @@ def expectation(f, samples, log_prob=None, use_reparametrization=True,
       # "DiCE: The Infinitely Differentiable Monte-Carlo Estimator"
       # https://arxiv.org/abs/1802.05098
       # DiCE ensures that any order gradients of the objective 
-      # are unbiased gradient estimators
+      # are unbiased gradient estimators.
+      # We exploit that
+      #   `h(x) - stop(h(x)) == zeros_like(h(x))`
+      # but its gradient is grad[h(x)].
+      # Note that IEEE754 specifies that `x - x == 0.` and `x + 0. == x`, hence
+      # this trick loses no precision. For more discussion regarding the
+      # relevant portions of the IEEE754 standard, see the StackOverflow
+      # question,
+      # "Is there a floating point value of x, for which x-x == 0 is false?"
+      # http://stackoverflow.com/q/2686644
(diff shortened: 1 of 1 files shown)


jakobnicolaus:
I signed it! (Just signed the cla)

jvdillon:
Hi Jakob, thanks for the contribution!  Its awesome that you've written a paper which deeply explores this trick! I think we'd love to accept your proposed change and cite your paper in our code. Given that we've used this objective function for a year now, would you consider updating your paper to cite our first implementation? As a happy side-effect, this might be a good way to prove the usefulness of the approach.

Regarding removal of the comments, we definitely want to keep those.  They specify that there's no loss of precision, assuming IEEE754 compliant hardware. This is is an important point in part because it isnt true on all hardware (some of which is very popular!).

I do have a technical question I'd also like to run by you. Your proposed change is very close to the original version,
https://github.com/tensorflow/tensorflow/commit/cf88415190b04409227819005d0dea6ebbbd8b01#diff-26cf6afab94612658d3096a8048fa6e0

Soon after the original, we changed the formula to what you now see, ie,
https://github.com/tensorflow/tensorflow/commit/9c08e99a02c682acca0dcb9be9f79c3a9d38b615#diff-26cf6afab94612658d3096a8048fa6e0

My thinking at the time was that avoiding the computational cost of the exp is good. Since we rarely need higher order derivatives in TF, this seemed a fair trade-off. However, I actually never profiled the difference. (My possibly incorrect preconception is that exp is ~80x slower than divide.)  Since this is a low-level piece of code, constants really matter--we don't know the asymptotic costs at play.  With this in mind, can you make the case why we want to restore things to the old style? Some realistic use case where using the second order derivative of the reparameterized output makes sense?

Another possibility here is that we add a bool which switches between the two approaches something which conveys the semantics of: "fast_but_only_differentiable_once". Although I'm not crazy about adding another arg, but Id like to hear your thoughts.

jakobnicolaus:
Hi Josh, neat that you had used the same trick in an older version of the code -- we'll update the paper to that effect. We are curious if your original motivation was to support higher order gradients in contrast to the 'Surrogate Loss' approach? If so, please let us know any reference for this (either in documentation, commentary or implementation) so we can appropriately mention it in the paper. 

Regarding use-cases: Meta-learning and all higher-order methods in reinforcement learning require higher order derivatives of the score-function gradient estimator. The use-case is certainly realistic for some interesting current research ([1](https://arxiv.org/abs/1703.03400),[2](http://www.jmlr.org/papers/volume17/15-414/15-414.pdf),[3](https://arxiv.org/abs/1710.03641),[4](https://arxiv.org/abs/1709.04326)).

Although this is low-level code, our impression is that for RL (a main use case for the non-reparameterized objective) it should be called comparatively infrequently (only the action-sampling distribution requires this treatment). Having said this, we are happy to add the switch functionality if you think it will be better for other use-cases.

Agreed that the comments on precision should usefully remain, we will put them back in.

Let us know if there’s anything else we can do to help!

dustinvtran:
pinging @jvdillon. They would love to get this in by camera-ready (which was yesterday), so they can link to this in their paper. Any way we can fast track this code review process?

jvdillon:
Hi Jakob. Thanks for offering to cite our previous implementation.  At the time there was no request for higher order derivatives; this is why I chose to drop the exp.

If its not too late, you might want to mention in your paper the alternative of not using exp and why exp is better.

Thanks again for the PR!

jvdillon:
I will now merge this PR. Due to time pressure, I slightly restructured your comment and s/l_dice/dice/. If youre unhappy with these minor edits, please feel free to submit another PR.

jvdillon:
Merge complete. I will now close this PR. As mentioned, I made some minor edits. If you disfavor these edits please start a new PR. Thanks again!

PR #84, opened 2018-06-20
Stochastic Block Model example

rlouf:
I took the example from Edward and tried to implement it with TFP. Since I could not find an implementation of MAP, I implemented the computation of the loss as well.

However, I could not get the optimization to converge. I must be doing something wrong (first time using tensorflow), and I started this PR to get some help.

### DIFF
--- a/tensorflow_probability/examples/stochastic_block_model.py
+++ b/tensorflow_probability/examples/stochastic_block_model.py
@@ -1,0 +1,86 @@
+import time
+
+from matplotlib import pylab as plt
+import numpy as np
+from observations import karate
+import tensorflow as tf
+from tensorflow_probability import edward2 as ed
+
+
+# Data & parameters
+# -----------------------------------------------------------------------------
+x_data, z = karate('~/data')
+V = x_data.shape[1]
+K = 2
+learning_rate = 1e-4
+max_steps = 10000
+epsilon= 0.001
+
+
+def compute_loss(latent_vars, data):
+    """ Compute the loss associated with MAP calculations.
+    """
+    dict_vals = {var: latent.value for var, latent in latent_vars.items()}
+    for x, datum in data.items(): # assuming we have the values as input
+        dict_vals[x] = datum
+    
+    log_prob = 0.0
+    for z in latent_vars.keys():
+        z_copy = z.distribution.copy()
+        log_prob += tf.reduce_sum(z_copy.log_prob(dict_vals[z]))
+
+    for x in data.keys():
+        x_copy = x.distribution.copy()
+        log_prob += tf.reduce_sum(x_copy.log_prob(dict_vals[x]))
+
+    reg_penalty = tf.reduce_sum(tf.losses.get_regularization_losses())
+    loss = -log_prob + reg_penalty
+    
+    return loss
+
+
+# Model
+# -----------------------------------------------------------------------------
+gamma = ed.Dirichlet(tf.ones([K]))
+pi = ed.Beta(tf.ones([K,K]), tf.ones([K,K]))
+z = ed.Multinomial(tf.ones([V]), gamma)
+x = ed.Bernoulli(tf.matmul(z, tf.matmul(pi, tf.transpose(z))))
+
+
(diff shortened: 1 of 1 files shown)


tensorflow_probability/examples/stochastic_block_model.py
```diff
@@ -0,0 +1,86 @@
...
+    """ Compute the loss associated with MAP calculations.
+    """
+    dict_vals = {var: latent.value for var, latent in latent_vars.items()}
+    for x, datum in data.items(): # assuming we have the values as input
+        dict_vals[x] = datum
+    
+    log_prob = 0.0
+    for z in latent_vars.keys():
+        z_copy = z.distribution.copy()
+        log_prob += tf.reduce_sum(z_copy.log_prob(dict_vals[z]))
```
jvdillon: I think this indirection is probably necessary because youre using Edward's wrapped TF distribution. Is there some advantage of using Edward here vs directly using TF distributions?

I ask because the typical use case of Edward would be to call make_log_joint from what is otherwise the generative specification. In your case you dont need make_log_joint so I think this example could me made simpler by directly using TF distributions, eg, similar to the Bayesian GMM colab.

Thoughts?

rlouf: Hi! Thanks for your help! The only advantage I see is simplicity: the model specification is a literal translation of the equations; I find it more explicit and expressive. It would make sense for me to implement something like `find_MAP` in edward2.

Now, I understand why it would make more sense here (and now) to use TF distributions directly. I will work on that. In the meantime, do you have any idea why the current scripts does not converge?

dustinvtran: I'd actually to push back on Josh's comment. You're writing a log joint which is just the reduce sum of individual distribution's log probs. This is clearly an application of Edward2 as the direct translation as a generative program + `ed.make_log_joint` is clearer.

Maybe to be more convincing, can you show generations or predictions of held out elements? This requires more than a log-joint and without Edward2, would require writing down the model twice.

dustinvtran:
I'd actually to push back on Josh's comment. You're writing a log joint which is just the reduce sum of individual distribution's log probs. This is clearly an application of Edward2 as the direct translation as a generative program + ed.make_log_joint is clearer.

Maybe to be more convincing, can you show generations or predictions of held out elements? This requires more than a log-joint and without Edward2, would require writing down the model twice.

>  I have re-read this several times and can't find any error. Is there something that I am missing?

How does it compare to Edward1's example? Note there may be an identification issue as z_true and z_pred can be the same up to relabeling.

rlouf:
This makes a lot more sense to me, so I changed the example to use Edward2. However, I get exactly the same result as previously (which, in a way, is reassuring):

```
z_pred = [0 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 0 0]
z_true = [0 0 0 0 0 0 0 0 0 1 0 0 0 0 1 1 0 0 1 0 1 0 1 1 1 1 1 1 1 1 1 1 1 1]
```

The result is still not great when if you switch the labels. Also, I found [this issue](https://github.com/blei-lab/edward/pull/715) describing a similar problem. It made me think of a [paper](https://arxiv.org/pdf/1008.3926.pdf) where Newman (re?)introduces degree-corrected stochastic blockmodels. If you look at Figure 1, you see that the standard SBM performs poorly because two individuals have a disproportionately large degree. On the other hand, the degree-corrected model gives results that make complete sense.
So I think it may be that we are trying to fit the wrong model. It makes for a great example of the specification-inference-criticism loop!

PS: I take note of your suggestions and will show prediction of heldout elements once I've figured the convergence issue out.

rlouf:
Yes, I'm up to it. I'm in a bit of a rush at work at the moment, but will do when the clouds clear.

rlouf:
Just a quick note to say I started working on this again! The improvement is more involved than expected, but I hope to have something soon.

Issue #82, opened 2018-06-15
Error in Linear_Mixed_Effects_Models notebook


huangy22:
Here is my version:
```
>>> sys.version
'3.6.5 (default, Apr 25 2018, 14:23:58) \n[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)]'
>>> tf.__version__
'1.10.0-dev20180610'
```

When running the example notebook for Linear Mixed Effects Models, I had the following error:
```
log_joint = ed.make_log_joint_fn(model_template)

def target_log_prob_fn(effect_students, effect_instructors, effect_departments):
  """Unnormalized target density as a function of states."""
  return log_joint(  # fix `features` and `ratings` to the training data
    features=features_train,
    effect_students=effect_students,
    effect_instructors=effect_instructors,
    effect_departments=effect_departments,
    ratings=labels_train)

# Set up E-step (MCMC).
effect_students = tf.get_variable(  # `trainable=False` so unaffected by M-step
    "effect_students", [num_students], trainable=False)
effect_instructors = tf.get_variable(
    "effect_instructors", [num_instructors], trainable=False)
effect_departments = tf.get_variable(
    "effect_departments", [num_departments], trainable=False)

hmc = tfp.mcmc.HamiltonianMonteCarlo(
    target_log_prob_fn=target_log_prob_fn,
    step_size=0.015,
    num_leapfrog_steps=3)

current_state = [effect_students, effect_instructors, effect_departments]
with warnings.catch_warnings():
  # TensorFlow raises a warning about converting sparse IndexedSlices to a
  # dense Tensor during gradient computation. This can consume a large amount
  # of memory. We're okay with that as the number of categories is small.
  warnings.simplefilter("ignore")
  next_state, kernel_results = hmc.one_step(
      current_state=current_state,
      previous_kernel_results=hmc.bootstrap_results(current_state))

expectation_update = tf.group(
    effect_students.assign(next_state[0]),
    effect_instructors.assign(next_state[1]),
    effect_departments.assign(next_state[2]))

# Set up M-step (gradient descent).
# The following should work. However, TensorFlow raises an error about taking
# gradients through IndexedSlices tensors. This may be a TF bug. For now,
# we recompute the target's log probability at the current state.
# loss = -kernel_results.accepted_results.target_log_prob
with tf.control_dependencies([expectation_update]):
  loss = -target_log_prob_fn(effect_students,
                             effect_instructors,
                             effect_departments)
[...]

Issue #59, opened 2018-05-31
small typo in example Linear_Mixed_Effects_Models.ipynb with "<stripped %d bytes>"%size


ngakrelidz:
Hi there,

Small typo in this example= examples/jupyter_notebooks/Linear_Mixed_Effects_Models.ipynb , in the function def strip_consts(graph_def, max_const_size=32), 

tensor.tensor_content = "<stripped %d bytes>"%size

must be replaced by : tensor.tensor_content = bytes("<stripped %d bytes>"%size, 'utf-8')

ngakrelidz:
i'm running Python 3.5, i didn't try with 2.7 .

PR #89, opened 2018-06-29
Linear_Mixed_Effects_Models: fix the typo in content size, fix the bug about variable reuse

huangy22:
Address the typo mentioned in issue #59 and the variable reuse error in issue #82.
In the function `strip_consts`, replace
```
tensor.tensor_content = "<stripped %d bytes>"%size
```
with 
```
tensor.tensor_content = bytes("<stripped %d bytes>"%size, 'utf-8')
```

In the "Parameter Estimation" session, add the following line
```
tf.reset_default_graph()
```
before
```
effect_students = tf.get_variable( ...
```


huangy22:
I signed it!

csuter:
I force reverted the master branch to the point before this PR. This is because we need to use our internal version control system and automated push process to make changes to the Github repo. We will pull this change in through that method shortly.

PR #97, opened 2018-07-05
Adding tutorial for probabilistic principal component analysis

kouariga:
I'm adding a tutorial for probabilistic PCA by basically adapting the original Edward's tutorial but with HMC inference instead of VI. I would appreciate reviews as I'm new to both Tensorflow and Bayesian statistics. I haven't understood how to use VI framework of Tensorflow Probability yet, but I'll add that too in the inference section, once I figure it out.


tensorflow_probability/examples/jupyter_notebooks/Probabilistic_Principal_Component_Analysis.ipynb
```diff
@@ -0,0 +1,444 @@
...
+    "We use simulated data. We'll talk about the individual variables and what they stand for in the next section. For this example, each data point is 2-dimensional, $\\mathbf{x}_{n}\\in\\mathbb{R}^{2}$."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def build_toy_dataset(N, D, K, sigma=1.0):\n",
```
davmre: We prefer to generate synthetic data directly from the Edward model, rather than writing separate sampling code. (I realize you inherited this sin from the original notebook :). This guarantees we're using the same model for data generation and inference, and also saves a lot of code. See the PPCA-VI notebook (https://github.com/tensorflow/probability/blob/master/tensorflow_probability/examples/jupyter_notebooks/Probabilistic_PCA.ipynb) for an example of doing this.

tensorflow_probability/examples/jupyter_notebooks/Probabilistic_Principal_Component_Analysis.ipynb
```diff
@@ -0,0 +1,444 @@
...
+    "We visualize the model with TensorBoard."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def strip_consts(graph_def, max_const_size=32):\n",
```
davmre: This is cool! Can you add a comment pointing to the original source of this code? (seems to be based on https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/tutorials/deepdream/deepdream.ipynb ?)

tensorflow_probability/examples/jupyter_notebooks/Probabilistic_Principal_Component_Analysis.ipynb
```diff
@@ -0,0 +1,444 @@
...
...val: 1.0\\n      }\\n    }\\n  }\\n}\\nnode {\\n  name: &quot;z/sample/random_normal/RandomStandardNormal&quot;\\n  op: &quot;RandomStandardNormal&quot;\\n  input: &quot;z/sample/concat&quot;\\n  attr {\\n    key: &quot;T&quot;\\n    value {\\n      type: DT_INT32\\n    }\\n  }\\n  attr {\\n    key: &quot;dtype&quot;\\n    value {\\n      type: DT_FLOAT\\n    }\\n  }\\n  attr {\\n    key: &quot;seed&quot;\\n    value {\\n      i: 0\\n    }\\n  }\\n  attr {\\n    key: &quot;seed2&quot;\\n    value {\\n      i: 0\\n    }\\n  }\\n}\\nnode {\\n  name: &quot;z/sample/random_normal/mul&quot;\\n  op: &quot;Mul&quot;\\n  input: &quot;z/sample/random_normal/RandomStandardNormal&quot;\\n  input: &quot;z/sample/random_normal/stddev&quot;\\n  attr {\\n    key: &quot;T&quot;\\n    value {\\n      type: DT_FLOAT\\n    }\\n  }\\n}\\nnode {\\n  name: &quot;z/sample/random_normal&quot;\\n  op: &quot;Add&quot;\\n  input: &quot;z/sample/random_normal/mul&quot;\\n  input: &quot;z/sample/random_normal/mean&quot;\\n  attr {\\n    key: &quot;T&quot;\\n    value {\\n      type: DT_FLOAT\\n    }\\n  }\\n}\\nnode {\\n  name: &quot;z/sample/mul&quot;\\n  op: &quot;Mul&quot;\\n  input: &quot;z/sample/random_n
```
davmre: wrong caption here?

davmre: It looks like several of the cells in this notebook aren't evaluated. We'd like the notebook to 'read' fluidly to a viewer who just clicks through on github (plus it makes it easier to review), so please evaluate all the cells to generating figures, results, etc.

davmre: Edward RV's are automatically treated as samples, so this can just be `sess.run(posterior)`.

davmre: At first glance I'm confused why you don't also need to replace the posterior mean for `z`. Are you sure this is correct?

tensorflow_probability/examples/jupyter_notebooks/Probabilistic_Principal_Component_Analysis.ipynb
```diff
@@ -0,0 +1,444 @@
+{
```
davmre: If you elect to check this in as a separate notebook, we should rename the file to something like `Probabilistic_PCA_HMC.ipynb` to distinguish from the VI example.

srvasude:
Hi @kouariga , @davmre what is the status here?

PR #98, opened 2018-07-05
Handle deprecated method inspect.getargspec()

kouariga:
This PR is related to the issue: [DeprecationWarning: inspect.getargspec() is deprecated #78](https://github.com/tensorflow/probability/issues/78).

As in Python 3.6, inspect.getargspec() was deprecated, and using inspect.signature() or inspect.getfullargspec() is encouraged.

With this PR changes, program_transformations.py uses inspect.getfullargspec() when it's available, i.e. when Python 3 is used, and fallback on inspect.getargspec() otherwise. The attributes in the returned object are slightly different, but we only care about the `args` attribute and it's unchanged.

### DIFF
--- a/tensorflow_probability/python/edward2/program_transformations.py
+++ b/tensorflow_probability/python/edward2/program_transformations.py
@@ -129,8 +129,11 @@ def _get_function_inputs(f, **kwargs):
     Dict of key-value pairs in `kwargs` which exist in `f`'s signature.
   """
   if hasattr(f, "_func"):  # functions returned by tf.make_template
-    argspec = inspect.getargspec(f._func)  # pylint: disable=protected-access
-  else:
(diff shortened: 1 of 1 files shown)


kouariga:
I followed your instruction. Thank you so much, it was very clear. Regarding the issue #78 , I'm aware this PR doesn't solve their problem, but I linked it anyway since it's at least "related". The same kind of patch should fix the core problem, though.

csuter:
Awesome, Kousuke! I'm very glad the above string of commands actually worked 😂!

I will pull this in shortly. Thanks again!

csuter:
This is merged in now.

FYI -- I just made some internal changes that should cause these PR's to be shown as merged and automatically closed once we pull them in. Hoped the new mechanism would catch this one, but it appears not to have. I am testing it on #99 instead.

PR #102, opened 2018-07-09
Fix sinh_arcsinh_bijector_test on ppc64le

wdirons:
Skipped the check that fails due to overflow error as float128 data type
is same as float64 instead of longdouble on platforms like ppc64le.

Original PR from Tensorflow: tensorflow/tensorflow#19860

### DIFF
--- a/tensorflow_probability/python/bijectors/sinh_arcsinh_test.py
+++ b/tensorflow_probability/python/bijectors/sinh_arcsinh_test.py
@@ -155,16 +155,24 @@ class SinhArcsinhBijectorTest(tf.test.TestCase):
         self.assertAllClose(
             x, self.evaluate(bijector.inverse(y)), rtol=1e-4, atol=0.)
 
-        # Do the numpy calculation in float128 to avoid inf/nan.
-        y_float128 = np.float128(y)
-        self.assertAllClose(
-            np.log(np.cosh(
-                np.arcsinh(y_float128) / tailweight - skewness) / np.sqrt(
-                    y_float128**2 + 1)) -
-            np.log(tailweight),
-            self.evaluate(bijector.inverse_log_det_jacobian(y, event_ndims=0)),
-            rtol=1e-4,
-            atol=0.)
+        # On IBM PPC systems, longdouble (np.float128) is same as double except that it can have more precision.
+        # Type double being of 8 bytes, can't hold square of max of float64 (which is also 8 bytes) and
+        # below test fails due to overflow error giving inf. So this check avoids that error by skipping square
+        # calculation and corresponding assert.
+
+        if np.amax(y) <= np.sqrt(np.finfo(np.float128).max) and \
+           np.fabs(np.amin(y)) <= np.sqrt(np.fabs(np.finfo(np.float128).min)):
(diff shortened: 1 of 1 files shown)


csuter:
Hi @wdirons, thanks for this fix and sorry it's been idling -- not sure how it slipped through the cracks.

It seems that the merge resolutions are causing some issues on our end when we try to pull this in. I think the best resolution is to re-base the change on top of the current state of TFP master. The following sequence of commands should put your fix branch in a state that our pull system will be happy with. Please give this a try and let me know if you have any issues. Assuming all is well, I will go ahead and pull this in ASAP.

```
# The below commands assume you are cd'ed into the folder where your fork is cloned

# Create a temporary new branch
$ git checkout -b fix_merge

# Force the new branch to point to HEAD of master branch
$ git reset --hard master

# Copy the bijector test from the branch containing the (merged) fix
$ git checkout fix_sinh_arcsinh_bijector_test_ppc64le -- tensorflow_probability/python/bijectors/sinh_arcsinh_test.py

# Create a commit with the same message as the original fix
# (this is a multi-line command, including newlines in the commit message;
# copy and paste all of it and run)
$ git commit -m 'Fix sinh_arcsinh_bijector_test on ppc64le

Skipped the check that fails due to overflow error as float128 data type
is same as float64 instead of longdouble on platforms like ppc64le.'

# Now we have a commit directly on top of master, containing just the change
# you originally intended to make. Let's point the PR branch at it. First check
# out the PR branch:
$ git checkout fix_sinh_arcsinh_bijector_test_ppc64le

# Now point the PR branch at our new commit:
$ git reset --hard fix_merge

# Finally, push the PR branch state out to your remote on GitHub. Note that
# we are re-writing history here, so if you have other clones of this repo lying
# around, they will be unable to pull down the remote PR branch after this.
# The workaround, if you care about keeping copies of this fix branch around,
# would be to delete the local branch then fetch from the remote repo.
$ git push -f origin fix_sinh_arcsinh_bijector_test_ppc64le
```

After the above steps, the PR should be much simplified, showing just the one commit.

Our apologies for not getting this merged in sooner -- if we had, I suspect this git pain could've been prevented. Please feel free to reach out to me directly (cgs@google.com) if you have any issues or questions. Thanks again for this contribution!

csuter:
One last command, after the above, will eliminate the temporary branch we created:

```
$ git branch -d fix_merge
```

wdirons:
Thank you @csuter for those steps. I never would have fixed it out. Next time I will use the resolve conflicts button in the UI first and if that doesn't work, I'll revert to these steps.

Thanks again. I had forgotten about this PR myself until it was approved.

csuter:
Ah, sorry to pile on some more comments, but it looks like this doesn't pass our lint checks, which require all lines to be <= 80 chars. Can you reflow the long lines in the test file?

csuter:
No trouble at all, thanks again for the fix. I actually didn't know about ci_sanity -- we should definitely provide something like it! Thanks for the pointer. In the mean time, this change should be pulled in and closed out very soon (we're all a little busy with the fall performance review cycle at the moment :))

Issue #106, opened 2018-07-20
Conditional RealNVP bijector


hartikainen:
It would be great if [RealNVP bijector](https://github.com/tensorflow/probability/blob/67a55b2ef5e94da203991ed102c9e41d085a840a/tensorflow_probability/python/bijectors/real_nvp.py) supported conditioning. The `shift_and_log_scale_fn` can be any non-invertible mapping, so the [ConditionalBijector's](https://github.com/tensorflow/tensorflow/blob/r1.9/tensorflow/contrib/distributions/python/ops/bijectors/conditional_bijector.py#L28) `condition_kwargs` could be fed directly to it and concatenated with the inputs at the beginning of the function.

I'm gonna go ahead and submit a PR for this. Happy to take any feedback on it.

SiegeLordEx:
Should be fixed as the relevant PR is merged. Thanks for your contribution!

PR #107, opened 2018-07-20
Conditional real nvp

hartikainen:
Extends RealNVP bijector to subclass from `ConditionalBijector` to allow conditioning.

#106

### DIFF
--- a/tensorflow_probability/python/bijectors/real_nvp.py
+++ b/tensorflow_probability/python/bijectors/real_nvp.py
@@ -19,8 +19,9 @@ from __future__ import division
 from __future__ import print_function
 
 import tensorflow as tf
-from tensorflow_probability.python.bijectors import bijector
 from tensorflow.python.layers import core as layers
+from tensorflow_probability.python.bijectors.conditional_bijector import ConditionalBijector
+
 
 
 __all__ = [
@@ -29,7 +30,7 @@ __all__ = [
 ]
 
 
-class RealNVP(bijector.Bijector):
+class RealNVP(ConditionalBijector):
   """RealNVP "affine coupling layer" for vector-valued events.
 
   Real NVP models a normalizing flow on a `D`-dimensional distribution via a
@@ -176,12 +177,12 @@ class RealNVP(bijector.Bijector):
         raise ValueError(
             "Number of masked units must be smaller than the event size.")
 
-  def _forward(self, x):
+  def _forward(self, x, **condition_kwargs):
     self._cache_input_depth(x)
     # Performs scale and shift.
     x0, x1 = x[..., :self._num_masked], x[..., self._num_masked:]
     shift, log_scale = self._shift_and_log_scale_fn(
-        x0, self._input_depth - self._num_masked)
+        x0, self._input_depth - self._num_masked, **condition_kwargs)
     y1 = x1
     if log_scale is not None:
       y1 *= tf.exp(log_scale)
@@ -190,12 +191,12 @@ class RealNVP(bijector.Bijector):
     y = tf.concat([x0, y1], axis=-1)
     return y
 
-  def _inverse(self, y):
+  def _inverse(self, y, **condition_kwargs):
     self._cache_input_depth(y)
     # Performs un-shift and un-scale.
     y0, y1 = y[..., :self._num_masked], y[..., self._num_masked:]
     shift, log_scale = self._shift_and_log_scale_fn(
-        y0, self._input_depth - self._num_masked)
+        y0, self._input_depth - self._num_masked, **condition_kwargs)
     x1 = y1
     if shift is not None:
       x1 -= shift
@@ -204,20 +205,20 @@ class RealNVP(bijector.Bijector):
     x = tf.concat([y0, x1], axis=-1)
     return x
 
-  def _inverse_log_det_jacobian(self, y):
+  def _inverse_log_det_jacobian(self, y, **condition_kwargs):
     self._cache_input_depth(y)
     y0 = y[..., :self._num_masked]
     _, log_scale = self._shift_and_log_scale_fn(
-        y0, self._input_depth - self._num_masked)
+        y0, self._input_depth - self._num_masked, **condition_kwargs)
     if log_scale is None:
       return tf.constant(0., dtype=y.dtype, name="ildj")
     return -tf.reduce_sum(log_scale, axis=-1)
 
-  def _forward_log_det_jacobian(self, x):
+  def _forward_log_det_jacobian(self, x, **condition_kwargs):
     self._cache_input_depth(x)
     x0 = x[..., :self._num_masked]
     _, log_scale = self._shift_and_log_scale_fn(
-        x0, self._input_depth - self._num_masked)
+        x0, self._input_depth - self._num_masked, **condition_kwargs)
     if log_scale is None:
       return tf.constant(0., dtype=x.dtype, name="fldj")
     return tf.reduce_sum(log_scale, axis=-1)
@@ -266,8 +267,11 @@ def real_nvp_default_template(hidden_layers,
 
   with tf.name_scope(name, "real_nvp_default_template"):
 
-    def _fn(x, output_units):
+    def _fn(x, output_units, **condition_kwargs):
       """Fully connected MLP parameterized via `real_nvp_template`."""
+      x = tf.concat(
+        [x] + [condition_kwargs[k] for k in sorted(condition_kwargs)], axis=-1)
+
       for units in hidden_layers:
         x = layers.dense(
             inputs=x,
--- a/tensorflow_probability/python/bijectors/real_nvp.py
+++ b/tensorflow_probability/python/bijectors/real_nvp.py
@@ -23,7 +23,6 @@ from tensorflow_probability.python.bijectors import conditional_bijector
 from tensorflow.python.layers import core as layers
 
 
-
 __all__ = [
     "RealNVP",
     "real_nvp_default_template"
--- a/tensorflow_probability/python/bijectors/real_nvp.py
+++ b/tensorflow_probability/python/bijectors/real_nvp.py
@@ -236,6 +236,11 @@ def real_nvp_default_template(hidden_layers,
   created once. It takes the `d`-dimensional input x[0:d] and returns the `D-d`
   dimensional outputs `loc` ("mu") and `log_scale` ("alpha").
 
+  The default template does not support conditioning, and will raise an
+  exception if `condition_kwargs` are passed to it. To use conditioning in a
+  real nvp bijector, implement a conditioned scale/translation template that
+  handles the `condition_kwargs`.
+
   Arguments:
     hidden_layers: Python `list`-like of non-negative integer, scalars
       indicating the number of units in each hidden layer. Default: `[512, 512].
@@ -256,7 +261,7 @@ def real_nvp_default_template(hidden_layers,
 
   Raises:
     NotImplementedError: if rightmost dimension of `inputs` is unknown prior to
-      graph execution.
+      graph execution, or if `condition_kwargs` is not empty.
 
   #### References
 
@@ -269,9 +274,9 @@ def real_nvp_default_template(hidden_layers,
 
     def _fn(x, output_units, **condition_kwargs):
       """Fully connected MLP parameterized via `real_nvp_template`."""
-      x = tf.concat(
-          [x] + [condition_kwargs[k] for k in sorted(condition_kwargs)],
-          axis=-1)
+      if condition_kwargs:
+        raise NotImplementedError(
+            "Conditioning not implemented in the default template.")
 
       for units in hidden_layers:
         x = layers.dense(
--- a/tensorflow_probability/python/bijectors/real_nvp.py
+++ b/tensorflow_probability/python/bijectors/real_nvp.py
@@ -19,8 +19,8 @@ from __future__ import division
 from __future__ import print_function
 
 import tensorflow as tf
-from tensorflow.python.layers import core as layers
 from tensorflow_probability.python.bijectors import conditional_bijector
+from tensorflow.python.layers import core as layers
 
 
 
@@ -270,7 +270,8 @@ def real_nvp_default_template(hidden_layers,
     def _fn(x, output_units, **condition_kwargs):
       """Fully connected MLP parameterized via `real_nvp_template`."""
       x = tf.concat(
-        [x] + [condition_kwargs[k] for k in sorted(condition_kwargs)], axis=-1)
+          [x] + [condition_kwargs[k] for k in sorted(condition_kwargs)],
+          axis=-1)
 
       for units in hidden_layers:
         x = layers.dense(
--- a/tensorflow_probability/python/bijectors/real_nvp.py
+++ b/tensorflow_probability/python/bijectors/real_nvp.py
@@ -20,7 +20,7 @@ from __future__ import print_function
 
 import tensorflow as tf
 from tensorflow.python.layers import core as layers
-from tensorflow_probability.python.bijectors.conditional_bijector import ConditionalBijector
+from tensorflow_probability.python.bijectors import conditional_bijector
 
 
 
@@ -30,7 +30,7 @@ __all__ = [
 ]
 
 
-class RealNVP(ConditionalBijector):
+class RealNVP(conditional_bijector.ConditionalBijector):
   """RealNVP "affine coupling layer" for vector-valued events.
 
   Real NVP models a normalizing flow on a `D`-dimensional distribution via a
--- a/tensorflow_probability/python/bijectors/real_nvp.py
+++ b/tensorflow_probability/python/bijectors/real_nvp.py
@@ -235,9 +235,9 @@ def real_nvp_default_template(hidden_layers,
   created once. It takes the `d`-dimensional input x[0:d] and returns the `D-d`
   dimensional outputs `loc` ("mu") and `log_scale` ("alpha").
 
-  The default template does not support conditioning, and will raise an
-  exception if `condition_kwargs` are passed to it. To use conditioning in a
-  real nvp bijector, implement a conditioned scale/translation template that
+  The default template does not support conditioning and will raise an
+  exception if `condition_kwargs` are passed to it. To use conditioning in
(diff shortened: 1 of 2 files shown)


tensorflow_probability/python/bijectors/real_nvp.py
```diff
@@ -20,7 +20,7 @@
 
 import tensorflow as tf
 from tensorflow.python.layers import core as layers
-from tensorflow.python.ops.distributions import bijector
+from tensorflow_probability.python.bijectors.conditional_bijector import ConditionalBijector
```
srvasude: Prefer to import the module vs. the class:

from tensorflow_probability.python.bijectors import conditional_bijector

tensorflow_probability/python/bijectors/real_nvp_test.py
```diff
@@ -97,6 +97,46 @@ def testBatchedBijectorWithMLPTransform(self):
...
+          fldj_,
+      ] = sess.run([
+          forward_x,
+          inverse_y,
+          forward_inverse_y,
+          ildj,
+          fldj,
+      ])
+      self.assertEqual("real_nvp", nvp.name)
+      self.assertAllClose(forward_x_, forward_inverse_y_, rtol=1e-1, atol=0.)
```
srvasude: Can these tolerances be reduced? 1e-1 seems quite high.

tensorflow_probability/python/bijectors/real_nvp_test.py
```diff
@@ -97,6 +97,46 @@ def testBatchedBijectorWithMLPTransform(self):
...
 
+  def testBijectorConditionKwargs(self):
+    N = 3
+    x_ = np.arange(N * 4 * 2).astype(np.float32).reshape(N, 4 * 2)
+    conditions = {
+        'a': tf.reshape(tf.range(N * 4, dtype=tf.float32), (N, 4)),
+        'b': tf.reshape(tf.range(N * 2, dtype=tf.float32), (N, 2)),
+    }
+
+    with self.test_session() as sess:
```
srvasude: While the rest of this file uses this style, I'm trying to move the codebase to a newer style (which should hopefully make it easier for you as well!)

You can usually prefer self.evaluate(...) instead of sess.run(...). This allows you to also remove this self.test_session scope (and the tf.global_variables_initializer can be self.evaluate(tf.global_variables_initializer())

self.evaluate(...) underneath the hood uses/reuses a tf.Session().

tensorflow_probability/python/bijectors/real_nvp.py
hartikainen: Thanks!

tensorflow_probability/python/bijectors/real_nvp_test.py
hartikainen: Changed. It seems like I just blindly copied the tolerances over from the other tests in the same file. I'm curious why they have so loose tolerances?

tensorflow_probability/python/bijectors/real_nvp_test.py
hartikainen: Nice, I did not know about `self.evaluate`. Thanks!

hartikainen:
Thanks for the feedback @srvasude! Should be ready for another look now.

tensorflow_probability/python/bijectors/real_nvp.py
```diff
@@ -266,8 +266,11 @@ def real_nvp_default_template(hidden_layers,
 
   with tf.name_scope(name, "real_nvp_default_template"):
 
-    def _fn(x, output_units):
+    def _fn(x, output_units, **condition_kwargs):
       """Fully connected MLP parameterized via `real_nvp_template`."""
```
srvasude: Actually I have a quick question here:

I'm thinking that the default should actually do no conditioning. It's not clear to me that doing tf.concat([x] + [condition_kwargs]] is the right thing to do / pass in.  I.e:  I can imagine other equally sane ways to have a MLP conditioned on this. I might suggest reverting the change here. WDYT?

hartikainen: I agree that there might be other ways users might want to condition their networks. I thought providing some sane default here would be nice. After all, this still remains usable in the default form when the `condition_kwargs` is empty.

Happy to revert this if you think it's better to keep the default clean from the conditions.

srvasude: That's fair. Yeah I imagined it looked a little cleaner with out it, but I'm fine with the change as is. I'll start the process for pulling this in for an internal review, and see if anyone has any concerns with this (what will happen is this will go through an internal code review, and I will post back / notify you of changes /comments to make). Thanks again for this change!

csuter:
Note that we will need to rebase the commits in this PR, eliminating the merges with master. Apologies for not speaking up sooner, and for the lack of documentation of this wrinkle in our workflow...It has to do with the way we pull the PR into our internal system and push back out again -- the pull/push system gets confused by the presence of branch-and-merge sequences in the PR. It needs "fast-forward" style commits.

@hartikainen are you comfortable enough with git to do what I'm describing? If not I can generate and post here a sequence of commands that will accomplish what I'm requesting.

Thanks for your contribution!

hartikainen:
@csuter done. Thanks!

csuter:
Brilliant, thanks @hartikainen!

tensorflow_probability/python/bijectors/real_nvp_test.py
```diff
@@ -97,6 +97,45 @@ def testBatchedBijectorWithMLPTransform(self):
       self.assertAllClose(x_, inverse_y_, rtol=1e-1, atol=0.)
       self.assertAllClose(ildj_, -fldj_, rtol=1e-6, atol=0.)
 
+  def testBijectorConditionKwargs(self):
+    N = 3
```
srvasude: Call this batch_size? Our internal linter is complaining about this name.

tensorflow_probability/python/bijectors/real_nvp.py
```diff
@@ -266,8 +266,11 @@ def real_nvp_default_template(hidden_layers,
...
+      x = tf.concat(
+        [x] + [condition_kwargs[k] for k in sorted(condition_kwargs)], axis=-1)
```
srvasude: Can you add two spaces to this line.

tensorflow_probability/python/bijectors/real_nvp.py
```diff
@@ -20,7 +20,7 @@
 
 import tensorflow as tf
 from tensorflow.python.layers import core as layers
-from tensorflow.python.ops.distributions import bijector
+from tensorflow_probability.python.bijectors import conditional_bijector
```
srvasude: Can you move this one line up?

hartikainen:
@srvasude I made some syntax changes. Your guides helped but probably would've caused some other linter errors so I had to do some additional changes blindly. Where can I find the style configs you guys are using?

tensorflow_probability/python/bijectors/real_nvp_test.py
```diff
@@ -97,6 +97,45 @@ def testBatchedBijectorWithMLPTransform(self):
...
+    x_ = np.arange(N * 4 * 2).astype(np.float32).reshape(N, 4 * 2)
+    conditions = {
+        'a': tf.reshape(tf.range(N * 4, dtype=tf.float32), (N, 4)),
```
srvasude: Forgot about this one. Can you make this double quotations instead of single quotes (to be consistent with the rest of the file).

csuter:
I think the commands prescribed on the TF page should work for our code too: https://github.com/tensorflow/tensorflow/blob/master/CONTRIBUTING.md#python-coding-style

I haven't tried it though.

srvasude:
Good question. @cgs: I believe we're also trying to get the linting tools to run for contributors, so they don't have to go through this back and forth as well, correct?

hartikainen:
I ran the linter with the instructions from @csuter's link and the latest changes pass. Thanks all!

tensorflow_probability/python/bijectors/real_nvp_test.py
```diff
@@ -98,11 +98,16 @@ def testBatchedBijectorWithMLPTransform(self):
...
   def testBijectorConditionKwargs(self):
-    N = 3
-    x_ = np.arange(N * 4 * 2).astype(np.float32).reshape(N, 4 * 2)
+    batch_size = 3
+    x_ = np.arange(batch_size * 4 * 2).astype(np.float32).reshape(
+        batch_size, 4 * 2)
     conditions = {
-        'a': tf.reshape(tf.range(N * 4, dtype=tf.float32), (N, 4)),
-        'b': tf.reshape(tf.range(N * 2, dtype=tf.float32), (N, 2)),
+        'a':
```
srvasude: Sorry to ping again, but can you change these quotations (this line and below) to be double quotes instead of single quotes? I guess the linter doesn't catch that (but the internal one checks if the whole file has the same quotations)

cgs:
> That...'s something I'll avoid from now on.

Appreciate that! ;)

tensorflow_probability/python/bijectors/real_nvp_test.py
```diff
@@ -97,6 +97,50 @@ def testBatchedBijectorWithMLPTransform(self):
...
+        fldj_,
+    ] = self.evaluate([
+        forward_x,
+        inverse_y,
+        forward_inverse_y,
+        ildj,
+        fldj,
+    ])
+    self.assertEqual("real_nvp", nvp.name)
+    self.assertAllClose(forward_x_, forward_inverse_y_, rtol=1e-6)
```
srvasude: Actually it seems like the tests are failing on our internal system...
To unblock you, can you change these to be:

      self.assertAllClose(forward_x_, forward_inverse_y_, rtol=1e-1, atol=0.)	      
      self.assertAllClose(x_, inverse_y_, rtol=1e-1, atol=0.)
      self.assertAllClose(ildj_, -fldj_, rtol=1e-6, atol=0.)

I'm going to file a bug internally to investigate where this numerical instability comes from, and then we can fix it wholesale for these tests (and you don't have to worry about it).

hartikainen: Changed.

tensorflow_probability/python/bijectors/real_nvp.py
srvasude: From talking internally, it seems like I think we might prefer to actually throw an exception for conditioning. concat'ing makes a fairly strong assumption that the condition_kwargs are all Tensors / Tensor-like. This might break when a user passes in a boolean condition_kwarg for instance. Also, in this case if they are tensors, then the user could also concat this to x and pass it in.

Can you instead raise NotImplementedError('Conditioning not implemented in default template.') or something of that nature?

Then in the test you can implement some version of this, and check if it works.

Thanks again!

hartikainen: Makes a lot of sense, I now do I agree that maybe having this be the default is not the best option.

One thing we could do is have a `condition_fn` argument for the template, which the user could pass to customize the condition handling. Something like this:

```python
def real_nvp_default_template(hidden_layers,
                              shift_only=False,
                              activation=tf.nn.relu,
                              condition_fn=None,
                              name=None,
                              *args,  # pylint: disable=keyword-arg-before-vararg
                              **kwargs):
  with tf.name_scope(name, "real_nvp_default_template"):

    def _fn(x, output_units, **condition_kwargs):
      """Fully connected MLP parameterized via `real_nvp_template`."""
      if condition_kwargs and condition_fn is None:
        raise NotImplementedError(
            "Conditioning not implemented in the default template.")

      if condition_fn is not None:
        x = condition_fn(x, condition_kwargs)

      ...
```

And then the usage would be something like:
```python
def my_condition_fn(x, condition_kwargs):
    x = tf.concat(
        [x] + [condition_kwargs[k] for k in sorted(condition_kwargs)],
        axis=-1)
    return x

template = real_nvp_default_template(
    hidden_layers=[3],
    shift_only=False,
    condition_fn=my_condition_fn)
```

Just a thought. It would provide an easy way and an example of using condition. On the other hand, it might actually be too complicated considering the simplicity of the template function.

srvasude: Yeah I think preferring simplicity here would be what we want. I can imagine also more complicated forms of conditioning (e.g. the conditions come up in the network construction, and make more sense when building the layers, then when doing input preprocessing) could come up, and I guess this could confuse those sorts of users.

Perhaps the right thing to do here is to raise the NotImplementedError, but add in a docstring a snippet of how to do conditioning? What do you think?

hartikainen: Yeah, you're totally right. I removed conditioning from the default template and added a small note to the docstring. Also changed tests accordingly.

srvasude:
Sorry for the delay. The tests seem to fail now. Do you mind taking a look?

tensorflow_probability/python/bijectors/real_nvp_test.py
```diff
@@ -97,6 +98,67 @@ def testBatchedBijectorWithMLPTransform(self):
       self.assertAllClose(x_, inverse_y_, rtol=1e-1, atol=0.)
       self.assertAllClose(ildj_, -fldj_, rtol=1e-6, atol=0.)
 
+  def testBijectorConditionKwargs(self):
+    batch_size = 3
+    x_ = np.arange(batch_size * 4 * 2).astype(np.float64).reshape(
```
hartikainen: @srvasude I had to change these to float64 in order to make the tests pass. Otherwise I would get the weird precision errors here:
```python
AssertionError:
Not equal to tolerance rtol=0.1, atol=0
Mismatched value: a is different from b.
(mismatch 4.166666666666671%)
 x: array([[ 0.,  1.,  2.,  3.,  4.,  5.,  6.,  7.],
       [ 8.,  9., 10., 11., 12., 13., 14., 15.],
       [16., 17., 18., 19., 20., 21., 22., 23.]], dtype=float32)
 y: array([[ 0.      ,  1.      ,  2.      ,  3.      ,  4.      ,  5.      ,
         5.999999,  7.      ],
       [ 8.      ,  9.      , 10.      , 11.      , 12.      , 13.000007,...

----------------------------------------------------------------------
Ran 18 tests in 6.292s
```

hartikainen:
@srvasude the tests should be fixed now. Can you take a look at my inline comment about the precision error?

tensorflow_probability/python/bijectors/real_nvp_test.py
csuter: why are we using atol = 0? seems unduly strict for floating point comparison...

hartikainen: I happened to copy the tolerances from the tests above. The problem is though, that even when increasing this tolerance, my test case fails when not using float64. There are some errors on the relative scale of 50%.

csuter:
Hey @hartikainen, I figured out the issue, and am pushing through a change shortly that changes the existing tests. Once that's pushed, if you rebase I can send a PR to your fork that makes the same change to the tests you've added.

The issue is that all these tests are using `np.arange` (or `tf.range`) to generate test inputs, many of which fall well outside the region in which RealNVP is nicely behaved. Typically, the inputs to RealNVP are coming from a `Normal(0, 1)`, so values in, eg, `range(0, 3 * 4 * 2)`, which eventually get sent through an exp, become more likely to see large floating point discrepancies upon inverse transform.

My change (currently under review internally) switches these to, eg, `np.random.uniform(0, 1, (3, 4 * 2))`, which seems to allow us to use much tighter tolerance (`rtol=1e-4`).

My change should be in momentarily, at which point, if you'll rebase, I can send you a patch. Stay tuned!

hartikainen:
Thanks @csuter, this sounds great! I'll do the rebase later today or tomorrow. It might be nicer to `linspace(0, 1)` instead of `uniform(0, 1)`, to not have to worry about any stochasticity in the tests. Just a thought, I don't know if that matters too much.

csuter:
Yeah, we may add a seed later but I ran the tests ~200 times with the new settings and 100% passed, so it should be alright.

The change is in now, so feel free to rebase. Then I can send the change, unless you want to make it yourself (it's quite simple once the other changes have bee pulled in). Just let me know!

Thanks again for all the iteration with us!

hartikainen:
@csuter, thanks! I just rebased and added couple of commits to fix the tolerance problems in the test cases. I ended up using linspace for the inputs, hopefully that's fine. Let me know if there's anything else that should be changed.

Also, I'm not quite sure what you mean by sending the patch. Happy to do that if you point me to the right direction with that.

tensorflow_probability/python/bijectors/real_nvp_test.py
```diff
@@ -97,6 +98,64 @@ def testBatchedBijectorWithMLPTransform(self):
...
+    batch_size = 3
+    x_ = np.linspace(-1.0, 1.0, (batch_size * 4 * 2)).astype(
+        np.float32).reshape((batch_size, 4 * 2))
+
+    conditions = {
+        "a": tf.random_normal((batch_size, 4), dtype=tf.float32),
+        "b": tf.random_normal((batch_size, 2), dtype=tf.float32),
+    }
+
+    def _condition_shift_and_log_scale_fn(x0, output_units, **condition_kwargs):
```
csuter: fwiw, here you could replace condition_kwargs with just the named args, a and b:

```
def _condition_shift_and_log_scale_fn(x0, output_units, a, b):
   x = tf.concat([x0, a, b], axis=-1)
...
```

csuter:
@hartikainen I left one more very minor comment, but this looks great now. Thanks again for iterating. We can go ahead and pull it in once you've had a chance to respond to my comment.

tensorflow_probability/python/bijectors/real_nvp_test.py
hartikainen: Noted. I changed this and also the one below in the `RealNVPConstantShiftScaleTest` which didn't require `condition_kwargs` at all.

Thanks!

csuter:
@srvasude can you update the CL with the latest state, and see if this is ready to submit?

@hartikainen thanks again for your contribution!

hartikainen:
Thanks @srvasude and @csuter!

PR #120, opened 2018-08-07
fixed parameter check when using analytic kl

a-arbabi:


a-arbabi:
I am currently an intern in Google and covered for CLA.

PR #123, opened 2018-08-08
Fixed examples\vae.py parsing on py3

m-colombo:
In py3 `"1" != b"1"`, thus all the images are fully black.


davmre:
Good catch; unfortunately our unit tests just verify that the examples run, not that they get good results on every platform, so this was definitely an oversight. :-) Thanks @m-colombo for the fix!

PR #137, opened 2018-08-26
Add an example notebook for Bayesian Switchpoint Analysis

Feh:
This is re-doing one of the standard examples in the pymc3
documentation; it might help people who are familiar with that system
and want to translate their problem to Edward2.

(You can see a rendering of the notebook here: https://github.com/Feh/probability/blob/switchpoint-analysis/tensorflow_probability/examples/jupyter_notebooks/Bayesian_Switchpoint_Analysis.ipynb)


dustinvtran:
This is great! Thanks @Feh for sharing. This looks good to me aside from minor comments. I'm adding comments here as commenting on a notebook's JSON is difficult.

Can another TensorFlow Probability developer also take a look and review? After the second approval, I can send the pull request to our internal system and have it submitted, which will reflect in the Github repository with credit to you.

> model description

It's unrelated to this PR, but I wonder what happens if you use the continuous Uniform and just apply a `tf.where` to switch between `e` and `l` for the rate. I don't know if there are any guarantees for HMC in this setting, but it may still work.

> _model

It's good practice to use the model's actual name, e.g., `disaster_count_model`.

> However, $r_t$ has a discontinuity at $s$,

Can you describe what each variable represents here?

>  # constant (i.e. observed values)

Instead of writing this, add a docstring describing what the function is. This avoids needing an inline comment.

Feh:
Fixed the two nitpicks in the amended commit, thanks.

> It's unrelated to this PR, but I wonder what happens if you use the continuous Uniform and just apply a tf.where to switch between e and l for the rate. I don't know if there are any guarantees for HMC in this setting, but it may still work.

Thanks for the hint! In the early stage of playing with this I had actually tried something based on `tf.cond` but couldn’t get it to work. Sometimes `ed.Poisson` complained about the wrong type of input argument, sometimes the model building complained about a `None` gradient.

I just tried it out, and using `[tf.where(x < switchpoint, early_disaster_rate, late_disaster_rate) for x in y]` does work just fine, and the HMC does seem to converge. The switchpoint results however look less normal (which seems reasonable given that a sigmoid switch is relatively “smooth”). If I change the sigmoid switch to `tf.sigmoid(100*(tf.to_float(y)-switchpoint))` the posterior distributions for the switchpoint look the same like with a `tf.where`.

Perhaps it’d be interesting to rewrite the notebook to propose the original and modified model, and have the results displayed side by side for comparison?

Feh:
I went ahead and created a new notebook with both models side by side. Tell me what you think?

csuter:
Hi @Feh, thanks for this great writeup! I have almost no critical feedback to give.

One adjustment I'd like to consider is giving separate names to the two log_joint functions, instead of embedding them in a list and using indexing to refer to them further on. It would help emphasize their distinctness, hopefully clarify the code a bit, and possibly avoid refactoring bugs if orders get changed. If we rearrange things this way, we'd probably just want to pass the log joint callable into make_chain.

I'm basically happy to approve after consideration of this tweak. Thanks again!

davmre:
Thanks @Feh, this is a really nice contribution! Just a couple of thoughts on top of Chris and Dustin's:

- It'd be nice to explicitly note somewhere that HMC doesn't formally work on discrete variables, so although it's cool that the "switch" model seems to work, in general there are no guarantees for writing a model in this form. (I'm actually kind of mystified how it's getting any gradient at all on the switchpoint -- I guess it probably isn't, and is just following the initial random momentum in each HMC step to essentially get some kind of discrete MH proposal?) The approach we'd recommend to users in general is to look for an explicitly continuous model parameterization, as in the sigmoid model.

- The rates you're doing inference over are positive-valued, but HMC doesn't know this, so it might explore negative values potentially leading to undefined behavior. In general it's good practice to explicitly constrain HMC to only explore valid regions of the space. The Bayesian GMM notebook (https://github.com/tensorflow/probability/blob/master/tensorflow_probability/examples/jupyter_notebooks/Bayesian_Gaussian_Mixture_Model.ipynb) has an example of doing this -- basically you create a `tfp.mcmc.TransformedTransitionKernel` wrapping your HMC kernel, and pass in a list of bijectors that transform unconstrained real values to your constrained space of choice. In this case that'd likely be `bijectors=[tfb.Sigmoid(), tfb.Softplus(), tfb.Softplus()]`, where the sigmoid constrains the switchpoint to the unit interval (you'd need to slightly modify the model to define the switchpoint on [0,1] instead of [0, num_years]) and the softplus constrains the rates to the positive real line.

derifatives:
Some comments on the CL:

========================================================================
tensorflow_probability/examples/jupyter_notebooks/Bayesian_Switchpoint_Analysis.ipynb&v=s1
------------------------------------
Line 14:
This notebook is reimplementing the Bayesian “Change point analysis” example from the [pymc3 documentation](https://docs.pymc.io/notebooks/getting_started.html#Case-study-2:-Coal-mining-disasters). I suggest you read their example description before proceeding here.

This reads a little snarky. Maybe s/I suggest/you suggest/ and s/here//
------------------------------------
Line 14:
This notebook is reimplementing the Bayesian “Change point analysis” example from the [pymc3 documentation](https://docs.pymc.io/notebooks/getting_started.html#Case-study-2:-Coal-mining-disasters). I suggest you read their example description before proceeding here.

s/is reimplementing/reimplements and extends/, since you add a second model not in the original (the sigmoid switch).
------------------------------------
Line 32:
The dataset is from [here](https://pymc-devs.github.io/pymc/tutorial.html#two-types-of-variables). Note, there is another version of this example [floating around](https://docs.pymc.io/notebooks/getting_started.html#Case-study-2:-Coal-mining-disasters), but it has “missing” data – in which case you’d need to imputate missing values. (Otherwise your model will not ever leave its initial parameters because the likelihood function will be undefined.)

Nit: s/imputate/impute/
------------------------------------
Line 54:
The actual disaster count is fixed (observed); any solution of this model will need to specify both the switchpoint and the “early” and “late” rate of disasters.

Is it common to refer to "solutions"? I'd imagine s/any solution of/a sample from/
------------------------------------
Line 68:
However, the mean disaster rate $r_t$ has a discontinuity at the switchpoint $s$, which makes it not differentiable. Differentiability might be desirable for some solvers, though it is not required in this case.

I'm not sure what's meant by a "solver".
------------------------------------
Line 70:
For model two, we modify the original model using a [sigmoid “switch”](https://en.wikipedia.org/wiki/Sigmoid_function) between *e* and *l* to make the transition differentiable, and use a continuous uniform distribution for the switchpoint $s$.  (One could argue this model is more true to reality, as a “switch” in mean rate would likely be stretched out over multiple years.) The new model is thus:

Maybe s/For model two/As a second model/
------------------------------------
Line 70:
For model two, we modify the original model using a [sigmoid “switch”](https://en.wikipedia.org/wiki/Sigmoid_function) between *e* and *l* to make the transition differentiable, and use a continuous uniform distribution for the switchpoint $s$.  (One could argue this model is more true to reality, as a “switch” in mean rate would likely be stretched out over multiple years.) The new model is thus:

Maybe s/we modify the original model using/we use
------------------------------------
Line 87:
  switchpoint = ed.Uniform(low=0., high=tf.to_float(len(years)),

I thought the switching model had a discrete uniform? (I think you can finesse this, but it can be made a little clearer.)
------------------------------------
Line 100:
                           name='switchpoint')

Maybe explain why you don't get into trouble using the same names in both models?
------------------------------------
Line 110:
log_joint = [ed.make_log_joint_fn(disaster_count_model_switch),

Is this definitely what we want to recommend to our users? I think line 125 is pretty ugly, and semantically it now feels like we're defining a single log_prob over both models simultaneously, but really we're bundling two log_probs into one function. I think this could either use more discussion or possibly just do two separate log_joints? I don't insist, I'd like to hear thoughts.
------------------------------------
Line 129:
1. Define the model via Edward distributions. The `disaster_rate_*` functions are called with an array of `[0, ..., len(years)-1]` to produce a vector of `len(years)` random variables – the years before the `switchpoint` are (essentially) `early_disaster_rate`, the ones after (essentially) `late_disaster_rate`.

I don't think you have `disaster_rate_*` in the current code; rephrase?
------------------------------------
Line 129:
1. Define the model via Edward distributions. The `disaster_rate_*` functions are called with an array of `[0, ..., len(years)-1]` to produce a vector of `len(years)` random variables – the years before the `switchpoint` are (essentially) `early_disaster_rate`, the ones after (essentially) `late_disaster_rate`.

The "essentially"'s are referring to the sigmoid switch? I might delete those words and say "(modulo the sigmoid switch)" at the end of the sentence. Your call.
------------------------------------
Line 130:
1. Make Edward compute the logprob function for the joint probability distribution. `log_joint` then needs to be called with the four named parameters in the model and outputs the log probability of that outcome. (See below for an example.)

Nit: Consider s/Make Edward compute the/Define the/
------------------------------------
Line 133:
Here is a sanity-check that the target log prob function is useful:

Consider s/useful/sane/
------------------------------------
Line 138:
  print([x.eval() for x in target_log_prob_fn(-10., 1., 1.)]) # Impossible result

Nit: Case doesn't agree with other comments immediately above.
------------------------------------
Line 193:
## Visualize the result

AFAICT the two models are giving equivalentish results? Why bother having both models?
========================================================================

Feh:
Thanks for all the reviews! I’ve changed the code in an amended commit to be more clear that these are two models and they have distinct joint probability functions; I’ve also addressed the grammar nits by reformulating.

Some questions about your comments, @davmre:

1. “It'd be nice to explicitly note somewhere that HMC doesn't formally work on discrete variables” – intuitively this makes sense; however, piecewise differentiable with a single discontinuity != discrete. So I’m unsure if it should say “HMC can’t really be relied on to work but it does in this case” or something else? [This paper on HMC convergence conditions](https://arxiv.org/pdf/1705.00166.pdf) in section 3 makes the fundamental assumption that the potential is continuously differentiable; perhaps cite this?
1. Related to the point above, Poisson distributions are discrete; does your statement imply that the whole model ought not work, even with a sigmoid transition between the two mean rates?
1. “The rates you're doing inference over are positive-valued, but HMC doesn't know this, so it might explore negative values potentially leading to undefined behavior.” – Interesting. I would have thought that for both the Poisson and uniform distributions, outside of the probability mass the pmfs are zero and that would not be a problem? I’d have to read up on bijectors before implementing this. And instead of constraining the switchpoint to [0;1], is it possible to simply transform ℝ → [e, l] instead?

(To set expectations, I’ll be travelling the next few days and might not find the time to get back to this before returning.)

davmre:
Ah, sorry I was being a bit lazy. By "HMC doesn't work with discrete variables" I mean, HMC relies on gradients wrt the latent variables you're doing inference over. It's fine to have discrete *observations*, as given by the Poisson likelihood, since they ultimately define a posterior over continuous latent variables (the Poisson rates) of which you can take gradients, etc. 

In the switchpoint model, the switchpoint is effectively discrete, because its gradient is zero almost everywhere (in the technical sense) except at the integers where the function is nondifferentiable. That's certainly not a standard regime in which you'd expect HMC to work; e.g., the original PyMC3 writeup concluded, "Unfortunately because they are discrete variables and thus have no meaningful gradient, we cannot use NUTS for sampling switchpoint or the missing disaster observations." That said, the way you've written the model with a continuous Uniform prior, the switchpoint is technically speaking a continuous variable, and it seems to work because HMC resorts to random-walk behavior in the absence of gradients. My high-level point is that this is weird enough to deserve some explanation. On reflection I think something like "although the model with the discrete switch provides no gradient signal to the switchpoint, HMC's random-walk behavior appears sufficient to explore the posterior" is a better explanation than "we don't expect this to work in general": the right answer is we *do* expect it to work whenever the posterior is sufficiently low-dimensional that a random walk can effectively find the areas of high mass.

Re bijectors, in general it's good practice to do inference in an unconstrained space rather than relying on hard constraints from the prior. Hard constraints mean that any proposal outside of the bounds will be rejected; using bijectors to reparameterize generally means you can take larger steps since it's impossible to go out of bounds. I'm also not confident that all TF distributions guarantee the behavior you'd want on invalid inputs -- it looks like the Uniform distribution gives log probs of `-inf` outside of its bounds, which properly implements a hard constraint, while the Exponential distribution gives `NaN`, which likely will propagate (if you've ever seen NaNs running inference on this model, this would likely be why). I think it's definitely worth learning to run models in unconstrained spaces, and you'd likely get better-behaved inference and set a better example for other users. That said, this is in the spirit of a suggestion rather than demand. :-)

axch:
@davmre, @derifatives, or @csuter: Could one of you please clarify the status of this PR?  I'm not sure I can infer it from the thread.  Is it ready to be merged as-is, or are we waiting for @Feh to change something?  If the latter, could I ask one of you to write a review with a clear set of requests?

Feh:
@axch I plan to incorporate @davmre’s suggestion of using bijectors to limit the domain to search over, and would like another review then. I was a bit busy – sorry for letting it linger so long!

axch:
@Feh No problem!  I just wanted to make sure we hadn't dropped the ball on our end.

Feh:
Sorry I’ve let this slide so long.

@davmre I added an explanation why we think HMC would work, but could you please expand on how I’d use Bijectors to make the model explore the space in a better way?

I’ve played around with various things but nothing seems to make an improvement (most of what I try produces garbage results). The model in its current form:

```python
def disaster_count_model_sigmoid():
  early_disaster_rate = ed.Exponential(rate=1., name='early_disaster_rate')
  late_disaster_rate = ed.Exponential(rate=1., name='late_disaster_rate')
  switchpoint = ed.Uniform(low=0., high=tf.to_float(len(years)),
                           name='switchpoint')
  def disaster_rate(ys):
    return (early_disaster_rate +
            tf.sigmoid((tf.to_float(ys)-switchpoint)) *
            (late_disaster_rate - early_disaster_rate))
  disaster_count = ed.Poisson(rate=disaster_rate(np.arange(len(years))),
                              name='disaster_count')
  return disaster_count
```

I believe your concerns are around the paramaters outside the model space: E.g. what happens if we have negative rates, or the switchpoint is outside of the years we’re looking at.

```python
with tf.Session():
  print(tfd.Exponential(rate=1.0).log_prob([-100, -10, -1, 0, 1, 10, 100]).eval())
  print(tfd.Poisson(rate=-1).log_prob([-100, -10, -1, 0, 1, 10, 100]).eval())
# Yields:
# [ 100.   10.    1.    0.   -1.  -10. -100.]
# [-inf -inf -inf  nan  nan  nan  nan]
```

Clearly, a log prob that’s NaN is not useful, and log probs > 0 are impossible. So we want to constrain away from negative rates for the Poisson distribution. This can be achieved by making sure that `{early,late}_disaster_rate` are always >= 0.

So we could change the following:

```python
  early_disaster_rate = ed.Exponential(rate=1., name='early_disaster_rate')
```

To

```python
  early_disaster_rate = ed.TransformedDistribution(
      distribution=tfd.Exponential(rate=1.),
      bijector=tfb.Softplus(), name='early_disaster_rate')
```

However, then: `early_disaster_rate.log_prob([-100, -10, -1, 0, 1, 10, 100])` yields: `[nan, nan, nan, inf, -8.264971e-02, -9.999908e+00, -1.000000e+02]`. That’s no good either – I’d have expected `-inf` probabilities for negative values for this to be useful.

Can you be more explicit whare I would use the bejctor? Thanks!

davmre:
Hi @Feh! I think your instincts are right, but you got mildly flipped around on the implementation. To transform a positive-valued rate into an unconstrained value, you need the inverse of a Softplus bijector (the plain softplus goes the other way). If you were writing this explicitly, you'd want to define the latent variable in an unconstrained space using an inverse softplus (so that inference 'sees' the unconstrained value), then project back into constrained space to get the actual rate:

```python
# Define the latent variable in unconstrained space for inference convenience
 early_disaster_rate_unconstrained = ed.TransformedDistribution(
      distribution=tfd.Exponential(rate=1.),
      bijector=tfb.Invert(tfb.Softplus()), name='early_disaster_rate_unconstrained')
early_disaster_rate = tfb.Softplus().forward(early_disaster_rate_unconstrained)
```

This is pretty annoying, so there are a few ways around writing out the transformation explicitly. For MCMC/HMC inference, the most natural is to wrap your inference kernel with [tfp.mcmc.TransformedTransitionKernel](https://www.tensorflow.org/probability/api_docs/python/tfp/mcmc/TransformedTransitionKernel). This lets you keep your model as-is; you just supply a list of bijectors and it essentially performs the above transformation behind the scenes (it applies the inverse for you, so you'd just provide bijectors in the constraining direction, e.g., in this case `tfb.Softplus()` for the rates and something like `tfb.Chain([tfb.AffineScalar(scale=tf.to_float(len(years))), tfb.Sigmoid()])` for the switchpoint). See the linked docs or the [Bayesian GMM](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/examples/jupyter_notebooks/Bayesian_Gaussian_Mixture_Model.ipynb) example for an example. (your notebook might be an even cleaner example once we get it checked in!) 

I can't guarantee that using `TransformedTransitionKernel` will significantly improve inference for this model, but it certainly shouldn't hurt, and it does represent the best practice for dealing with constrained variables.

Feh:
Thanks @davmre, that was a very helpful and clear explanition – it was not immediately obvious to me which way the bijectors would work. I went with the `TransformedTransitionKernel` approach because it fits much nicer with the narrative of how to construct the code (the model code in Python resembles the formulas, and it comes into play only upon evaluation).

I squashed all the commits together and re-uploaded the new ipynb file. PTAL at [the result](https://github.com/Feh/probability/blob/switchpoint-analysis/tensorflow_probability/examples/jupyter_notebooks/Bayesian_Switchpoint_Analysis.ipynb). Thanks!

davmre:
This looks great to me. Nice job!

One nit: the step size adaptation performed by `make_simple_step_size_update_policy` doesn't in general preserve the stationary distribution of the MCMC chain, so you should limit adaptation to the burnin period:
`make_simple_step_size_update_policy(num_adaptation_steps=num_burnin_steps * 0.8)`
(the `* 0.8` is there to exercise the nit-picky point that one should in principle give the chain some time to reach the stationary distribution after adaptation ends before you start drawing samples). Can you make this change?

Other than that this looks basically ready to merge to me. Thanks for sticking with this for so long! @csuter are you still happy/able to approve this? (I know you're on vacation so might respond until next week).

jvdillon:
Possibly relevant: "Example: Inferring behaviour from text-message data"
From: https://github.com/CamDavidsonPilon/Probabilistic-Programming-and-Bayesian-Methods-for-Hackers/blob/master/Chapter1_Introduction/Ch1_Introduction_TFP.ipynb

(Apologies in advance if this example is not helpful or relevant.)

Feh:
@jvdillon – thanks for mentioning it! I had actually read the book, but before [this commit](https://github.com/CamDavidsonPilon/Probabilistic-Programming-and-Bayesian-Methods-for-Hackers/commit/d85ad358cc8369a696ffd59d54aa615b33af6361#diff-0f5cef320cc933d6c9ceb2f9d5cd81bd) which apparently pivoted the book towards TFP.

@davmre – I made the changes (slightly more involved as I’m running two models in parallel, so each needs a different step counter), PTAL.

Feh:
Friendly ping?

PR #155, opened 2018-09-05
Example for CIFAR-10 with ResNet18 and VGG16

jmzeng:
This is an example for CIFAR-10 with ResNet18 and VGG16. With the current setup, the ResNet18 hits about 80% validation accuracy and 90% training accuracy.

The visualization for MNIST has been removed because the plots did not really look as I expected. The validation accuracy would be pretty high but the distributions plotted are still like a random dist. I'm unsure as to why that is happening so I removed it for now. If you would like it for debugging, I can include it. 

Please let me know if there are style changes needed.

### DIFF
--- a/tensorflow_probability/examples/cifar10_bnn.py
+++ b/tensorflow_probability/examples/cifar10_bnn.py
@@ -19,13 +19,13 @@ The architecture can be either ResNet [1] or VGG [2].
 #### References
 
 [1]: He, Kaiming, Xiangyu Zhang, Shaoqing Ren, and Jian Sun.
-    "Deep residual learning for image recognition."
+     "Deep residual learning for image recognition."
      _Proceedings of the IEEE_, 2016.
      https://arxiv.org/abs/1512.03385
 [2]: Simonyan, Karen, and Andrew Zisserman.
-    "Very deep convolutional networks for large-scale image recognition."
-    arXiv preprint arXiv:1409.1556 (2014).
-    https://arxiv.org/pdf/1409.1556.pdf
+     "Very deep convolutional networks for large-scale image recognition."
+     arXiv preprint arXiv:1409.1556 (2014).
+     https://arxiv.org/pdf/1409.1556.pdf
 """
 
 from __future__ import absolute_import
@@ -82,16 +82,19 @@ flags.DEFINE_string("architecture",
                     default="resnet",
                     help="Network architecture to use.")
 flags.DEFINE_integer(
-    "kernel_scale_mean",
+    "kernel_posterior_scale_mean",
     default=-9,
     help="Initial kernel posterior mean of the scale (log var) for q(w)")
 flags.DEFINE_float(
-    "kernel_scale_constraint",
+    "kernel_posterior_scale_constraint",
     default=0.2,
     help="Posterior kernel constraint for the scale (log var) of q(w).")
 flags.DEFINE_integer("kl_annealing",
-                     default=100,
+                     default=50,
                      help="Epochs to anneal the KL term (anneals from 0 to 1)")
+flags.DEFINE_boolean("subtract_pixel_mean", 
+                     default=True,
+                     help="Boolean for normalizing the images")
 
 FLAGS = flags.FLAGS
 
@@ -108,6 +111,11 @@ def build_input_pipeline(x_train, x_test, y_train, y_test,
 
   y_train = y_train.flatten()
   y_test = y_test.flatten()
+  
+  if FLAGS.subtract_pixel_mean:
+    x_train_mean = np.mean(x_train, axis=0)
+    x_train -= x_train_mean
+    x_test -= x_train_mean
 
   print('x_train shape:' + str(x_train.shape))
   print(str(x_train.shape[0]) + ' train samples')
@@ -158,10 +166,11 @@ def main(argv):
   else:
     model_fn = bayesian_vgg
   
-  model = model_fn(IMAGE_SHAPE,
-                   num_classes=10,
-                   kernel_posterior_scale_mean=FLAGS.kernel_scale_mean,
-                   kernel_posterior_scale_constraint=FLAGS.kernel_scale_constraint)
+  model = model_fn(
+      IMAGE_SHAPE,
+      num_classes=10,
+      kernel_posterior_scale_mean=FLAGS.kernel_posterior_scale_mean,
+      kernel_posterior_scale_constraint=FLAGS.kernel_posterior_scale_constraint)
   logits = model(images)
   labels_distribution = tfd.Categorical(logits=logits)
 
@@ -170,7 +179,8 @@ def main(argv):
   t = tf.Variable(0.0)
   kl_regularizer = t / (FLAGS.kl_annealing * len(x_train) / FLAGS.batch_size)
 
-  # Compute the -ELBO as the loss, averaged over the batch size.
+  # Compute the -ELBO as the loss. The kl term is annealed from 0 to 1 over 
+  # the epochs specified by the kl_annealing flag.
   log_likelihood = labels_distribution.log_prob(labels)
   neg_log_likelihood = -tf.reduce_mean(log_likelihood)
   kl = sum(model.losses) / len(x_train) * tf.minimum(1.0, kl_regularizer)
--- a/tensorflow_probability/examples/cifar10_bnn.py
+++ b/tensorflow_probability/examples/cifar10_bnn.py
@@ -84,7 +84,7 @@ flags.DEFINE_integer("num_monte_carlo",
                      default=50,
                      help="Network draws to compute predictive probabilities.")
 flags.DEFINE_string("arch",
-                    default="vgg",
+                    default="resnet",
                     help="Network architecture to use.")
 flags.DEFINE_integer(
     "kernel_mean",
--- a/tensorflow_probability/examples/cifar10_bnn.py
+++ b/tensorflow_probability/examples/cifar10_bnn.py
@@ -22,7 +22,6 @@ The architecture can be either ResNet [1] or VGG [2].
     "Deep residual learning for image recognition."
      _Proceedings of the IEEE_, 2016.
      https://arxiv.org/abs/1512.03385
-
 [2]: Simonyan, Karen, and Andrew Zisserman.
     "Very deep convolutional networks for large-scale image recognition."
     arXiv preprint arXiv:1409.1556 (2014).
@@ -44,8 +43,8 @@ import tensorflow as tf
 
 from tensorflow.python.keras.datasets import cifar10
 
-from models.BResNet import BayesianResNet
-from models.BVGG import BayesianVGG
+from models.bayesian_resnet import bayesian_resnet
+from models.bayesian_vgg import bayesian_vgg
 
 import warnings
 warnings.simplefilter(action='ignore')
@@ -64,14 +63,10 @@ flags.DEFINE_integer("epochs",
 flags.DEFINE_integer("batch_size",
                      default=128,
                      help="Batch size.")
-flags.DEFINE_string(
-    "data_dir",
-    default=os.path.join(
-        os.getenv(
-            "TEST_TMPDIR",
-            "/tmp"),
-        "bayesian_neural_network/data"),
-    help="Directory where data is stored (if using real data).")
+flags.DEFINE_string("data_dir",
+                    default=os.path.join(os.getenv("TEST_TMPDIR", "/tmp"),
+                                         "bayesian_neural_network/data"),
+                    help="Directory where data is stored (if using real data).")
 flags.DEFINE_string(
     "model_dir",
     default=os.path.join(os.getenv("TEST_TMPDIR", "/tmp"),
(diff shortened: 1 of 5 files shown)


tensorflow_probability/examples/cifar10_bnn.py
```diff
@@ -0,0 +1,246 @@
...
+                     help="Initial posterior mean of the log variance for q(w)")
+flags.DEFINE_float("variance_threshold",
+                     default=0.2,
+                     help="Posterior variance threshold constraint. Log var <= log(threshold).")
+flags.DEFINE_integer("kl_annealing",
+                     default=100,
+                     help="Epochs to anneal the KL term (anneals from 0 to 1)")
+
+FLAGS = flags.FLAGS
+
```
dustinvtran: You should run `pep8`. For example, you need two blanklines before after every top-level function def.

tensorflow_probability/examples/cifar10_bnn.py
```diff
@@ -0,0 +1,246 @@
...
+def main(argv):
+  del argv  # unused
+  if tf.gfile.Exists(FLAGS.model_dir):
+    tf.logging.warning(
+        "Warning: deleting old log directory at {}".format(FLAGS.model_dir))
+    tf.gfile.DeleteRecursively(FLAGS.model_dir)
+  tf.gfile.MakeDirs(FLAGS.model_dir)
+
+  # Load training and testing data
```
dustinvtran: I recommend removing comments like this, "Set training steps" below, etc. This puts more pressure on the code to be self-documenting. At the moment, it seems like it is.

tensorflow_probability/examples/cifar10_bnn.py
```diff
@@ -0,0 +1,246 @@
...
+    # pass of the probabilistic layers. They are cheap but noisy predictions.
+    predictions = tf.argmax(logits, axis=1)
+    with tf.name_scope("train"):
+        train_accuracy, train_accuracy_update_op = tf.metrics.accuracy(
+                labels=labels, predictions=predictions)
+    with tf.name_scope("valid"):
+        valid_accuracy, valid_accuracy_update_op = tf.metrics.accuracy(
+                labels=labels, predictions=predictions)
+
+    with tf.name_scope("training"):
```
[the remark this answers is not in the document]
dustinvtran: See https://github.com/tensorflow/probability/commit/e00ab7df4b3865054a39f8af64e85e31123aba5c.

jmzeng:
Updated with the suggested style changes. I decided to keep the functional format for the models and instead turned them into classes.

tensorflow_probability/examples/models/BResNet.py
```diff
@@ -0,0 +1,101 @@
...
+def BayesianResNet(input_dim,
+                   num_classes=10,
+                   kernel_posterior_mean=-9,
+                   kernel_posterior_stddev=0.1,
+                   kernel_posterior_constraint=0.2):
+  """
+  Args:
+    input_dim: A Numpy `array` of IMAGE_SHAPE.
+    kernel_posterior_mean: Python `int` number for the initial kernel
+      posterior log mean. The smaller the mean the closer is the
```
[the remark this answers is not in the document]
jmzeng: yes. These values are actually setting the mean and stddev for the scale parameter distribution of the kernel_posterior. Hence, it's setting the variance to a smaller value. I'll clarify that.

jmzeng:
Please see the new updates. Let me know if the kernel posterior initialization is more clear.

tensorflow_probability/examples/models/bayesian_resnet.py
```diff
@@ -0,0 +1,104 @@
...
+import tensorflow_probability as tfp
+
+
+def bayesian_resnet(input_dim,
+                    num_classes=10,
+                    kernel_posterior_scale_mean=-9,
+                    kernel_posterior_scale_stddev=0.1,
+                    kernel_posterior_scale_constraint=0.2):
+  """
+  Constructs a ResNet18 model for training.
```
dustinvtran: Move this line into the same line as the """.

tensorflow_probability/examples/models/bayesian_resnet.py
```diff
@@ -0,0 +1,104 @@
+"""Builds a Bayesian ResNet18 Model"""
```
dustinvtran: Add period to end of this docstring.

tensorflow_probability/examples/models/bayesian_resnet.py
```diff
@@ -0,0 +1,104 @@
...
+
```
dustinvtran: You need to import the three future statements as in your script.

tensorflow_probability/examples/models/bayesian_resnet.py
dustinvtran: nit: This isn't just for training, right? Maybe omit "for training".

tensorflow_probability/examples/models/bayesian_vgg.py
```diff
@@ -0,0 +1,78 @@
+"""Builds a Bayesian VGG16 Model"""
```
dustinvtran: Apply same changes noted above.

tensorflow_probability/examples/models/bayesian_resnet.py
```diff
@@ -0,0 +1,104 @@
...
+def bayesian_resnet(input_dim,
+                    num_classes=10,
+                    kernel_posterior_scale_mean=-9,
+                    kernel_posterior_scale_stddev=0.1,
+                    kernel_posterior_scale_constraint=0.2):
+  """
+  Constructs a ResNet18 model for training.
+
+  Args:
+    input_dim: A `tuple` indicating the Tensor shape.
```
dustinvtran: rename to `input_shape`. dim refers to a single value (i.e., one element of a Tensor's rank).

tensorflow_probability/examples/cifar10_bnn.py
```diff
@@ -0,0 +1,253 @@
...
+                     default=400,
+                     help="Frequency at which to validate the model.")
+flags.DEFINE_integer("num_monte_carlo",
+                     default=50,
+                     help="Network draws to compute predictive probabilities.")
+flags.DEFINE_string("architecture",
+                    default="resnet",
+                    help="Network architecture to use.")
+flags.DEFINE_integer(
+    "kernel_scale_mean",
```
dustinvtran: To be consistent, this should be called `kernel_posterior_scale_mean`. Same for the constraint.

tensorflow_probability/examples/cifar10_bnn.py
```diff
@@ -0,0 +1,253 @@
...
+
+  # Perform KL annealing. The optimal number of annealing steps 
+  # depends on the dataset and architecture.   
+  t = tf.Variable(0.0)
+  kl_regularizer = t / (FLAGS.kl_annealing * len(x_train) / FLAGS.batch_size)
+
+  # Compute the -ELBO as the loss, averaged over the batch size.
+  log_likelihood = labels_distribution.log_prob(labels)
+  neg_log_likelihood = -tf.reduce_mean(log_likelihood)
+  kl = sum(model.losses) / len(x_train) * tf.minimum(1.0, kl_regularizer)
```
dustinvtran: The proper ELBO doesn't scale by batch size on the KL, right? If I understand correctly, you should write a comment saying that you're scaling the KL term.

jmzeng:
Updated! I did a merge upstream. Let me know if that worked.

tensorflow_probability/examples/cifar10_bnn.py
```diff
@@ -0,0 +1,263 @@
...
+flags.DEFINE_integer("eval_freq",
+                     default=400,
+                     help="Frequency at which to validate the model.")
+flags.DEFINE_integer("num_monte_carlo",
+                     default=50,
+                     help="Network draws to compute predictive probabilities.")
+flags.DEFINE_string("architecture",
+                    default="resnet",
+                    help="Network architecture to use.")
+flags.DEFINE_integer(
```
dustinvtran: This should be a float, right?

tensorflow_probability/examples/cifar10_bnn.py
```diff
@@ -0,0 +1,263 @@
...
+from tensorflow.python.keras.datasets import cifar10
+
+from models.bayesian_resnet import bayesian_resnet
+from models.bayesian_vgg import bayesian_vgg
+
+import warnings
+warnings.simplefilter(action='ignore')
+
+
+tfd = tf.contrib.distributions
```
dustinvtran: Use `import tensorflow_probability as tfp` above, and then write here `tfd = tfp.distributions`.

tensorflow_probability/examples/models/bayesian_resnet.py
```diff
@@ -0,0 +1,107 @@
...
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
+import tensorflow_probability as tfp
+
+
+def bayesian_resnet(input_shape,
+                    num_classes=10,
+                    kernel_posterior_scale_mean=-9,
```
dustinvtran: Following above note, set the defualt to be a float, i.e., `9.`.

tensorflow_probability/examples/models/bayesian_vgg.py
```diff
@@ -0,0 +1,81 @@
...
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
+import tensorflow_probability as tfp
+
+
+def bayesian_vgg(input_shape,
+                 num_classes=10,
+                 kernel_posterior_scale_mean=-9,
```
dustinvtran: Following above note, set the defualt to be a float, i.e., `9.`.

dustinvtran:
> Updated! I did a merge upstream. Let me know if that worked.

Can you try it again? In your branch, you don't have the `experimental/` subdirectory for example. This is causing issues for our internal merging tool.

@srvasude: Can you also review these changes as the idiom reviewer? I would send you the internal pull request but am getting the error from above.

jmzeng:
Fixed the changes. This should have worked?

dustinvtran:
Thanks to @cgs' help, it looks like you need to add the LICENSE to all files, not just the script.

jmzeng:
License info added.

csuter:
Hi @jmzeng, in order to pull this in we will need to `rebase` the commits on the head of master. The merge causes issues for our import system (and, in general, rebases are cleaner anyway). The fix should be pretty easy; just run these commands:


```bash
# Add upstream repo as a remote (you may already have this; if so, replace 'google'
# in subsequent commands with whatever your remote is called).
git remote add google https://github.com/tensorflow/probability

# Fetch upstream assets. Do this even if you already had our remote repo
# configured (just s/google/your-remote-name/), since you may need to pull down
# very recent changes before rebasing on top.
git fetch google

# Make sure your feature branch is checked out
git checkout bayesianCNNs

# Make a new branch pointing at the same commit
git checkout -b fix

# Move fix branch back to the commit right before the merge with master
git reset --hard 2a5edc2

# rebase everything up to this point on the upstream master
git rebase google/master

# cherrypick the license change that came after the merge
git cherrypick 9a64513

# verify the diff with feature branch is empty (it should be)
git diff bayesianCNNs HEAD

# Move back to the feature branch
git checkout bayesianCNNs

# Force feature branch to point to where fix is now
git reset --hard fix

# Force push the edited branch history to your remote fork
git push -f bayesianCNNs

# Delete the temporary branch we used for fixing
git branch -D fix
```

After the above sequence, the PR should be up to date with our master and we should be able to pull it in without any issues. If you have any questions or run into any issues, please let me know and I'll work with you to resolve ASAP.

In the future, you can avoid some of this pain by using the first few steps to rebase your working feature branch (workflow, from your feature branch: `git remote fetch google`, `git rebase google/master`). This will always take your most recent changes and move them to be on top of the most recent master branch state, instead of branching off from an earlier point (changing the "base" of the branch, hence "rebase" :))

Thanks for your patience and great contribution to TFP!

jmzeng:
Hi @csuter, I'm on the diff step and here is what's printed. Is it okay to force merge the two branches at this point? 

```
diff --git a/tensorflow_probability/python/sts/local_linear_trend_test.py b/tensorflow_probability/python/sts/local_linear_trend_test.py
index be7a359..7485962 100644
--- a/tensorflow_probability/python/sts/local_linear_trend_test.py
+++ b/tensorflow_probability/python/sts/local_linear_trend_test.py
@@ -27,7 +27,7 @@ from tensorflow_probability.python.sts import LocalLinearTrendStateSpaceModel
 from tensorflow.python.framework import test_util
 from tensorflow.python.platform import test
 
-tfl = tf.contrib.linalg
+tfl = tf.linalg
 tfd = tfp.distributions
 
 
diff --git a/tensorflow_probability/python/sts/structural_time_series.py b/tensorflow_probability/python/sts/structural_time_series.py
index 8bdb8cb..6a353ff 100644
--- a/tensorflow_probability/python/sts/structural_time_series.py
+++ b/tensorflow_probability/python/sts/structural_time_series.py
@@ -27,7 +27,7 @@ from tensorflow_probability.python.sts.internal import util as sts_util
 from tensorflow.contrib.distributions.python.ops import distribution_util
 from tensorflow.contrib.distributions.python.ops.seed_stream import SeedStream
 
-tfl = tf.contrib.linalg
+tfl = tf.linalg
 
 Parameter = collections.namedtuple('Parameter', ['name', 'prior', 'bijector'])
```

csuter:
Ah this is fine. This change went in a few minutes ago, and while it's in the `fix` branch, it's not in your feature branch yet, hence the diff. But this is fine because the next step will move the feature branch to point to `fix` anyway. As that's the only difference between them, it'll be fine.

jmzeng:
Pushed the new branch. Let me know if that worked.

csuter:
That did the trick! Thanks again, @jmzeng

jmzeng:
Great! and np, thanks for helping me through the process! @csuter @dustinvtran

csuter:
Hi @jmzeng -- I just sent a PR to your repo adding a couple things that will be necessary for us to pull this in. Please have a look and let me know if you have any questions. Apologies for the delay here, as I've been a bit sidetracked and also needed to figure out the right way to pull this example in, given that it is the first example to introduce and rely upon submodules in the examples directory.

jmzeng:
@csuter, I just merged the request. Let me know if you need to change other things.

dustinvtran:
Hi @jmzeng. It looks your PR is breaking the linting test. For convenience I sent you a PR in order to fix them:

https://github.com/jmzeng/probability/pull/2

dustinvtran:
How funny. @cgs: Following the note, can you flip the cla bit to yes?

csuter:
Verified CLA, as both Dustin and I are Google employees and have previously done everything necessary to make such contributions.

dustinvtran:
Thanks again @jmzeng for iterating with us on this!

jmzeng:
Thank you @csuter, @dustinvtran for helping me through this! I'm glad the examples are now included!

PR #162, opened 2018-09-24
rename build file

bstriner:
The filename `BUILD` conflicts with the directory name `build` on Windows (case insensitive), so I get an error like the below. This PR just adds the `.bazel` extension so there are no conflicts. Bazel seems to still find the file correctly. Adding the extension to the other `BUILD` files is unnecessary but wouldn't be a bad idea as well.

> running bdist_wheel
> running build
> running build_py
> creating build
> error: could not create 'build': Cannot create a file when that file already exists

Cheers


csuter:
Hi @bstriner -- did you run the setup script directly? The supported way to install from source is to install Bazel and use it to build a pip package as described here: https://github.com/tensorflow/probability#installing-from-source

Can you give this a try?

bstriner:
That is what I originally tried and ran into a different error.

> bazel-bin\pip_pkg: line 36: cd: bazel-bin/pip_pkg.runfiles/tensorflow_probability: No such file or directory

The created directory is actually `pip_pkg.exe.runfiles`, and there is no `tensorflow_probability` subdirectory.

csuter:
I see. Thanks for raising this. Our windows test coverage is lacking at the moment, so this was overlooked. We will need to fix this a bit more deeply on our end, and this will take a bit of research. The current change unfortunately won't be able to be accepted; we don't support direct installation through setup.py since we'll (eventually) use bazel to compile native targets as well as just piecing together the package contents.

In the mean time, you can get very up-to-date code by pip installing tfp-nightly[-gpu]. I'm afraid I don't have an easy workaround for making local changes and installing. I will file an issue about this, though, and follow up once we've properly fixed the Windows build.

bstriner:
@csuter I get that this doesn't fix the root issue, but the extensions should be changed regardless. `BUILD.bazel` is just for quality-of-life. Works the same as before, but you can setup `.bazel` extensions in your IDE or whatever. Is there any particular reason to not do a rename? Should probably do the other files as well.

I was thinking about making a pure python pip_pkg and fixing it for Windows. Please let me know the plan and if there is anything that needs help. I've been building frequently on both windows and linux.

csuter:
Thanks, Ben. So, we maintain the source code for TFP in two locations -- on GitHub as well as Google-internal source control systems. We have an automated process for pushing internal changes out to GitHub and a semi-automated process for pulling external changes in from GitHub. In general we try to keep the number of in-bound and out-bound transformations (directory shuffling, file renaming) minimal. Renaming all the BUILD files, while feasible, seems excessive. Re: IDE, is the extension necessary (vs just matching all files called 'BUILD')?

Can you clarify what you mean about making a "pure python pip_pkg and fixing it for Windows"?

In general I expect it to be pretty difficult for outside contributors to predict the sometimes strange/esoteric requirements on the build and release procedures for TFP, given that we straddle two parallel universes of version control, and many of the continuous integration and automated release management processes are Google-internal and sadly can't always be transparent.

We will prioritize getting this fixed and working for Windows users quickly, though. There are mechanisms for making this work and testing it to ensure issues like this don't reoccur, we just haven't had a chance to set it all up yet.

bstriner:
Versioning is always a headache, so best of luck. It is easy to do git renames that maintain history, but that probably wouldn't sync correctly.

I don't love that pip_pkg.sh calls a specific python executable. I'm guessing a py_binary might be more universal than a sh_binary. Might take some testing but shouldn't be too hard to get the windows bazel build correct.

For now, if anyone is having issues, they can just do the rename locally.

Are you looking at Travis or Jenkins or something else?

csuter:
Ah, the issue isn't just versioning for changing the names -- internally they *have* to be called "BUILD" (no extension). So the issue is it's another item on a list of automated transformations we need to perform on every export and every import. We'd like to keep this list minimal (since it's just another thing to maintain/debug/fix if it breaks), as well as keep things in the OSS version as similar as possible all around.

pip_pkg should call whatever python is in the PATH (or windows equiv.) when invoked -- what would you rather see it do? Take an argument? I typically invoke this in a virtualenv where I know which python is being used. Not sure what others typically do...

We have substantial internal CI systems and it's unlikely we would duplicate that on Travis or Jenkins externally in the near future. We have some work on the queue to automatically export our internal CI signals to GitHub when people send a PR, which should help a bit. As part of that, we'd like to make it easier to, e.g., run lint checks and automatically kick off bazel tests, so contributors can be more confident their submissions will be able to be checked in smoothly/promptly.

bstriner:
Not sure why the sh_binary isn't building correctly, but switching to a py_binary builds correctly. The sh_binary is just creating the manifest but not copying the files. If pip_pkg.py was a python script, it would just use whatever python it is called from. Would make the argument parsing and all that cleaner anyways.

I use virtualenvs on one machine and a bunch of alt-installed pythons on another. Not ideal, but shouldn't assume python is always called python. Even just on default ubuntu, you might want to call "python3" or something. Annoying to constantly alias or update-alternatives.

bstriner:
@csuter Just got back from class and made most of the changes I was mentioning. Please check out the new commits. This builds correctly on Windows.

`pip_pkg` is now a python script, so should be easier to maintain, has better argument handling, etc. I reverted the `.bazel` rename.

Only other change I was considering is checking for some sort of "PYTHON_EXE" env variable.

bstriner:
Here is the update: https://github.com/tensorflow/probability/pull/165

PR #165, opened 2018-09-24
Fix Windows Build

bstriner:
Change from a bash script to a python script. Bazel build works correctly on windows. See discussion here: https://github.com/tensorflow/probability/pull/162


csuter:
Hey @bstriner, sorry for the delay here.

I tested this script locally as written and it has at least 2 issues:
  1. fails in python2
  2. doesn't place the output dir in the same location as the previous script

At a minimum, we'd need such a script to meet these criteria. We'll also need to tidy up the code to match google python style guide. Some immediate style issues I see:
  - code directly in the `if __name__ == "__main__" block (prefer to define a main function and call it
  - some long lines (> 80 chars)
  - prefer absl.flags over argparse

The idea of invoking setup.py directly is interesting. I think being cd'ed into the appropriate dir would still be necessary (since bazel does some shuffling around of things, and will eventually also do c++ library compilation and possibly other stuff), so i'm not sure it would work out of the box. I'm all for minimalism though, so if there's a way to work around the pwd issue, I'm all for it!

bstriner:
@csuter I think with adding a `os.chdir(os.path.dirname(__file__))` to the top of setup.py the py_binary should work with just the below. I'll give it a try later. Would make more sense to call that target something like `setup` or `install`.

```
py_binary(
    name = "setup",
    srcs = ["setup.py"],
    data = [
        "//tensorflow_probability"
    ],
)
```

Not sure what the py2 issue is with the current PR. Probably something quick to fix, but using the setup.py directly seems cleaner.

The directory structure in bazel-bin might be different, but the executable is in the same place. Running the binary will extract into a temp directory but won't do the CD. That way the directory structure doesn't really matter as long as you have the binary in the right place.

csuter:
For now, in the interest of expediency and minimizing number of changes at once, why don't we focus on getting a python-equivalent to the shell script (pip_pkg.py) that works in all the required environments.

As mentioned before, there are internal pieces that aren't externally visible, and I won't know without further investigation and tinkering how all these interact. I'll need to think more about not having the wrapper script, as well as implications of changing the behavior of setup.py in this way.

csuter:
Hey @bstriner -- are you still thinking about getting this checked in? I think there is some more work to do (see previous comments above), but I think it'd be great to be able to properly support Windows builds.

Issue #69, opened 2018-06-02
Item response theory model example/tutorial


dustinvtran:
See also https://github.com/blei-lab/edward/blob/master/examples/irt.py.

joshchang:
I'm attempting to implement a graded response model in tfp as a learning problem but I cannot figure out where I am going wrong. Apologies in advance since I am new to tensorflow

The GRM has likelihood

```latex
\begin{align}
\Pr( X_{pi} = j \vert \theta_p, \tau_{i,j}, \tau_{i,j+1} , \lambda_i  )  &= \Pr(X_{pi}\geq j \vert\theta_p, \tau , \lambda )- \Pr(X_{pi}\geq j+1 \vert\theta_p, \tau , \lambda ) \nonumber\\
&=  S(\lambda_i(\theta_p-\tau_{i,j})) - S(\lambda_i(\theta_p-\tau_{i,j+1})),
\end{align}
```

where  S is the sigmoid function and I'm using the simple prior

```latex
\begin{align}
    \lambda_i&\sim C^+(0,5) \nonumber\\
    \tau_{i,0}\vert \mu_i & \sim N(\mu_i,3) \nonumber\\
    \mu_i &\sim N(0,3) \nonumber \\
    \tau_{i,j}|\tau_{i,j-1} &\sim N^+(\tau_{i,j-1},3)
\end{align}
```
I coded these up

```python

"""
Probabilities for a single person
discriminations   I (item)
difficulties I x K-1
abilities N 

returns
prob N x I x K
"""
def grm_model_prob(abilities, discriminations, difficulties, expanded=False):
    augmented_difficulties = tf.pad(difficulties, paddings=(
         (0, 0), (1, 0)), mode='constant', constant_values=-np.inf)
    augmented_difficulties = tf.pad(augmented_difficulties, paddings=(
         (0, 0), (0, 1)), mode='constant', constant_values=np.inf)
    offsets = augmented_difficulties[tf.newaxis,:,:] - abilities[:, tf.newaxis, tf.newaxis]
    scaled = offsets*discriminations[tf.newaxis, :, tf.newaxis]
    logits = 1.0/(1+tf.exp(scaled))
    probs = logits[:, :, :-1] - logits[:, :, 1:]
    
    return probs

"""
Probabilities for single items
discriminations   I (item)
difficulties  I x K - 1
abilities N

mu (difficulty local) I
tau (difficulty) I x K-2


"""
[...]

PR #171, opened 2018-10-01
Add example for Item response theory

souravsingh:
Fixes issue #69

### DIFF
--- a/tensorflow_probability/examples/irt.py
+++ b/tensorflow_probability/examples/irt.py
@@ -1,0 +1,144 @@
+"""Bayesian Item Response Theory (IRT) Mixed Effects Model
+using variational inference.
+Simulates data and fits y ~ 1 + (1|student) + (1|question)
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import edward as ed
+import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


davmre:
Hi @souravsingh, thanks for the PR! I think we'd need a couple of changes to incorporate this in TFP:

- Currently you're using the inference module from Edward 1, but TFP doesn't have an Edward dependency, so we'd need to write variational inference directly in TF. There are a few examples of doing this: see, e.g., the [deep exponential family](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/examples/deep_exponential_family.py) or [Probabilistic PCA](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/examples/jupyter_notebooks/Probabilistic_PCA.ipynb) examples.

- The command-line examples we have that use Matplotlib generally save plots to a file; however, for examples like this one where the main output is a plot, we prefer a Jupyter notebook format (see examples at https://github.com/tensorflow/probability/tree/master/tensorflow_probability/examples/jupyter_notebooks).

It might be some work to update this example, but if you're willing to do so we'd be happy to iterate on it with you!

souravsingh:
@davmre Thanks for the review. I will look into the comments and fix the PR.

Issue #172, opened 2018-10-02
Can't pass variable size to bfgs_minimize


rmrao:
I've been trying to pass a variable size tensor to bfgs_minimize and have been running into issues. It looks like around line [200](https://github.com/tensorflow/probability/blob/76011694f512ac91c2024d05f4491f36d9510410/tensorflow_probability/python/optimizer/bfgs.py#L200-L203), the shape of the hessian is obtained using `tensor.shape` as opposed to `tf.shape(tensor)` which causes issue on the subsequent reshape.

Changing this to `tf.shape(tensor)` however, causes in issue with the loop (specifically, the size of the tensor within the while loop gets changed from its initial size to `<unknown>`). Tensorflow suggests passing in `shape_invariants` to deal with the issue.

I'm curious if there is some reason that bfgs minimize does not support variable size tensors or if there's a reason to believe it wouldn't work properly. If not, I can try and edit it to work with variable size tensors and submit a pull request.

csuter:
Thanks for sending this, @rmrao! Our colleagues in London, who developed this part of TFP, were *just* talking about this issue on an internal list, and will be taking a look at your PR in the AM (London time).

rmrao:
Sounds good! Let me know if there are any issues, I'm happy to debug since I'm using it in a project now.

perezju:
Should be fixed after this [commit](https://github.com/tensorflow/probability/commit/1508ef5adf078696dd13353cf87b7f9118e957ab).

PR #173, opened 2018-10-02
change bfgs to work with variable size inputs

rmrao:
Fix for #172. Allows variable size tensors to be passed to bfgs_minimize.

### DIFF
--- a/tensorflow_probability/python/optimizer/bfgs.py
+++ b/tensorflow_probability/python/optimizer/bfgs.py
@@ -197,10 +197,10 @@ def minimize(value_and_gradients_function,
                                        name='x_tolerance')
     max_iterations = tf.convert_to_tensor(max_iterations, name='max_iterations')
 
-    domain_shape = initial_position.shape
+    domain_shape = tf.shape(initial_position)
 
     if initial_inverse_hessian_estimate is None:
-      inv_hessian_shape = domain_shape.concatenate(domain_shape)
+      inv_hessian_shape = tf.concat((domain_shape, domain_shape), 0)
       initial_inv_hessian = tf.eye(tf.size(initial_position), dtype=dtype)
       initial_inv_hessian = tf.reshape(initial_inv_hessian,
                                        inv_hessian_shape,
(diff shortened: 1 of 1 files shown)


rmrao:
Signed the CLA

tensorflow_probability/python/optimizer/bfgs.py
```diff
@@ -604,12 +604,12 @@ def _do_update_fn():
     """Updates the Hessian estimate."""
     # The quadratic form: y^T.H.y.
     n = len(grad_delta.shape.as_list())
```
saxena-ashish-g: We would prefer to have full support for dynamic shapes which includes statically unknown rank as well. shape.as_list() will raise an exception if shape = None.

tensorflow_probability/python/optimizer/bfgs.py
```diff
@@ -197,10 +197,10 @@ def quadratic(x):
                                        name='x_tolerance')
     max_iterations = tf.convert_to_tensor(max_iterations, name='max_iterations')
 
-    domain_shape = initial_position.shape
+    domain_shape = tf.shape(initial_position)
 
     if initial_inverse_hessian_estimate is None:
-      inv_hessian_shape = domain_shape.concatenate(domain_shape)
+      inv_hessian_shape = tf.concat((domain_shape, domain_shape), 0)
```
saxena-ashish-g: Please use a list instead of a tuple for consistency with the rest of the code base.

rmrao:
Sure I can work on editing these - it might be a few days before I get around to it.

tensorflow_probability/python/optimizer/bfgs.py
rmrao: Hmm. Do you know what the best way of getting the number of dimensions of a tensor with unknown shape is? I think `tf.shape(tf.shape(grad_delta))` will work, although it's kind of ugly. Passing in a tensor there might be enough if the input shape is already `<unknown>`, since the only thing that matters is preserving the shape info through the tensordot.

saxena-ashish-g: [tf.rank()](https://www.tensorflow.org/api_docs/python/tf/rank) will give you the dynamic rank of the tensor.

perezju:
FYI - I'm looking into a fix for this now.

brianwa84:
#172 is closed now

Issue #161, opened 2018-09-19
MultivariateNormalDiagPlusLowRank inconsistent with practice


Bonnevie:
This is maybe more of a feature request, but in short, the fact that the distribution
[MultivariateNormalDiagPlusLowRank](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/MultivariateNormalDiagPlusLowRank) applies the eponymous low-rank assumption to a _scale_ matrix, and not the covariance, is somewhat unorthodox and not really in line with what most practitioners would expect.

In my opinion, it is fairly rare to see people explicitly model a dense scale matrix (although cholesky factors do show up for computational reasons) let alone assume structure. Diagonal-plus-low-rank structure is however very normal, with PPCA having marginal covariance $AA^T+kI$ for instance.

My guess is that this formulation is due to the easy definition as a TransformedDistribution? I think it would be worth it to include a diagonal-plus-low-rank _covariance_ normal distribution since there is a significant computational gain to be had, but unfortunately I do not think there is a way to do this in terms of a transformation. I also think it would be reasonable to emphasize the distinction between low-rank scale and low-rank covariance.

csuter:
Great points, @Bonnevie. One impediment to actually fixing this as you suggest is the lack of a low-rank Cholesky update procedure in TensorFlow. This is something I've been thinking about recently and am working on some proposals to address. No reliable ETA as of yet though. Thanks for raising this issue!

Bonnevie:
@csuter I don't think you really need a low-rank cholesky update, unless I am missing something. By applying [Woodbury's matrix identity](https://en.wikipedia.org/wiki/Woodbury_matrix_identity) you should be able to convert the matrix inverse in the large event space to an inverse in the lower-dimensional latent space. 

I implemented it in Stan at one point, which might be mildly useful if you can decipher the notation (it is for a N(x|mu,B*B^T + diag(sigma)) density):
```
    real multi_normal_lowrank_cholesky_lpdf(vector x, vector mu, vector sigma, matrix B, matrix L) {
        int lowrank = cols(B);  
        vector[lowrank] v = mdivide_left_tri_low(L, B' * ((x-mu) ./ square(sigma)));
        return normal_lpdf(x | mu, sigma) + dot_self(v)/2. - sum(log(diagonal(L)));
    }

    real multi_normal_lowrank_lpdf(vector x, vector mu, vector sigma, matrix B) {
        int lowrank = cols(B);  
[...]

PR #184, opened 2018-10-11
MultivariateNormalDiagPlusLowRankCovariance

Bonnevie:
This PR tries to resolve issue #161 by implementing a new variant of the multivariate normal that applies a diagonal plus low-rank assumption to the covariance, instead of the ill-defined scale matrix currently employed in `MultivariateNormalDiagPlusLowRank`.

The implementation is adapted from existing variants of the multivariate normal. It adapts the source code of `MultivariateNormalFullCovariance`, but uses code snippets from `MultivariateNormalDiagPlusLowRank` to construct the covariance matrix via `LinearOperator`s.
It then overwrites `_log_prob` and `_log_normalization` with more efficient versions exploiting the low-rank construction.

Tests are direct conversions of the tests for `MultivariateNormalDiagPlusLowRank` and could be replaced with something more case-specific. Density evaluations have been briefly validated against a similarly parameterized `FullCovariance`.


Bonnevie:
signed

csuter:
My apologies for not replying sooner on the main issue. I had been hoping (and some internal design discussion was proceeding in parallel) to get this behavior implemented as a special case in MultivariateNormalFullCovariance. The idea would be to allow the covariance argument to be a `Tensor` or a `LinearOperator` and (possibly) take advantage of more efficient ops in the LinOp case.

We, as a rule, would prefer to nudge people towards using loc+scale parameterization because, with some forethought, often an O(N^3) Cholesky decomposition can be avoided altogether. We'd like to minimize the number of covariance-parameterized distributions floating around, both for the sake of maximal consistency and for encouraging what we see as a best practice.

I'm not sure that folding this and the existing MVNFullCov together using the double-sampling approach will work, though, at least not cleanly. Another possible avenue would be to add the LinearOperator-based parameterization of MVNFullCov, and once we implement the efficient low-rank Cholesky updates, that will magically become more efficient.

However. I'd prefer not to introduce a new MVN variant (there are already a ton), especially with a non-(TFP-)standard parameterization. My apologies again for not speaking sooner; I didn't realize this (substantial!!) effort was already underway here.

srvasude:
@csuter , @davmre . What is the status here?

Bonnevie:
I had a long reply to this that I thought I had already posted, but I must have gotten derailed and forgotten to press "Comment".

I am happy to take this implementation in some other direction, but I would of course rather not spend time implementing something that is not going to be used (even if this particular implementation was useful for my own purposes).

The issue with collecting this under `MultivariateNormalFullCovariance` is that it implicitly uses `MultivariateNormalTriL` under the hood and Cholesky factors are not the best way to calculate the density for diagonal-plus-low-rank covariances. I think your idea of letting the class take an arbitrary `LinearOperator` as its covariance input is probably the way to go, forwarding the burden of optimizing the matrix solves/log-determinant to the operator. `TriL`could be incorporated by designing a `SymmetricPositiveDefinite` `LinearOperator` class.

I think advocating the loc+scale parameterization is commendable, but it makes a lot more sense in the univariate setting as scale is not defined properly in the multivariate setting (or is at the very least non-identifiable). 

What would be the recommended route forward?

brianwa84:
I've recently found a use-case where I want a diag+LR update covariance MVN (and not our unconventional diag+LR update scale MVN). Are you interested in resurrecting this PR?

If you want to, I have a suggestion for how we might get it in. I'd suggest:
1. we take most of the existing implementation of `MultivariateNormalDiagPlusLowRank` and turn it into a new private class in the same file `_MultivariateNormalDiagPlusLowRankScale`
2. we add a new class in the same file `_MultivariateNormalDiagPlusLowRankCov`
  This should not be a `TransformedDistribution`, instead should implement `_sample_n` and `_log_prob` directly.
3. we make (a new) `MultivariateNormalDiagPlusLowRank` directly extend `tfd.Distribution`, keep the existing constructor args (`scale_diag=None, scale_identity_multiplier=None, scale_perturb_factor=None, scale_perturb_diag=None`), and add some new cov args (`covariance_diag=None, covariance_identity_multiplier=None, covariance_perturb_factor=None, covariance_perturb_diag=None`)
4. `MultivariateNormalDiagPlusLowRank.__init__` looks at its parameters and sets `self._impl` to one of `_MultivariateNormalDiagPlusLowRankScale` or `_MultivariateNormalDiagPlusLowRankCov`, then defers to `self._impl` in `_log_prob`, `_sample_n`, etc.
5. (optional). add a `@deprecation.deprecated_args('2019-08-01', 'Eliminating unconventional parameterization. Email tfprobability@tensorflow.org if this was useful to you.', 'scale_diag', 'scale_identity_multiplier', 'scale_perturb_factor', 'scale_perturb_diag')` to `MultivariateNormalDiagPlusLowRank.__init__` to deprecate the existing (nonsensical, right?) scale parameterization.

langmore:
Note that a distribution similar to this PR has been added. Thanks @Bonnevie for the PR, and sorry this didn't get closed earlier. See [`MultivariateNormalDiagPlusLowRankCovariance`](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/distributions/mvn_diag_plus_low_rank_covariance.py).  Also of interest is the (hidden) base class [`MultivariateNormalLowRankUpdateLinearOperatorCovariance`](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/distributions/mvn_low_rank_update_linear_operator_covariance.py). It's hidden because I'm not sure if the base class should be more general.

PR #202, opened 2018-10-30
Add Laplace-Laplace KL divergence

JuliusKunze:


JuliusKunze:
I signed it!

tensorflow_probability/python/distributions/laplace.py
```diff
@@ -201,3 +202,26 @@ def _mode(self):
...
+      b: instance of a Laplace distribution object.
+      name: (optional) Name to use for created operations.
+        default is "kl_laplace_laplace".
+
+    Returns:
+      Batchwise KL(a || b)
+    """
+    with tf.name_scope(name, 'kl_laplace_laplace', [a.loc, b.loc, a.scale, b.scale]):
+        # Consistent with http://www.mast.queensu.ca/~communications/Papers/gil-msc11.pdf, page 38
+        one = tf.constant(1, dtype=a.dtype)
```
srvasude: You can just 1. inlined below (this should be automatically casted to the right type).

tensorflow_probability/python/distributions/laplace.py
```diff
@@ -201,3 +202,26 @@ def _mode(self):
...
+        distance = tf.abs(a.loc - b.loc)
+        ratio = a.scale / b.scale
+
+        return -tf.log(ratio) - one + distance / b.scale + ratio * tf.exp(- distance / a.scale)
```
srvasude: nit: extra space tf.exp(-distance / a.scale)

tensorflow_probability/python/distributions/laplace.py
```diff
@@ -201,3 +202,26 @@ def _mode(self):
...
+      a: instance of a Laplace distribution object.
+      b: instance of a Laplace distribution object.
+      name: (optional) Name to use for created operations.
+        default is "kl_laplace_laplace".
+
+    Returns:
+      Batchwise KL(a || b)
+    """
+    with tf.name_scope(name, 'kl_laplace_laplace', [a.loc, b.loc, a.scale, b.scale]):
+        # Consistent with http://www.mast.queensu.ca/~communications/Papers/gil-msc11.pdf, page 38
```
csuter: just eyballing, this looks possibly >80 chars. can you follow the [instructions from TF](https://github.com/tensorflow/tensorflow/blob/master/CONTRIBUTING.md#python-coding-style) on running a linter on the code?

tensorflow_probability/python/distributions/laplace_test.py
```diff
@@ -340,5 +343,30 @@ def testLaplaceNonPositiveInitializationParamsRaises(self):
...
+    a_scale = np.array([[0.1] * event_size] * batch_size, dtype=np.float32)
+    b_loc = np.array([[0.4] * event_size] * batch_size, dtype=np.float32)
+    b_scale = np.array([[0.2] * event_size] * batch_size, dtype=np.float32)
+
+    a = tfd.Laplace(loc=a_loc, scale=a_scale)
+    b = tfd.Laplace(loc=b_loc, scale=b_scale)
+
+    distance = tf.abs(a_loc - b_loc)
+    ratio = a_scale / b_scale
+    true_kl = -tf.log(ratio) - 1 + distance / b_scale + ratio * tf.exp(- distance / a_scale)
```
csuter: this is essentially re-writing the code under test in the test. maybe at least use numpy here? i think that's what we usually do elsewhere (others might chime in?)

srvasude: I don't have a strong feeling here since using numpy is basically equivalent to using TF here. Unfortunately we don't have a 'KL test harness' that we can test against, so testing this will inherently require writing a copy of the KL divergence function somewhere (I think the main things we want to test are that you recover zero when you pass in the same input, that batching / broadcasting works). I guess writing the function in numpy could potentially force writing a simpler thing (although in this case the code is pretty simple).

I guess another thing we can do is check that the sample KL is consistent with the analytic KL.

tensorflow_probability/python/distributions/laplace.py
JuliusKunze: Fixed.

tensorflow_probability/python/distributions/laplace_test.py
```diff
@@ -340,5 +342,31 @@ def testLaplaceNonPositiveInitializationParamsRaises(self):
...
+    a_scale = np.array([[0.1] * event_size] * batch_size, dtype=np.float32)
+    b_loc = np.array([[0.4] * event_size] * batch_size, dtype=np.float32)
+    b_scale = np.array([[0.2] * event_size] * batch_size, dtype=np.float32)
+
+    a = laplace_lib.Laplace(loc=a_loc, scale=a_scale)
+    b = laplace_lib.Laplace(loc=b_loc, scale=b_scale)
+
+    distance = tf.abs(a_loc - b_loc)
+    ratio = a_scale / b_scale
+    true_kl = -tf.log(ratio) - 1 + distance / b_scale + \
```
csuter: Nit: Prefer parentheses to newline escaping:

```suggestion
    true_kl = (-tf.log(ratio) - 1 + distance / b_scale +
                     ratio * tf.exp(-distance / a_scale))
```

tensorflow_probability/python/distributions/laplace.py
```diff
@@ -201,3 +202,28 @@ def _mode(self):
...
+  """
+  with tf.name_scope(name, 'kl_laplace_laplace',
+                     [a.loc, b.loc, a.scale, b.scale]):
+    # Consistent with
+    # http://www.mast.queensu.ca/~communications/Papers/gil-msc11.pdf, page 38
+    distance = tf.abs(a.loc - b.loc)
+    ratio = a.scale / b.scale
+
+    return -tf.log(ratio) - 1 + distance / b.scale + \
+           ratio * tf.exp(-distance / a.scale)
```
csuter: Here, too

```suggestion
    return (-tf.log(ratio) - 1 + distance / b.scale + 
            ratio * tf.exp(-distance / a.scale))
```

tensorflow_probability/python/distributions/laplace_test.py
JuliusKunze: Added the sample KL test.

tensorflow_probability/python/distributions/laplace_test.py
```diff
@@ -340,5 +342,36 @@ def testLaplaceNonPositiveInitializationParamsRaises(self):
...
+
+    x = a.sample(int(1e4), seed=0)
+    kl_sample = tf.reduce_mean(a.log_prob(x) - b.log_prob(x), 0)
+
+    true_kl_, kl_, kl_sample_ = self.evaluate([true_kl, kl, kl_sample])
+    self.assertAllEqual(true_kl_, kl_)
+    self.assertAllClose(true_kl_, kl_sample_, atol=0., rtol=1e-1)
+
+    zero_kl = kl_divergence(a, a)
+    true_zero_kl_, zero_kl_ = self.evaluate([tf.zeros_like(true_kl_), zero_kl])
```
brianwa84: consider zeros_like(true_kl) to reference the tensor instead of the np array

srvasude:
Linter doesn't complain, yay! Thanks again for being patient and for your quick responses!

PR #208, opened 2018-11-02
fix RationalQuadratic kernel

yejiming:
fix bugs regarding RationalQuadratic kernels, also update the tests.

given formula:
k(x, y) = (1 + d(x, y)^2 / (2*scale_mixture_rate * length_scale^2)) ^ (-scale_mixture_rate) * (amplitude ^ 2)

### DIFF
--- a/tensorflow_probability/python/positive_semidefinite_kernels/rational_quadratic.py
+++ b/tensorflow_probability/python/positive_semidefinite_kernels/rational_quadratic.py
@@ -116,7 +116,7 @@ class RationalQuadratic(psd_kernel.PositiveSemidefiniteKernel):
           amplitude, scale_mixture_rate, length_scale], tf.float32)
 
       if amplitude is not None:
-        amplitude = tf.convert_to_tensor(
+        self._amplitude = tf.convert_to_tensor(
             amplitude, name="amplitude", dtype=dtype)
         with tf.control_dependencies([
             tf.assert_positive(self._amplitude)] if validate_args else []):
@@ -146,13 +146,13 @@ class RationalQuadratic(psd_kernel.PositiveSemidefiniteKernel):
     if self.length_scale is not None:
       length_scale = util.pad_shape_right_with_ones(
           self.length_scale, ndims=param_expansion_ndims)
-      difference /= length_scale ** 2
+      difference /= (length_scale ** 2)
 
     scale_mixture_rate = 1.
     if self.scale_mixture_rate is not None:
       scale_mixture_rate = util.pad_shape_right_with_ones(
           self.scale_mixture_rate, ndims=param_expansion_ndims)
-      difference /= scale_mixture_rate
+    difference /= (2 * scale_mixture_rate)
 
     result = (1. + difference) ** -scale_mixture_rate
 
@@ -165,17 +165,23 @@ class RationalQuadratic(psd_kernel.PositiveSemidefiniteKernel):
   @property
   def amplitude(self):
     """Amplitude parameter."""
-    return self.amplitude
+    if "_amplitude" in self.__dict__:
+        return self._amplitude
+    return None
 
   @property
   def length_scale(self):
     """Length scale parameter."""
-    return self.length_scale
+    if "_length_scale" in self.__dict__:
+        return self._length_scale
+    return None
 
   @property
   def scale_mixture_rate(self):
     """scale_mixture_rate parameter."""
-    return self.scale_mixture_rate
+    if "_scale_mixture_rate" in self.__dict__:
+        return self._scale_mixture_rate
+    return None
 
   def _batch_shape(self):
     shape_list = [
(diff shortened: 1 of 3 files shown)


tensorflow_probability/python/positive_semidefinite_kernels/rational_quadratic.py
```diff
@@ -116,7 +116,7 @@ def __init__(
           amplitude, scale_mixture_rate, length_scale], tf.float32)
 
       if amplitude is not None:
```
srvasude:
```suggestion
      if amplitude is not None:
       amplitude = tf.convert_to_tensor(amplitude, name="amplitude", dtype=dtype)
     self._amplitude = _validate_args_if_not_none(amplitude, tf.assert_positive, validate_args)
```

Same here and below. That should fix the issue of it is missing or not.

tensorflow_probability/python/positive_semidefinite_kernels/rational_quadratic.py
```diff
@@ -165,17 +165,23 @@ def _apply(self, x1, x2, param_expansion_ndims=0):
   @property
   def amplitude(self):
     """Amplitude parameter."""
-    return self.amplitude
+    if "_amplitude" in self.__dict__:
```
srvasude: Can remove these statements and simply return self._amplitude

tensorflow_probability/python/positive_semidefinite_kernels/rational_quadratic.py
```diff
@@ -146,13 +146,13 @@ def _apply(self, x1, x2, param_expansion_ndims=0):
...
           self.length_scale, ndims=param_expansion_ndims)
-      difference /= length_scale ** 2
+      difference /= (length_scale ** 2)
 
     scale_mixture_rate = 1.
     if self.scale_mixture_rate is not None:
       scale_mixture_rate = util.pad_shape_right_with_ones(
           self.scale_mixture_rate, ndims=param_expansion_ndims)
-      difference /= scale_mixture_rate
+    difference /= (2 * scale_mixture_rate)
```
srvasude: Maybe move the / 2 up (move it to where the squared difference is calculated).

tensorflow_probability/python/positive_semidefinite_kernels/rational_quadratic_test.py
```diff
@@ -31,10 +31,10 @@ class RationalQuadraticTest(tf.test.TestCase, parameterized.TestCase):
   def _rational_quadratic(
       self, amplitude, length_scale, scale_mixture_rate, x, y):
     return (1. + np.sum((x - y) ** 2) / (
-        2 * scale_mixture_rate * length_scale ** 2)) ** (-scale_mixture_rate)
+        2 * scale_mixture_rate * length_scale ** 2)) ** (-scale_mixture_rate) * (amplitude ** 2)
```
srvasude: Put this at the front of the expression.

tensorflow_probability/python/positive_semidefinite_kernels/rational_quadratic.py
peterroelants: `self._amplitude` is not set in `__init__` if the parameter is None.

peterroelants: > Can remove these statements and simply return self._amplitude

`self._amplitude` is not set in `__init__` if the parameter is None.

brianwa84: we should always set it, even if it gets set to none.
 as done in other kernels, eg https://github.com/tensorflow/probability/blob/3f7c64714be3fad6cb6f0c25748f44c4cb9b7d3a/tensorflow_probability/python/positive_semidefinite_kernels/exp_sin_squared.py

srvasude: Yep. My above comment (on the __init__ changes) would always set self._amplitude, and so these properties would only have to return the hidden attribute.

yejiming: Yes I got you. Have fixed this.

tensorflow_probability/python/positive_semidefinite_kernels/rational_quadratic.py
```diff
@@ -129,24 +130,27 @@ def __init__(
             tf.assert_positive(
                 self._scale_mixture_rate)] if validate_args else []):
           self._scale_mixture_rate = tf.identity(self._scale_mixture_rate)
+      self._scale_mixture_rate = _validate_arg_if_not_none(scale_mixture_rate, tf.assert_positive, validate_args)
```
srvasude: This should either take self._scale_mixture_rate (since the positivity check is embedded above in to that), or all the statements above should only modify scale_mixture_rate.

srvasude: The same below for every other block

yejiming: Sure, just have fixed it

srvasude: Sorry, can you remove all the control_dependencies lines? _validate_arg_if_not_none already does that. The minimal block should be:

```suggestion
if amplitude is not None:
  amplitude = tf.convert_to_tensor(amplitude, name="amplitude", dtype=dtype)
self._amplitude = _validate_args_if_not_none(
  amplitude, tf.assert_positive, validate_args)
```

yejiming: sure

PR #222, opened 2018-11-09
Fix #220

### LINKED ISSUE #220: On the gradients of von Mises CDF when the series expansion is used
In the following lines of `_von_mises_cdf_series`, the comment states that the gradient for the clipped values should be set to zero.
https://github.com/tensorflow/probability/blob/955a0d7ea3f91e403c867bdb3f7802e7a01f2f8e/tensorflow_probability/python/distributions/von_mises.py#L409-L415

However the `cdf` is already clipped to have values between `[0, 1]`, thus the cast operation always emits the vector of all ones, and I think the code would not make gradients of clipped values be zero.
I think the lines 412--415 should be corrected as follows:
```python
cdf_ = tf.clip_by_value(cdf, zero, one) 
dcdf_dconcentration *= tf.cast((cdf >= zero) & (cdf <= one), dtype) 
cdf = cdf_
```
though I found that the current implementation already makes the gradients very close to zero.

```
In [38]: from tensorflow_probability.python.distributions import von_mises

In [39]: x = np.array([[3, 0.1, 3.20]]).astype(np.float32)

In [40]: kappa = np.array([[10, 5, 20]]).astype(np.float32)

In [41]: tf.enable_eager_execution()

In [42]: von_mises._von_mises_cdf_series(x, kappa, 20, tf.float32)
Out[42]:
(<tf.Tensor: id=9796, shape=(1, 3), dtype=float32, numpy=array([[0.99999994, 0.5859968 , 1.        ]], dtype=float32)>,
 <tf.Tensor: id=9801, shape=(1, 3), dtype=float32, numpy=array([[7.8744311e-10, 9.0264026e-03, 3.0936786e-07]], dtype=float32)>)
```

Please let me know if I understood the code wrong!

jihunchoi:
When computing von Mises CDF using the series expansion, it may make some values be above one or below zero, and it is corrected by clipping the values to [0, 1].
The gradients for clipped values should be zero, however the current implementation does not consider this correctly.
Discussed in #220.

### DIFF
--- a/tensorflow_probability/python/distributions/von_mises.py
+++ b/tensorflow_probability/python/distributions/von_mises.py
@@ -409,12 +409,12 @@ def _von_mises_cdf_series(x, concentration, num_terms, dtype):
   # Clip the result to [0, 1].
   zero = tf.cast(0., dtype)
   one = tf.cast(1., dtype)
-  cdf = tf.clip_by_value(cdf, zero, one)
+  cdf_clipped = tf.clip_by_value(cdf, zero, one)
   # The clipped values do not depend on concentration anymore, so set their
   # derivative to zero.
   dcdf_dconcentration *= tf.cast((cdf >= zero) & (cdf <= one), dtype)
 
-  return cdf, dcdf_dconcentration
+  return cdf_clipped, dcdf_dconcentration
 
 
 def _compute_value_and_grad(f, x):


mfigurnov:
Looks good, thank you! It would be great to add a test, but I can't think of a reliable way to force `cdf` outside the `[0, 1]` range. As discussed in #220, this only happens due to numerical instability, so is by definition sensitive to the implementation details (`num_terms`, `dtype` and such). @brianwa84 do you have any suggestions?

PR #230, opened 2018-11-19
Delete inappropriate tags in sample code.

pshiko:
Fixed a typo where syntax highlighting was not performed correctly because inappropriate tags were mixed in the sample code.
I do not know why this tag exists.
I am sorry if it is an incorrect fix or if it has already been fixed.

### DIFF
--- a/tensorflow_probability/g3doc/api_docs/python/tfp/glm/fit.md
+++ b/tensorflow_probability/g3doc/api_docs/python/tfp/glm/fit.md
@@ -121,7 +121,8 @@ def make_dataset(n, d, link, scale=1., dtype=np.float32):
         dtype)
   elif link == 'logit':
     response = tfd.Bernoulli(logits=linear_response).sample(seed=44)
-* <b>`else`</b>:     raise ValueError('unrecognized true link: {}'.format(link))
+  else:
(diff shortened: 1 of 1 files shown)


csuter:
Thanks for the PR, but actually these api docs are auto-generated from the python docstrings, so to fix we need to make the change in the python code, then internally we can rebuild and push the docs update. Do you mind moving the change to the python source?

pshiko:
Of course.
However, when I confirmed the source, docstring seems to have no problem.
This appears to be a bug in gdoc3 when generating tags in doxygen format.
I confirmed the same problem even in tfp.glm.fit_sparse md.
Since I do not have any knowledge about gdoc 3's problem, can you please fix it?

csuter:
Ah, I see. Thanks very much for bringing this to our attention. We'll iterate on our internal tools and hopefully have a fix out soon!

MarkDaoust:
Yes. Thanks for the report.

It appears the problem is in the tool that parses doc-strings to markdown, where it tries to convert Args/Returns/Raises blocks.

https://github.com/tensorflow/docs/blob/9aa6df82399c4fbca552af98d42c6332ca4614be/tools/tensorflow_docs/api_generator/parser.py#L427

I'll look into a fix.

PR #231, opened 2018-11-20
Fix T-distribution sampling parameter

nickyungyung:
As title, this is proposed as a bug-fix initially in tensorflow/tensorflow, which I then realized it was deprecated. The pull request has been reviewed and for reference: https://github.com/tensorflow/tensorflow/pull/23301


csuter:
Note also, there is a unit test for the StudentT sample method, which confirms that the first and second central moments are correct. That test breaks if I patch this change in.

csuter:
I'm going to close this, but thanks again for the engagement; do let me know if there's more to discuss here (e.g., could we document the code better?). We welcome your future contributions!

PR #232, opened 2018-11-20
Fixed example code in GP regression

gftabor:
Example code threw an error because of a missing parenthesis


csuter:
Thanks! These docs are actually auto-generated. Would you mind moving the fix to the docstring in the corresponding python source code?

gftabor:
@csuter fixed in the python source code.

PR #235, opened 2018-11-23
Fix: Added the prefixes to the returned variables.

ValentinMouret:


ValentinMouret:
I signed it!

davmre:
Good catch! Unfortunately we don't have a way to automatically test embedded code in Markdown files, so thanks for noticing and fixing this!

I've started the process of merging this internally; should be updated here soon.

PR #236, opened 2018-11-24
Don't globally set warning filters

ywkim:
### Reference Issues/PRs
Fixes https://github.com/tensorflow/probability/issues/237

### What does this Pull Request accomplish?

- We shouldn't have been setting the global warning filters
- Do not call `warnings.simplefilter()` as part of the API
- It is now up to the user of the API to set the warnings filter appropriately for them

### DIFF
--- a/tensorflow_probability/python/mcmc/metropolis_hastings.py
+++ b/tensorflow_probability/python/mcmc/metropolis_hastings.py
@@ -32,12 +32,6 @@ __all__ = [
     'MetropolisHastings',
 ]
 
-
-# Cause all warnings to always be triggered.
-# Not having this means subsequent calls wont trigger the warning.
-warnings.simplefilter('always')
-
-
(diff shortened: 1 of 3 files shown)


ywkim:
I signed it!

davmre:
This looks great to me. @jvdillon can you sanity check? is there any reason we need to set the warnings filter globally?

brianwa84:
I think this seems fine, Dave are you going to take care of the merge?

davmre:
I went ahead and merged a slightly modified version of this PR: we keep the filter setting but restrict it to the local module, and allow users to override it. E.g.,

```python
warnings.filterwarnings('always',
                        module='tensorflow_probability.*metropolis_hastings',
                        append=True)  # Don't override user-set filters.
```

I believe this should address the issue. Thanks again @ywkim for reporting this!

PR #240, opened 2018-11-28
Fix typo: edward2 README

yukinagae:
Just fix typo from `kernels_results` to `kernel_results` as `kernel_results` is declared few lines of code before (not `kernels_results`).


davmre:
Thanks for the fix @yukinagae!

PR #242, opened 2018-12-02
Update categorical.py

harrismirza:
Added change to Categorical Entropy to ignore non-finite probability values.

### DIFF
--- a/tensorflow_probability/python/distributions/categorical.py
+++ b/tensorflow_probability/python/distributions/categorical.py
@@ -300,7 +300,7 @@ class Categorical(distribution.Distribution):
 
   def _entropy(self):
     entropies = self.probs * tf.log(self.probs)
-    return -tf.reduce_sum(tf.boolean_mask(entropies, tf.is_finite(entropies)), axis=-1)
+    return -tf.reduce_sum(tf.xlogy(self.probs, self.probs), axis=-1)
 
   def _mode(self):
     ret = tf.argmax(self.logits, axis=self._batch_rank)
--- a/tensorflow_probability/python/distributions/categorical.py
+++ b/tensorflow_probability/python/distributions/categorical.py
@@ -299,8 +299,8 @@ class Categorical(distribution.Distribution):
                                                            logits=logits)
 
   def _entropy(self):
-    return -tf.reduce_sum(
-        tf.nn.log_softmax(self.logits) * self.probs, axis=-1)
+    entropies = self.probs * tf.log(self.probs)
+    return -tf.reduce_sum(tf.boolean_mask(entropies, tf.is_finite(entropies)), axis=-1)
 
   def _mode(self):
     ret = tf.argmax(self.logits, axis=self._batch_rank)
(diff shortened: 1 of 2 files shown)


harrismirza:
CLA is done

tensorflow_probability/python/distributions/categorical.py
```diff
@@ -299,8 +299,8 @@ def _log_prob(self, k):
                                                            logits=logits)
 
   def _entropy(self):
-    return -tf.reduce_sum(
-        tf.nn.log_softmax(self.logits) * self.probs, axis=-1)
+    entropies = self.probs * tf.log(self.probs)
+    return -tf.reduce_sum(tf.boolean_mask(entropies, tf.is_finite(entropies)), axis=-1)
```
langmore: Thank you for the contribution.  I see two issues here.

First, I see how this happens when probs is zero and someone does `tf.log(self.probs)`.  However, `tf.nn.log_softmax(self.logits)` is different.  What am I missing?

Second, note that `tf.boolean_mask(entropies, tf.is_finite(entropies))` will return a rank 1 Tensor, and then reduction over axis = -1 will return a scalar.  This is not what you want...you need to reduce only over the last dimension.  I guess the tests were not complete enough to catch this.

harrismirza: Sorry about the commit being wrong, I'm quite new to Tensorflow.

Log Softmax is implemented as `logits - log(reduce_sum(exp(logits), axis))` according to the API docs, and when a Categorical Distribution is created with the probabilities `[0, 0.5, 0.5]`, it stores the logits as `[-inf, -0.69..., -0.69...]`.

So when calculating Log Softmax for the first value it calculates:
= `-inf - log(e^-inf + e^-0.69 + e^0.69)
= -inf - log(0 + 0.5 + 0.5)
= -inf - log(1)
= -inf - 0
= -inf
`
And when this is multiplied by the probability of 0, python evaluates -inf * 0 as -inf.

axch: I suggest just computing the entropy terms with `tf.xlogy` and dispensing with the mask, like this:

```suggestion
    return -tf.reduce_sum(tf.xlogy(self.probs, self.probs), axis=-1)
```

`xlogy` is designed for exactly this: by definition, `xlogy(0, 0) == 0`, not `NaN`.

For us to approve the pull request, it would also be great if you could add a unit-test of entropy in the presence of zeros that fails before the change and passes after.

Thanks again for your contribution!

tensorflow_probability/python/distributions/categorical.py
```diff
@@ -299,8 +299,8 @@ def _log_prob(self, k):
                                                            logits=logits)
 
   def _entropy(self):
-    return -tf.reduce_sum(
-        tf.nn.log_softmax(self.logits) * self.probs, axis=-1)
+    entropies = self.probs * tf.log(self.probs)
```
brianwa84: entropies looks unused

ebrevdo: remove this line.

tensorflow_probability/python/distributions/categorical.py
```diff
@@ -299,8 +299,8 @@ def _log_prob(self, k):
...
+    return -tf.reduce_sum(tf.xlogy(self.probs, self.probs), axis=-1)
```
ebrevdo: can you change this to:

```
return -tf.reduce_sum(tf.math.multiply_no_nan(tf.nn.log_softmax(self.logits), self.probs), axis=-1)
```

thanks!

PR #245, opened 2018-12-07
Multivariate normal with linear operator support

Bonnevie:
TL;DR new version of `MultivariateNormalFullCovariance` that is backwards-compatible but also able to take `LinearOperator`-type covariance matrices as input.

I have previously touched on the need for a way to work with structured covariance matrices in both issue #161 and the closed PR #184 which demonstrated a prototype of this design. The concerns raised in the PR was that there were already too many variants of the `MultivariateNormal`. As suggested in the PR, this new version instead revamps the standard `MultivariateNormalFullCovariance` with a version that can take both tensors and linear operators as input when specifying the covariance. I have not touched the existing tests, so as it stands this new version should be completely compatible with existing code (assuming test coverage).
### Design decisions
I was unsure whether it would be worthwhile to inherit from any of the existing multivariate normal classes as `MultivariateNormalFullCovariance` used to do. I ultimately decided not to, basing it on the similar `MultivariateStudentT` instead, which was already using a `LinearOperator`-based design. This was mostly to ensure that a cholesky-factor was not computed in cases where that would not be efficient or desirable.  

 ### Incorporating `MultivariateNormalTril`
This could eventually subsume the need for `MultivariateNormalTril`, but currently `MultivariateNormalFullCovariance` depends on it for sampling which is most easily done with a Cholesky factor. In addition, `MultivariateNormalTril` is slightly more efficient since it only has to solve the system `inv(L)*x` once and then take a norm, as opposed to a linear operator-based solve which will always solve twice even if it knows the cholesky decomposition `inv(L*L')*x`. Since `LinearOperator` is getting cholesky support, it might make sense to also have special methods for `x'*A*x` and `x'*inv(A)*x` as well as a PSD linear operator specified via a cholesky factor.

### DIFF
--- a/tensorflow_probability/python/distributions/mvn_full_covariance.py
+++ b/tensorflow_probability/python/distributions/mvn_full_covariance.py
@@ -24,7 +24,7 @@ from tensorflow_probability.python.internal import dtype_util
 from tensorflow.python.ops import control_flow_ops
 
 import functools
-
+import numpy as np
 
 from tensorflow_probability.python import math
 from tensorflow_probability.python.distributions import kullback_leibler
@@ -49,7 +49,7 @@ class MultivariateNormalFullCovariance(distribution.Distribution):
 
   The Multivariate Normal distribution is defined over `R^k` and parameterized
   by a (batch of) length-`k` `loc` vector (aka "mu") and a (batch of) `k x k`
-  `covariance` matrices that are the covariance.
+  `covariance_matrix` matrices that are the covariance.
   This is different than the other multivariate normals, which are parameterized
   by a matrix more akin to the standard deviation.
 
@@ -66,10 +66,10 @@ class MultivariateNormalFullCovariance(distribution.Distribution):
   where:
 
   * `loc` is a vector in `R^k`,
-  * `covariance` is an `R^{k x k}` symmetric positive definite matrix,
+  * `covariance_matrix` is an `R^{k x k}` symmetric positive definite matrix,
   * `Z` denotes the normalization constant.
 
-  Additional leading dimensions (if any) in `loc` and `covariance` allow
+  Additional leading dimensions (if any) in `loc` and `covariance_matrix` allow
   for batch dimensions.
 
   #### Examples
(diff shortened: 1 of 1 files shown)


Bonnevie:
@csuter I think you mentioned you were working on this, right?

csuter:
Hi @Bonnevie, thank you so much for following up with this update to the previous PR! I have not done any actual work on it, so we're not overlapping efforts or anything (in case that was your concern in asking if I was working on it).

I am traveling this week and may be time constrained but will take a good look at this ASAP. Anyone else from the team, is of course, welcome to comment. @brianwa84 and @langmore may wish to comment (as they were looking at the previous iteration).

Bonnevie:
@csuter yeah, on rereading the old thread I was a bit worried that you'd had been putting time into it too. Hope it comes close to the design you were envisioning. 

There was a problem with sampling in the new class which I have just fixed - apparently the tests don't verify that sampling works?

tensorflow_probability/python/distributions/mvn_full_covariance.py
```diff
@@ -96,10 +101,10 @@ class MultivariateNormalFullCovariance(mvn_tril.MultivariateNormalTriL):
   # Initialize a 2-batch of 3-variate Gaussians.
   mu = [[1., 2, 3],
         [11, 22, 33]]              # shape: [2, 3]
-  covariance_matrix = ...  # shape: [2, 3, 3], symmetric, positive definite.
+  covariance = ...  # shape: [2, 3, 3], symmetric, positive definite.
   mvn = tfd.MultivariateNormalFullCovariance(
       loc=mu,
-      covariance=covariance_matrix)
+      covariance=covariance)
```
brianwa84: the param can't be renamed

brianwa84: similar above

tensorflow_probability/python/distributions/mvn_full_covariance.py
```diff
@@ -118,71 +123,219 @@ def __init__(self,
...
+      covariance_matrix = tf.convert_to_tensor(
+        covariance_matrix, name="covariance_matrix")
+      if validate_args:
+        covariance_matrix = control_flow_ops.with_dependencies([
+          tf.assert_near(
+              covariance_matrix,
+              tf.matrix_transpose(covariance_matrix),
+              message="Matrix was not symmetric")
+          ], covariance_matrix)
+      # LinearOperator applies cholesky which will fail if not PSD
```
brianwa84: Should probably say MVNLinearOperator, not LinearOperator

tensorflow_probability/python/distributions/mvn_full_covariance.py
```diff
@@ -118,71 +123,219 @@ def __init__(self,
...
-                    covariance_matrix,
-                    tf.matrix_transpose(covariance_matrix),
-                    message="Matrix was not symmetric")
-            ], covariance_matrix)
-          # No need to validate that covariance_matrix is non-singular.
-          # LinearOperatorLowerTriangular has an assert_non_singular method that
-          # is called by the Bijector.
-          # However, cholesky() ignores the upper triangular part, so we do need
-          # to separately assert symmetric.
-          scale_tril = tf.cholesky(covariance_matrix)
```
brianwa84: I'm a bit concerned that in eager mode we will be recomputing this a lot.

tensorflow_probability/python/distributions/mvn_full_covariance.py
```diff
@@ -118,71 +123,219 @@ def __init__(self,
...
+    return self._covariance_matrix.to_dense()
+
+  def _stddev(self):
+    return tf.sqrt(self._variance())
+
+  def _sample_n(self, n, seed=None):
+    seed = seed_stream.SeedStream(seed, salt="multivariate normal")
+
+    loc = _broadcast_to_shape(self.loc, self._sample_shape())
+    mvn = mvn_tril.MultivariateNormalTriL(
```
brianwa84: it seems like this won't be able to (efficiently) sample a low rank + diag because of the cholesky.

100376348:
Hi @Bonnevie, I am having a problem with the original MultivariateNormalFullCovariance code (Cholesky decomposition was not successful) and I think it might be a bug. Do you think your code could solve this problem? Thanks!

PR #254, opened 2018-12-21
Bijector mapping  deep tuple fix

hartikainen:
I have similar [PR open on the tensorflow side](https://github.com/tensorflow/tensorflow/pull/21534), but haven't received any response to it for months. I thought that now that the bijector lives in probability, maybe it would be best to move this here.

Right now, Bijector _Mapping's _deep_tuple doesn't allow nested dicts. This problem causes some of the conditioning to not work. See: https://github.com/tensorflow/tensorflow/issues/21543. This PR is a possible way to fix the issue. I'm happy to take feedback on this, and potentially fix the problem in some other way if needed.

### DIFF
--- a/tensorflow_probability/python/bijectors/conditional_bijector_test.py
+++ b/tensorflow_probability/python/bijectors/conditional_bijector_test.py
@@ -91,55 +91,33 @@ class ConditionalBijectorTest(tf.test.TestCase):
     class ConditionalChain(tfb.ConditionalBijector, tfb.Chain):
       pass
 
-    test_bijector = _TestBijector()
-    passthrough_bijector = _TestPassthroughBijector()
-    chain = ConditionalChain((test_bijector, passthrough_bijector))
+    for name in ["forward", "inverse"]:
+      test_bijector = _TestBijector()
+      passthrough_bijector = _TestPassthroughBijector()
+      chain_components = (test_bijector, passthrough_bijector)
+      if name == "inverse":
+          chain_components = chain_components[::-1]
+      chain = ConditionalChain(chain_components)
 
-    name = "forward"
-    assert not passthrough_bijector._called[name]
-    method = getattr(chain, name)
-    with self.assertRaisesRegexp(ValueError, name + ".*b1.*b2"):
-      method(
-          1.,
-          test_bijector={"arg1": "b1", "arg2": "b2"},
-          test_passthrough_bijector={"arg1": "b1", "arg2": "b2"})
-    assert passthrough_bijector._called[name], name
+      assert not passthrough_bijector._called[name]
+      method = getattr(chain, name)
+      with self.assertRaisesRegexp(ValueError, name + ".*b1.*b2"):
+        method(
+            1.,
+            test_bijector={"arg1": "b1", "arg2": "b2"},
+            test_passthrough_bijector={"arg1": "b1", "arg2": "b2"})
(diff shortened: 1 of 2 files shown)


tensorflow_probability/python/bijectors/conditional_bijector_test.py
```diff
@@ -48,6 +48,32 @@ def _forward_log_det_jacobian(self, _, arg1, arg2):
     raise ValueError("forward_log_det_jacobian", arg1, arg2)
 
 
+class _TestPassthroughBijector(_TestBijector):
```
SiegeLordEx: I understand you're trying to reproduce the test case from your bug report, but it feels like it would be a lot simpler to just directly test the nested dicts by passing dict args to _TestBijector. What do you think?

tensorflow_probability/python/bijectors/conditional_bijector_test.py
```diff
@@ -48,6 +48,32 @@ def _forward_log_det_jacobian(self, _, arg1, arg2):
...
+  def __init__(self, *args, **kwargs):
+      super(_TestPassthroughBijector, self).__init__(*args, **kwargs)
```
SiegeLordEx: The lines in this method are indented too much, it should be a total of 4 spaces (2 per indent level).

tensorflow_probability/python/bijectors/conditional_bijector_test.py
```diff
@@ -48,6 +48,32 @@ def _forward_log_det_jacobian(self, _, arg1, arg2):
...
+      self._called = {
+          name: False for name in
+          ('forward', 'inverse',
```
SiegeLordEx: Would you mind using " instead of ', for consistency with the rest of the file (our internal lints don't like us being inconsistent).

tensorflow_probability/python/bijectors/conditional_bijector_test.py
hartikainen: I added simpler `testNestedCondition` test (see https://github.com/tensorflow/probability/pull/254/commits/462a231df8fdaef093b44b2af1848237d91eadcd). Should I remove the explicit conditional chain test or do you want me to keep it?

hartikainen:
@SiegeLordEx Thanks a lot for your feedback! All your comments should be addressed with the latest changes.

tensorflow_probability/python/bijectors/conditional_bijector_test.py
SiegeLordEx: Let's remove it if you don't mind, I think testNestedCondition is a little more clear and covers what we want to test.

tensorflow_probability/python/bijectors/conditional_bijector_test.py
```diff
@@ -18,6 +18,8 @@
 from __future__ import division
 from __future__ import print_function
 
+import collections
```
SiegeLordEx: Just one last thing, this is now unused.

hartikainen: Good catch

SiegeLordEx:
Perfect. Now I'll pass it through our internal review process, and when that goes through this'll get automatically merged. Thanks a lot!

hartikainen:
Great. Thanks a lot for the review and feedback!

PR #261, opened 2019-01-10
Fix typo in tfd.GaussianProcess API docs

darthdeus:
A minor change I found following the API examples. `tfp.GaussianProcess` does not exist, as is also visible in the example just above it. The correct module is `tfd`.

### DIFF
--- a/tensorflow_probability/python/distributions/gaussian_process.py
+++ b/tensorflow_probability/python/distributions/gaussian_process.py
@@ -169,10 +169,10 @@ class GaussianProcess(mvn_linear_operator.MultivariateNormalLinearOperator):
       amplitude=tf.get_variable('amplitude', np.float32),
       length_scale=tf.get_variable('length_scale', np.float32))
 
-  gp = tfp.GaussianProcess(kernel, observed_index_points)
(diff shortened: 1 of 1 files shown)


csuter:
Thanks for the fix! These api docs are auto-generated from the python docstrings -- would you mind fixing there? Then we can regenerate the docs.

darthdeus:
@csuter Done, including one more typo fix in the same docstring :)

csuter:
Beautiful, thanks!

I'll pull this into our internal VCS, after which it'll be autopushed out and this PR will be marked as merged/closed.

Thanks again!

darthdeus:
@csuter Sorry for bringing this up again, but I somehow managed to miss that there is another issue in the code example just a few lines later. The example first defines `neg_log_likelihood`, but then uses a variable `nll` (which is undefined).

I'll make a new PR, but wanted to ask if I should make all of it `nll`, or all of it `neg_log_likelihood`?

csuter:
D'oh! Thanks for the careful reading and proactive fixes.

We usually err on the side of not abbreviating (at least when i'm not apparently being lazy! 😬) so let's go with `neg_log_likelihood`.

Thanks!

PR #272, opened 2019-01-18
Support dynamic shape in `sqrt_with_finite_grads`.

oahziur:
I get an error when I use a tensor with dynamic shape as the index_points in `tfd.GaussianProcess`. This change fixed the problem.


oahziur:
I signed it!

PR #274, opened 2019-01-22
Overhaul Travis configuration.

csuter:
1. Create separate "build stages" for lint and test. This lets lint
     failures short circuit the build. It also prevents them being run
     twice, once for each python version.
  2. Move away from using 'install' entries in .travis.yml. This
     prevents, e.g., installing bazel before running lints. Instead, now
     test scripts do their own installation of things. Travis's install
     mechanism is kinda fancy (supports language-aware concepts like
     looking around for requirements.txt, for python) but we weren't
     really using it the "right way" and anyway our build configuration
     is pretty unusual for an OSS python project. This change also makes
     the installation steps "quiet" to reduce log cruft.

Here's an example of the tests running on my clone of the TFP repo: Note
that there are 2 stages, "Lint" and "Test".

Example of lint failing and stopping build:
  https://travis-ci.org/csuter/probability/builds/483136955

Example of everything running to completion:
  https://travis-ci.org/csuter/probability/builds/483138401

Example of a failure before these changes (only lint failed, but it's hard to spot the error message and the whole thing continued running for ~30min):
  https://travis-ci.org/tensorflow/probability/builds/482038723


csuter:
Oh, I also 
  1. got rid of our explicit installation of openjdk8 -- it's already there by default and takes a while to download.
  2. added explicit --notest_keep_going to cause tests to stop running if something fails (we used to have --test_keep_going and removed it...but I think that was the default! so removing it didn't do anything :P)

.travis.yml
```diff
@@ -17,21 +17,31 @@ language: python
...
   global:
-    # We only test against the nightly version of TensorFlow.
-    - TF_VERSION="tf-nightly"
     - NUM_SHARDS=2
   matrix:
-    # We shard our tests to avoid timeouts. Additionally, only shard 0 runs the
-    # lints.
     - SHARD=0
     - SHARD=1
-install:
```
SiegeLordEx: I appreciate the reasoning behind not using this, but I still think a separate install script (placed in the script's section) has benefits. One specific benefit I have in mind is that in its log display, it'll collapse the outputs of individual scripts so we can be verbose as we want as in the install script. In particular, I like printing out the packages that pip installed so we can have an easier time debugging things if anything went wrong. Naturally, it also collapses the `set -x` spam.

.travis.yml
```diff
@@ -17,21 +17,31 @@ language: python
...
 python:
   - "2.7"
   - "3.6"
 env:
   global:
-    # We only test against the nightly version of TensorFlow.
-    - TF_VERSION="tf-nightly"
     - NUM_SHARDS=2
   matrix:
-    # We shard our tests to avoid timeouts. Additionally, only shard 0 runs the
```
SiegeLordEx: Any reason not to keep the first sentence?

testing/run_tests.sh
```diff
@@ -43,8 +76,14 @@ shard_tests=$(bazel query 'tests(//tensorflow_probability/...)' |
...
-  --action_env=PATH --action_env=LD_LIBRARY_PATH \
-  --test_output=errors
+echo "${shard_tests}" \
+  | xargs bazel test \
+    --copt=-O3 \
+    --copt=-march=native \
+    --notest_keep_going \
+    --test_tag_filters=-gpu,-requires-gpu-sm35 \
+    --test_timeout 300,450,1200,3600 \
+    --build_tests_only \
```
SiegeLordEx: I actually think this isn't needed, since we explicitly pass only tests to bazel test.

csuter: ah good point!

.travis.yml
csuter: Hm, i don't think the collapsing thing is true for multiple scripts (i think it is for, e.g., separate install, before_script, etc sections) but i'll play around with it (tomorrow) and see if i can finagle some better collapsing. FWIW I believe both the apt and pip install commands will still print errors if anything goes wrong. If it's just to see what *was* installed, we could also add a 'pip freeze' or something. But I'm fine with the verbosity if I can figure out the collapsability.

.travis.yml
csuter: No, just an oversight. Will re-add.

.travis.yml
csuter: also someone on the internet points to [this](https://github.com/travis-ci/travis-rubies/blob/9f7962a881c55d32da7c76baefc58b89e3941d91/build.sh#L38-L44) as an example of how to add custom folds by hand. could give that a try

csuter: ok tried it with the start_fold/end_fold commands. Here's an example output: https://travis-ci.org/csuter/probability/jobs/483687631

WDYT?

SiegeLordEx: That looks great, thanks!

PR #284, opened 2019-01-31
Deep exponential family Notebook

matthew-mcateer:
Added notebook for a Deep Exponential Family, based on [this Edward1 example](https://github.com/blei-lab/edward/blob/master/examples/deep_exponential_family.py) and [this TFP example](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/examples/deep_exponential_family.py) (without [edward2](https://www.tensorflow.org/probability/api_docs/python/tfp/edward2), though a notebook that does work while using edward2 is available [in this colab](https://colab.research.google.com/drive/1O_B6otzPaBCPrlP189e-Z19Dly3VG5fh)) and [this paper](http://www.cs.toronto.edu/~lcharlin/papers/def_aistats.pdf), created by [the Blei Lab](http://www.cs.columbia.edu/~blei/)


matthew-mcateer:
Okay, looking at all the other file commits that carried over, I don't think this one can be approved until the other 10 are committed.

PR #309, opened 2019-02-25
Added CAPM and GLM Example (Jupyter Notebook)

matthew-mcateer:
Added example on how to set up the CAPM model for stocks, and how this compares to the simpler GLM interface for Tensorflow Probability.

This one was NOT based on an example from the edward1 library.


Pindar777:
@matthew-mcateer Thanks for adding a CAPM model.
I have some general remarks which I'd like to share with you:
In the standard preamble of this and all the other examples there is IMHO some room for improvement:
- please remove "For the most recent nightly GPU-connected version of TFP: pip3 install -q tfp-nightly-gpu"
- please think of a way to make the code either solely working with TF 2.0 or working with "both worlds"

In order to achieve the latter find some ideas as follows:

- the very helpful evaluate function should work with TF 2.0 by running these SLOC
`
def evaluate(tensors):
    if tf.executing_eagerly():
        return tf.nest.pack_sequence_as(
            tensors,
            [t.numpy() if tf.is_tensor(t) else t
             for t in tf.nest.flatten(tensors)])
    return sess.run(tensors)`

- after importing tensorflow something like that could be integrated:
`import tensorflow as tf
from tensorflow.python import tf2
if not tf2.enabled():
  tf.enable_v2_behavior()
  assert tf2.enabled()`

PR #312, opened 2019-03-01
indicate that these arguments are not optional

skeydan:
Hi, I just noticed this during creating R bindings for bijectors... I'm imagining it would be clearer not to have defaults for these arguments?


skeydan:
Hi, I'm not sure what's the state, I signed the PR, although the email address is different from the one I use on github.
However, I was not able to add an "alternate" email address to my Google account as described on the help page linked from the PR page (https://support.google.com/accounts/answer/176347)

axch:
Thanks for the catch!  (And sorry for the slow reply.)  We'll pull this in shortly.

skeydan:
thanks!

PR #326, opened 2019-03-11
Duplicate parenthesis fix in MixtureSameFamily doc example

cyrilou242:
A duplicate parenthesis is at the end of line 8 in the example code.
Removed it.


axch:
Thank you for your attention!  That file (and most everything in the g3doc/ directory) is actually auto-generated, and it looks like the mistake has already been fixed in the source docstring: https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/layers/distribution_layer.py#L1152.  The generated stuff (and the web site) is out of sync because we only rebuild them when we cut a new stable release, so you should see this fixed in TFP 0.7.

Thanks again!

PR #335, opened 2019-03-21
fix(docs): typos in HMC SAEM example

janosh:
- fixes two typos in the [HMC SAEM example](https://www.tensorflow.org/probability/api_docs/python/tfp/mcmc/HamiltonianMonteCarlo#estimate_parameters_of_a_more_complicated_posterior)


janosh:
I signed the CLA.

SiegeLordEx:
Hi @janosh, thanks for your PR. These docs are actually generated from the Python docstrings, here https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/mcmc/hmc.py#L353-L372. However, it appears that this typos has already been fixed (and that example reworked a little bit), and the docs are just out of date (i.e. need to be regenerated). Currently we only regenerate the docs once we make a release, which may come soon.

janosh:
I see. Looking forward to the new release.

PR #340, opened 2019-03-25
fix broken example GaussianProcessRegressionModel

trainorpj:
The example in [Marginalization of model hyperparameters](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/GaussianProcessRegressionModel#marginalization_of_model_hyperparameters) section crashes, because the `index_points` variable doesn't have the correct shape. This change fixes that.

In this snippet, `index_points` is never set. However, it is set in a previous snippet, but the shape is not correct, and this example crashes when it goes to plot the results.

[Here's the working code in action](https://repl.it/@PjTrainor/tfp-gpr-text)

Per the CONTRIBUTING.md, this seemed like it was small enough for a PR without an Issue 🙂

### DIFF
--- a/tensorflow_probability/python/distributions/gaussian_process_regression_model.py
+++ b/tensorflow_probability/python/distributions/gaussian_process_regression_model.py
@@ -292,7 +292,7 @@ class GaussianProcessRegressionModel(
   gprm = tfd.GaussianProcessRegressionModel(
       # Batch of `num_results` kernels parameterized by the MCMC samples.
       kernel=psd_kernels.ExponentiatedQuadratic(amplitudes, length_scales),
-      index_points=np.linspace(-2., 2., 200)[..., np.newaxis],
+      index_points=index_points,
       observation_index_points=observation_index_points,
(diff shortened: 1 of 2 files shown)


trainorpj:
I signed it! ✍️

srvasude:
Hi!
  Our docs are autogenerated from the doc strings in the corresponding file. Would it be possible to move your fix to the GaussianProcess docstring? Thanks!

tensorflow_probability/python/distributions/gaussian_process_regression_model.py
```diff
@@ -288,6 +288,7 @@ def unnormalized_log_posterior(amplitude, length_scale, noise_variance):
 
   # Now we can sample from the posterior predictive distribution at a new set
   # of index points.
+  index_points = np.linspace(-1., 1., 200)[..., np.newaxis]
```
srvasude: Below a numpy object is passed in to index_points in the gprm. Can you replace that with `index_points=index_points`

PR #351, opened 2019-04-05
Minor bug fix in the example code

idofr:
The print at the end of the second example had a grammar error. The first `plt` call was also wrong

### DIFF
--- a/tensorflow_probability/g3doc/api_docs/python/tfp/mcmc/HamiltonianMonteCarlo.md
+++ b/tensorflow_probability/g3doc/api_docs/python/tfp/mcmc/HamiltonianMonteCarlo.md
@@ -231,13 +231,13 @@ with tf.Session() as sess:
         avg_acceptance_ratio,
     ], feed_dict={weights_chain_start: weights_[iter_]})
     print('iter:{:>2}  loss:{: 9.3f}  scale:{:.3f}  '
-          'step_size:{:.4f}  avg_acceptance_ratio:{:.4f}').format(
+          'step_size:{:.4f}  avg_acceptance_ratio:{:.4f}'.format(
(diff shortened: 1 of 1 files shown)


idofr:
I signed it!

SiegeLordEx:
Hi @idofr, thanks for the pull request!

This documentation is automatically generated every release from the Python docstring (in `python/mcmc/hmc.py`). Could you make your pull request changing that? Note that we've made a bunch of changes to that example, so some of your changes may be obsolte.

idofr:
Sure, I'll try to do it later. Do you know where I can find these more recent changes you've made?

SiegeLordEx:
https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/mcmc/hmc.py

idofr:
Alright, this does make my changes rather oblique. Then i guess we can just drop the merge requeset.
Thanks for the support :)

janosh:
@SiegeLordEx [Deja vu!](https://github.com/tensorflow/probability/pull/335) :) Out of curiosity, is there a reason not to adopt [continuous integration](https://en.wikipedia.org/wiki/Continuous_integration), i.e. rebuilding the docs every time an update is pushed? Seems like outdated docs result in multiple redundant PRs that could be avoided with CI.

SiegeLordEx:
I've suggested the same internally: the main issue is logistical, we host our documentation on tensorflow.org, and there's no good way to do nightly documentation there. We also can't adjust the docs for the API (in this case), because sometimes new API is added. Maybe we can try a little harder about hosting the nightly docs somewhere though, this is a somewhat annoying issue.

janosh:
Well good luck with your suggestion. :)

PR #368, opened 2019-04-16
fixed typos and bugs in gaussian_process doc

trainorpj:
This PR fixes a typo and some bugs in the `GaussianProcess` class.

The typo was simple, but I'm not sure about the bugs 🐛. This is what happens:

```py
# TypeError: int() argument must be a string, a bytes-like object or a number, not 'type'
tf.get_variable('amplitude', np.float32)

# this works, but crashes on a dtype conflict with another variable
tf.get_variable('amplitude', shape=(), dtype=np.float32)

# this works!
tf.get_variable('amplitude', shape=(), dtype=np.float64)
```

See a [running repl.it](https://repl.it/@PjTrainor/tfp-gpr-example) for more details. **Note**: sometimes this crashes with `Input matrix is not invertible.`

The repl is running on tensorflow-1.13.1, so maybe there's an issue with that 🤷‍♂️

### DIFF
--- a/tensorflow_probability/python/distributions/gaussian_process.py
+++ b/tensorflow_probability/python/distributions/gaussian_process.py
@@ -148,7 +148,7 @@ class GaussianProcess(mvn_linear_operator.MultivariateNormalLinearOperator):
       kernel=kernel,
       index_points=index_points,
       observation_noise_variance=.05)
-  noisy_samples = noise_gp.sample(10)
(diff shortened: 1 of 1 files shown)


csuter:
Hi, can you please rebase your branch? Then I'll be able to pull it in.

csuter:
actually i think i'll be able to pull it in without a rebase; ignore for now

PR #375, opened 2019-04-23
Dual averaging step size

ColCarroll:
Wanted to push this early -- I am pattern matching a lot from the nicely documented `SimpleStepSizeAdaptation` class, and may ask questions here.

This implements dual averaging step size. The reference is Section 3.2 in "The No-U-Turn Sampler: Adaptively Setting Path Lengths in Hamiltonian Monte Carlo.", and a reference implementation using `autograd` is [here](https://colindcarroll.com/2019/04/21/step-size-adaptation-in-hamiltonian-monte-carlo/).

Here are two experiments comparing warmup with this and the `SimpleStepSizeAdaptation`:

image
image

TODO:
[ ] Clean up reused code
[ ] Add tests
[ ] Update variable names (descriptive are preferred to the names from the paper?)
[ ] Update documentation
[ ] Update style

Code for the experiments above. I can add the plotting code if desired:

```python
import tensorflow as tf
tfd = tfp.distributions

print(tfp.__version__)  # 0.7.0-dev
print(tf.__version__)  # 2.0.0-dev20190422

target_log_prob_fn = tfd.Normal(loc=0., scale=1.1).log_prob
num_results = 1000
num_chains = 64
step_size = 0.01

inner_kernel = tfp.mcmc.HamiltonianMonteCarlo(target_log_prob_fn=target_log_prob_fn, num_leapfrog_steps=8, step_size=step_size)

simple_kernel = tfp.mcmc.SimpleStepSizeAdaptation(inner_kernel=inner_kernel, num_adaptation_steps=num_results)
averaging_kernel = tfp.mcmc.DualAveragingStepSizeAdaptation(inner_kernel=inner_kernel, num_adaptation_steps=num_results)

def simple_trace_fn(x, pkr):
    return [pkr.inner_results.accepted_results.step_size, pkr.inner_results.log_accept_ratio]

def averaging_trace_fn(x, pkr):
    return [pkr.inner_results.accepted_results.step_size, pkr.inner_results.log_accept_ratio, pkr.log_averaging_step]


_, [simple_step_size, simple_log_accept_ratio] = tfp.mcmc.sample_chain(
    num_results=num_results,
    num_burnin_steps=0,
    current_state=tf.zeros(num_chains),
    kernel=simple_kernel,
    trace_fn=simple_trace_fn)

_, [averaging_step_size, averaging_log_accept_ratio, log_averaging_step] = tfp.mcmc.sample_chain(
    num_results=num_results,
    num_burnin_steps=0,
    current_state=tf.zeros(num_chains),
    kernel=averaging_kernel,
    trace_fn=averaging_trace_fn)
```

@junpenglao may be interested in this.

### DIFF
--- a/tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
+++ b/tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
@@ -465,8 +465,8 @@ def _maybe_validate_target_accept_prob(target_accept_prob, validate_args):
   if not validate_args:
     return target_accept_prob
   with tf.control_dependencies([
-      tf.assert_positive(
-          target_accept_prob, message='`target_accept_prob` must be > 0.'
+      tf.assert_greater(
+          target_accept_prob, 0., message='`target_accept_prob` must be > 0.'
       ),
       tf.assert_less(
           target_accept_prob,
--- a/tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
+++ b/tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
@@ -298,10 +298,10 @@ class DualAveragingStepSizeAdaptation(kernel_base.TransitionKernel):
   def _one_step_part(
       self,
       *parts,
-      log_accept_prob_rank,
-      log_accept_prob,
-      target_accept_prob,
-      previous_kernel_results):
+      log_accept_prob_rank=None,
+      log_accept_prob=None,
+      target_accept_prob=None,
+      previous_kernel_results=None):
     """Compute new step sizes for each step size part.
 
     If step size part has smaller rank than the corresponding state part, then
--- a/tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
+++ b/tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
@@ -317,12 +317,14 @@ class DualAveragingStepSizeAdaptation(kernel_base.TransitionKernel):
       error_sum_parts = tf.nest.flatten(previous_kernel_results.error_sum)
       log_averaging_step_parts = tf.nest.flatten(
           previous_kernel_results.log_averaging_step)
-      shrinkage_target_parts = tf.nest.flatten(previous_kernel_results.shrinkage_target)
+      shrinkage_target_parts = tf.nest.flatten(
+          previous_kernel_results.shrinkage_target)
       new_error_sum_parts = []
       new_step_size_parts = []
       new_log_averaging_step_parts = []
-      for step_size_part, state_part, error_sum_part, log_averaging_step_part, shrinkage_target_part in zip(
-            step_size_parts, state_parts, error_sum_parts, log_averaging_step_parts, shrinkage_target_parts):
+      for parts in zip(
+          step_size_parts, state_parts, error_sum_parts,
+          log_averaging_step_parts, shrinkage_target_parts):
         # Compute new step sizes for each step size part. If step size part has
         # smaller rank than the corresponding state part, then the difference is
         # averaged away in the log accept prob.
@@ -358,6 +360,8 @@ class DualAveragingStepSizeAdaptation(kernel_base.TransitionKernel):
         # states, and B are the shared dimensions of all the states, which are
         # equal to the step size. When we subtract B, we will always get a
         # number >= L, which means we'll get the full reduction we want.
+        (step_size_part, state_part, error_sum_part,
+         log_averaging_step_part, shrinkage_target_part) = parts
         num_reduce_dims = tf.minimum(
             log_accept_prob_rank,
             tf.rank(state_part) - tf.rank(step_size_part))
@@ -379,9 +383,10 @@ class DualAveragingStepSizeAdaptation(kernel_base.TransitionKernel):
         step = tf.cast(previous_kernel_results.step, initial_t.dtype) + 1.
         soft_t = initial_t + step
 
-        new_log_step = (shrinkage_target_part -
-            (tf.reshape(new_error_sum, shrinkage_target_part.shape) * tf.math.sqrt(step)) /
-            (soft_t * previous_kernel_results.exploration_shrinkage))
+        new_log_step = (shrinkage_target_part - (
+            tf.reshape(new_error_sum, shrinkage_target_part.shape) *
+            tf.math.sqrt(step)) / (
+                soft_t * previous_kernel_results.exploration_shrinkage))
 
         eta = tf.math.pow(step, -previous_kernel_results.decay_rate)
         new_log_averaging_step = (eta * new_log_step +
@@ -443,14 +448,17 @@ class DualAveragingStepSizeAdaptation(kernel_base.TransitionKernel):
             reduced_log_accept_prob, reduce_indices, keepdims=True)
         error_sum.append(tf.zeros_like(reduced_log_accept_prob, dtype=dtype))
         log_averaging_step.append(tf.zeros_like(step_size_part, dtype=dtype))
-        shrinkage_target.append(tf.math.log(tf.constant(10, dtype=dtype) * step_size_part))
+        shrinkage_target.append(
+            tf.math.log(tf.constant(10, dtype=dtype) * step_size_part))
 
       return DualAveragingStepSizeAdaptationResults(
           inner_results=inner_results,
           step=tf.constant(0, dtype=tf.int32),
-          target_accept_prob=tf.cast(self.parameters['target_accept_prob'], dtype),
+          target_accept_prob=tf.cast(self.parameters['target_accept_prob'],
+                                     dtype),
           shrinkage_target=shrinkage_target,
-          exploration_shrinkage=tf.cast(self.parameters['exploration_shrinkage'], dtype),
+          exploration_shrinkage=tf.cast(
+              self.parameters['exploration_shrinkage'], dtype),
           initial_t=tf.cast(self.parameters['initial_t'], dtype),
           decay_rate=tf.cast(self.parameters['decay_rate'], dtype),
           error_sum=error_sum,
--- a/tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
+++ b/tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
@@ -404,7 +404,6 @@ class DualAveragingStepSizeAdaptation(kernel_base.TransitionKernel):
       log_accept_prob = self.log_accept_prob_getter_fn(new_inner_results)
       target_accept_prob = previous_kernel_results.target_accept_prob
 
-      state_parts = tf.nest.flatten(current_state)
       step_size = self.step_size_getter_fn(new_inner_results)
       step_size_parts = tf.nest.flatten(step_size)
       log_accept_prob_rank = tf.rank(log_accept_prob)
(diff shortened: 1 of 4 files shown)


ColCarroll:
I signed it!

tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
```diff
@@ -0,0 +1,299 @@
+"""DualAveragingStepSizeAdaptation TransitionKernel."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import collections
+
+import numpy as np
+import tensorflow as tf
```
jvdillon: Would it work if you do:
`import tensorflow.compat.v2 as tf`

You'll need to update downstream things, eg, s/tf.compat.v1/tf/ etc.

If you can do this it will make the review process a lot easier. We use a backend conversion script to ensire TF2 compatibility and it can be a very nitpick script.

tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
```diff
@@ -0,0 +1,299 @@
...
+from tensorflow_probability.python.mcmc import kernel as kernel_base
+from tensorflow_probability.python.mcmc.internal import util as mcmc_util
+
+
+@mcmc_util.make_innermost_setter
+def _hmc_like_step_size_setter_fn(kernel_results, new_step_size):
+  return kernel_results._replace(
+    accepted_results=kernel_results.accepted_results._replace(
+      step_size=new_step_size
+    )
```
jvdillon: nit: Can you place these closing parens on the line above. In TFP we dont follow the open-paren style you use here. (Although it does have its merits, to be sure!)

tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
```diff
@@ -0,0 +1,299 @@
...
+    gamma=0.05,
+    t0=10.0,
+    kappa=0.75,
+    step_size_setter_fn=_hmc_like_step_size_setter_fn,
+    step_size_getter_fn=_hmc_like_step_size_getter_fn,
+    log_accept_prob_getter_fn=_hmc_like_log_accept_prob_getter_fn,
+    validate_args=False,
+    name=None,
+  ):
+    inner_kernel = mcmc_util.enable_store_parameters_in_results(inner_kernel)
```
jvdillon: We'll need to add class and init docstrings in the style of other TransitionKernel objects.

tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+class DualAveragingStepSizeAdaptation(kernel_base.TransitionKernel):
+  def __init__(
+    self,
+    inner_kernel,
+    num_adaptation_steps,
+    target_accept_prob=0.75,
+    gamma=0.05,
+    t0=10.0,
```
jvdillon: We try to avoid using greek in the API. Can you think of an english name? Ideally one which we've used elsewhere. (We try to limit the number nouns we use.)

tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
```diff
@@ -0,0 +1,299 @@
...
+        # left, so we do an additional reduction over dimensions where their
+        # shapes differ.
+        reduce_indices = _get_differing_dims(
+          reduced_log_accept_prob, step_size_part
+        )
+        reduced_log_accept_prob = _reduce_logmeanexp(
+          reduced_log_accept_prob, reduce_indices, keepdims=True
+        )
+
+        t = previous_kernel_results.t0 + tf.cast(previous_kernel_results.step, previous_kernel_results.t0.dtype)
```
jvdillon: nit: Line wrap @ 80char

tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
```diff
@@ -0,0 +1,299 @@
+"""DualAveragingStepSizeAdaptation TransitionKernel."""
```
jvdillon: Can you also add a unit test? You should be able to use SimpleStepSize as an example.

tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
junpenglao: maybe: 
gamma --> exploration_shrinkage 
kappa --> decay_rate

ColCarroll:
Thanks for the comments! I got the pylint check passing, and tightened up the parens a bit -- adding back docs, tests, and tidying next.

The only difficulty in using `tf.compat.v2 as tf` was `tf.name_scope` not taking `values` as an argument, and I could not find docs suggesting a migration (it looks like `name_scope` verifies the `values` are from the same graph). So

```python
    with tf.compat.v1.name_scope(
        mcmc_util.make_name(name, 'simple_step_size_adaptation', '__init__'),
        values=[target_accept_prob, adaptation_rate,
                num_adaptation_steps]) as name:
```
turned into something like
```python
    with tf.name_scope(
        mcmc_util.make_name(name, 'simple_step_size_adaptation', '__init__'),
    ) as name:
```

Let me know if _just removing the keyword argument_ was wrong and I can keep looking!

brianwa84:
Removing the values arg from name_scope is fine.

ColCarroll:
(this is still being worked on -- will push more updates soon)

janosh:
This looks really useful! Excited to try it out.

ColCarroll:
I think this is finally just about ready. I also have some thoughts about how to improve this for a follow-up.

I am trying to write unit tests, but am having trouble figuring out how to run tests: for what it is worth, I have a (conda) environment that can successfully run and fit models.

The short version is that I am not doing very well running `bazel`, but I can mostly get by using `pytest` instead. It would be helpful to have a minimal example of how to run the test suite, a single test file, and a single test.

Attempts:
------------
1. **Using pytest** Running
```bash
python -m pytest -v tensorflow_probability/python/mcmc/simple_step_size_adaptation_test.py
```
fails on `tensorflow_probability/python/mcmc/simple_step_size_adaptation_test.py::SimpleStepSizeAdaptationTest::testExample` because some `abseil` flags are not set.

2. **Using bazel** Installing `conda install -c conda-forge bazel` and running things like
```bash
bazel test --copt=-O3 --copt=-march=native tensorflow_probability/python/mcmc/simple_step_size_adaptation*
```
Give me a success message very quickly, and appear to build but not run any tests.

3. **Using available scripts** Running 
```bash
./testing/run_tests.sh 
```
Fails with unset flags. Running
```bash
SHARD=0 NUM_SHARDS=1 ./testing/run_tests.sh 
```
Ignores the libraries I have installed in conda, prompts for passwords to install some libraries (globally?), and then throws a number of errors, but reports that the build completed successfully. It does not look like any tests were actually run.

ColCarroll:
There's one test failure on travis doesn't seem to be my fault. 

Here are a few other funny bits:

- I reuse a lot of machinery from `simple_step_size_adaptation.py`. Would you prefer that I import them from there; factor them out into `step_size_utils.py` or similar; or leave as is?
- I also reuse a lot of the tests from `simple_step_size_test.py`, and have the same question about not repeating myself.
- The optimization path is not as obvious here: I write the equation for the expected output in a few places, but ended up hard coding the expected output in the broadcasting test (which _did_ help iron out some shape bugs).
- The main loop is pretty nasty in terms of readability:

```python
      for parts in zip(
          step_size_parts, state_parts, error_sum_parts,
          log_averaging_step_parts, shrinkage_target_parts):
        (step_size_part, state_part, error_sum_part,
         log_averaging_step_part, shrinkage_target_part) = parts
```
It feels like a `namedtuple` or similar would be nicer here.

tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
```diff
@@ -0,0 +1,484 @@
...
+
+        # - If still adapting, return an exploring step size,
+        # - If just finished, return the averaging step size
+        # - Otherwise, do not update
+        new_step_size_parts.append(
+            tf.where(
+                previous_kernel_results.step < self.num_adaptation_steps,
+                tf.math.exp(new_log_step),
+                tf.where(
+                    previous_kernel_results.step > self.num_adaptation_steps,
```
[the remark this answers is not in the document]
ColCarroll: Hm -- I think when `step = num_adaptation_steps` we want to switch from the noisy step size to the log-averaged step size, and then after that use the constant step size.

Here is what that ends up looking like:

image

tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
```diff
@@ -0,0 +1,484 @@
...
+                    step_size_part,
+                    tf.math.exp(new_log_averaging_step))))
+        new_log_averaging_step_parts.append(
+            tf.where(
+                previous_kernel_results.step > self.num_adaptation_steps,
```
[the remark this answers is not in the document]
ColCarroll: Let me know if I misunderstood above, since I bet both should have the same answer.

I think keeping this running after fixing the step size will be somewhat misleading, since you do not have the accumulating error informing the step size being used -- just checking acceptance probabilities should give you the same information.

tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
```diff
@@ -0,0 +1,484 @@
...
+        reduced_log_accept_prob = _reduce_logmeanexp(log_accept_prob,
+                                                     tf.range(num_reduce_dims))
+        reduce_indices = _get_differing_dims(reduced_log_accept_prob,
+                                             step_size_part)
+        reduced_log_accept_prob = _reduce_logmeanexp(
+            reduced_log_accept_prob, reduce_indices, keepdims=True)
+        error_sum.append(tf.zeros_like(reduced_log_accept_prob, dtype=dtype))
+        log_averaging_step.append(tf.zeros_like(step_size_part, dtype=dtype))
+        shrinkage_target.append(
+            tf.math.log(tf.constant(10, dtype=dtype) * step_size_part))
```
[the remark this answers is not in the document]
ColCarroll: This is what it was set to in the original paper, and at least pymc3 and stan have used this (non-configurable) constant. I have not done any experiments with the value, but can make it configurable if you'd like! 

I believe this constant is meant to bias the exploration towards a *larger* value (10 times the initial step size). You made a good point that tfp parametrizes by number of steps, so just biasing away from zero might be enough (i.e., to `step_size_part`), but I stuck with implementing the algorithm as it is used.

tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
```diff
@@ -0,0 +1,484 @@
...
+class DualAveragingStepSizeAdaptation(kernel_base.TransitionKernel):
+  """Adapts the inner kernel's `step_size` based on `log_accept_prob`.
+
+  The dual averaging policy uses a noisy step size for exploration, while
+  averaging over tuning steps to provide a smoothed estimate of an optimal
+  value. It is based on [section 3.2 of Hoffman and Gelman (2013)][1], which
+  modifies the [stochastic convex optimization scheme of Nesterov (2009)][2].
+  The modified algorithm applies extra weight to recent iterations while
+  keeping the convergence guarantees of Robbins-Monro, and takes care not
+  to make the step size too small too quickly, to avoid expensive early
```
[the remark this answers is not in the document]
ColCarroll: That's a very nice point, thank you.

ColCarroll:
I have updated this, but:

1. I'm afraid that `tf.nest.map_structure` is throwing 
```pythontb
ValueError: The two structures don't have the same nested structure.
```
on most tests and examples, and I can not seem to find tensorflow 2 documentation to figure out how to (or whether to?) fix this.

2. Running `python -m pytest -v tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation_test.py` works, but `testExample` consistently fails with
```pythontb
absl.flags._exceptions.UnparsedFlagAccessError: Trying to access flag --fixed_seed before flags were parsed.
```

brianwa84:
I'm guessing the map_structure issue is a mismatch between the step size list and the state list.
IIRC @SiegeLordEx has a broadcasting map structure variant that broadcasts the step size over the states; or it shouldn't be too hard to make.

brianwa84:
As to the testing / flags parsing issue, the best bet would be to run tests with bazel from within the virtualenv. Something like `bazel test     --copt=-O3     --copt=-march=native     --notest_keep_going     --test_tag_filters=-gpu,-requires-gpu-sm35,-no-oss-ci     --test_timeout 300,450,1200,3600     --action_env=PATH     --action_env=LD_LIBRARY_PATH     --test_output=errors  relative/path/to:your_test` should work for py3.

brianwa84:
That invocation comes from https://github.com/tensorflow/probability/blob/master/testing/run_tests.sh

SiegeLordEx:
Thanks Colin, things look great. Not sure why `python -m` didn't work out for you, it works locally (perhaps there was some top-level TF/TFP code, which is typically a bad idea for other reasons). I think this is in a good state to send to our internal review system, which when accepted will automatically merge this. Thanks again for your awesome contribution!

I do have to wonder, however, why we're still using dual averaging and not some other subgradient method that came out post 2012 (e.g. Adam? Amsgrad?), I think I'll look into this question at some point.

tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py
```diff
@@ -0,0 +1,507 @@
+# Copyright 2019 The TensorFlow Probability Authors.
```
SiegeLordEx: One more small issue, could you change this to be 2018? Our import process breaks otherwise... sorry for missing this earlier!

csuter: Sorry for the chatter but we should definitely not put the wrong year here. We will come up with a fix internally.

csuter: This should be good to go now (with 2019 as the date)

janosh:
Is this ready to be merged?

brianwa84:
Yes, @csuter will pull it in.

PR #382, opened 2019-04-25
tfd.Binomial does not need custom forward anymore

skeydan:
@junpenglao 

... and thanks for the help today (will probably be back ;-))

IDK if the notebook is in correct state, just erased output but some meta stuff will have sneaked in...


skeydan:
> Check out this pull request on ReviewNB

--> probably this is not directed at me right?

csuter:
> > Check out this pull request on ReviewNB
> 
> --> probably this is not directed at me right?

this is spam; ignore.

junpenglao:
Thanks @skeydan! Could you only make the changes without removing the outputs?

junpenglao:
The meta stuff changes makes it quite difficult to review - maybe the easiest way is to pull from master again and only edit the relevant  part in a text editor?

skeydan:
but I think some cells get evaluated - should I evaluate those then? Or will that happen with the build process?

junpenglao:
In this case I dont think you need to reevaluate those cells, as the custom function should gives the same format of output as the .sample()

skeydan:
wow this was difficult, I hope I did it correctly :-)
(((in fact I _do_ appreciate R markdown even more now ;-)))

junpenglao:
Yeah, plus the size of the file it might make editing pretty difficult - it is still showing a lot of changes in the diff tool, maybe you can try downloading the file from master and open it with a text editor? That way the states etc shouldnt be modified any more :-)

skeydan:
oh but that's what I did? My `git diff` was pretty short, too...

The file from master had lots of binary, so I assume that was correct?

skeydan:
also the diff here

https://github.com/tensorflow/probability/pull/382/files

is short...

tensorflow_probability/examples/jupyter_notebooks/Modeling_with_JointDistribution.ipynb
```diff
@@ -2074,21 +2058,8 @@
...
-        "\n",
-        "  y = tfd.Binomial(\n",
-        "      total_count=tf.cast(at_bats, dtype),\n",
-        "      probs=thetas\n",
-        "  )\n",
-        "  return phi, kappa_log, thetas, y\n",
-        "\n",
-        "phi, kappa_log, thetas, y = custom_forward(4, n)\n",
-        "phi, kappa_log, thetas, y, y.log_prob(hits)"
+       "phi, kappa_log, thetas, y = mdl_baseball.sample(4)\n",
```
junpenglao:
```suggestion
       "source": [
       "phi, kappa_log, thetas, y = mdl_baseball.sample(4)\n",
```

junpenglao:
You are right! Left a comment.

skeydan:
oops, thanks!

PR #388, opened 2019-04-30
Add deprecation notice for GPU builds

brianwa84:


csuter:
are you planning to build a new gpu 0.4 release from this?

maybe we can just delete the projects from pypi?

brianwa84:
Was going to build a new pip pkg. We could just delete, I guess. Prefer that?

brianwa84:
We will just delete the old pips.

PR #392, opened 2019-05-01
Fix Triangular bounds (fixes #391)

### LINKED ISSUE #391: Triangular distribution has density of 1 at low point
Repro below. There's a definite mismatch with both written spec and scipy behavior.

```
import tensorflow as tf

with tf.Session() as sess:
    ph = tf.placeholder(dtype=tf.float32)
    print(sess.run(tfp.distributions.Triangular(low=2.0, peak=4.0, high=6.0).prob(ph), {ph: 2}))
```

Output:

```
1.0
```

Expected output:
```
0.0
```

cchan:
Previously produced a nonzero density at the low point, since the left_of_peak condition did not include the low point.

### DIFF
--- a/tensorflow_probability/python/distributions/triangular.py
+++ b/tensorflow_probability/python/distributions/triangular.py
@@ -234,7 +234,7 @@ class Triangular(distribution.Distribution):
 
     broadcast_x_to_high = _broadcast_to(x, [self.high])
     left_of_peak = tf.logical_and(
-        broadcast_x_to_high > self.low, broadcast_x_to_high <= self.peak)
+        broadcast_x_to_high >= self.low, broadcast_x_to_high <= self.peak)
 
     interval_length = self.high - self.low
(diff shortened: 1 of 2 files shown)


cchan:
> Could you add a test for this change?

I've written something locally, but can't figure out how to run the test. How do I do this? Do I have to install Bazel?

I tried `python tensorflow_probability/python/distributions/triangular_test.py` but that dies with an `ImportError: cannot import name 'test_case' from 'tensorflow_probability.python.internal'`.

brianwa84:
You would have to either install bazel or wait on Travis to run tests for
you.

brianwa84:
If you can just add the test to the PR I can make sure it passes as I pull it in.

cchan:
Seems to work now! Let me know if these changes look correct to you.

brianwa84:
Will look it in tomorrow

PR #393, opened 2019-05-01
Fix numpy_test on Travis.

SiegeLordEx:
This fixes it by switching the python interpreter version on Travis to
be the one provided by the Ubuntu repositories. Concretely, that version
of the interpreter has the 'random' module built in, so it doesn't get
overridden by a local copy of it in the numpy backend directory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,9 @@ jobs:
 # of python versions x SHARD will be expanded and run in parallel.
 python:
   - "2.7"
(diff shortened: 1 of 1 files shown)


SiegeLordEx:
This change makes it use whatever is available in Ubuntu's apt repositories, and this version of Ubuntu only has 3.4. That said, I could make it use a newer Ubuntu version which has 3.5. 3.6 is right out.

PR #424, opened 2019-05-27
check for use_exact_kl == True instead of None

skeydan:
I'm not completely sure why the existing code checks for  `use_exact_kl is None`, but wouldn't the user expect that if she passes `True` here she gets the exact KL?
Perhaps I'm missing sth but I don't see what...


skeydan:
you're welcome!

PR #444, opened 2019-06-03
'Added SkewGeneralizedNormal distribution'

inhandan:
Added a distribution class for the generalized Gaussian distribution - type 2, as defined on Wikipedia
https://en.wikipedia.org/wiki/Generalized_normal_distribution
This is distribution features the traditional Gaussian as a special case, where a third parameter, which I call shape, is set to 0. The first two parameters then reduce to mean and variance.

I've had use with this distribution in that it allows for modelling a variable with high probability density near it's upper bound


inhandan:
I signed it

inhandan:
I signed it!

PR #468, opened 2019-06-18
pass props from TransformedDistribution to RelaxedOneHotCategorical

### LINKED ISSUE #441: RelaxedOneHotCategorical has no probs attribute
While both `OneHotCategorical` and `ExpRelaxedOneHotCategorical` have a `probs` attribute, `tfp.distributions.RelaxedOneHotCategorical` does not have a `probs` attribute.

I am using tfp-nightly and tf-nightly.

trainorpj:
Fixes #441 

If you have a `RelaxedOneHotCategorical` distribution:

```py
dist = tfd.RelaxedOneHotCategorical(temperature, probs=probs)
```

Now you can access `probs` and `logits` directly, instead of going through the `distribution` attribute, since `RelaxedOneHotCategorical` is a `TransformedDistribution`.

```py
# before
dist.distribution.probs

# after this PR
dist.probs
```

## Notes

- 🤔Should other attributes be passed, e.g. `event_size`?
- 🤕I couldn't figure out how to set up the test env


SiegeLordEx:
Travis will run the tests for you, for now it's complaining about lints: https://travis-ci.org/tensorflow/probability/jobs/547424198 (the function redifinition lints are there because your newly added functions are indented too much).

trainorpj:
Some Python 2.7 Builds are failing tests with `ImportError: No module named 'numpy'
`

Not sure if this is an old problem or if it has something to do with a change I made, although that seems unlikely 🤷‍♀

SiegeLordEx:
Yeah, Python 2 is broken for unrelated reasons.

Would you mind also adding an accessor like that for `temperature` as well, it's odd to leave that out.

Otherwise, it seems good to go, thanks!

tensorflow_probability/python/distributions/relaxed_onehot_categorical.py
```diff
@@ -407,15 +407,26 @@ def __init__(
...
+
+  
+
     dist = ExpRelaxedOneHotCategorical(temperature,
                                        logits=logits,
                                        probs=probs,
                                        validate_args=validate_args,
                                        allow_nan_stats=allow_nan_stats)
+
+    self._temperature = dist.temperature
```
SiegeLordEx: Why do you need to this? Doesn't `self.distribution.temperature` not work?

trainorpj: You're right, I got a bit confused. Changing now.

trainorpj:
I also added an accessor for `event_size`, since it's a property for `ExpRelaxedOneHotCategorical `, and it's mentioned in the docstring for `RelaxedOneHotCategorical`

SiegeLordEx:
Thanks, I'll send it in to our internal review system and we'll get this merged.

Issue #461, opened 2019-06-12
Feature request: Optimizer printing number of iterations every n iterations 


timudk:
It would be great if optimizer, e.g., tfp.optimizer.bfgs_minimize(), would have an argument like print_number_iterations so that you know in runtime how much longer the optimization procedure might take.

I am willing to contribute.

Best, 

Tim

saxena-ashish-g:
This can be added relatively easily. I would also be happy to review a PR for this.

timudk:
@saxena-ashish-g I will work on it and send a PR.

brianwa84:
@midfield FYI

PR #475, opened 2019-06-23
Fixing issue 461

timudk:
This problem fixes [issue 461](https://github.com/tensorflow/probability/issues/461). It might not be the best solution. 

Do you have any other ideas @saxena-ashish-g?


tensorflow_probability/python/optimizer/bfgs.py
```diff
@@ -78,7 +78,8 @@ def minimize(value_and_gradients_function,
              max_iterations=50,
              parallel_iterations=1,
              stopping_condition=None,
-             name=None):
+             name=None,
+             print_n_iterations=50):
```
perezju: maybe this should default to None (no printing), and only add the print_op if not None?

PR #480, opened 2019-06-26
Fix typo in real_nvp.py

emiliojorge:
Fixing a small typo in spelling of memorizes in real_nvp.py


emiliojorge:
`I signed it!`

SiegeLordEx:
That's actually correct as is, the comment means this https://en.wikipedia.org/wiki/Memoization. Thanks for reaching out!

brianwa84:
Arguably memoize is a bad word here, though, because it's not [exactly] like we are sometimes doing work, sometimes using cached work.
Perhaps "specializes the bijector for `D`" or "initializes weights matching size `D`".

SiegeLordEx:
The related internal function is literally called 'cache', so 'caches' is possibly the right verb. @emiliojorge , if you're interested, you could change it to that and we'll merge this change in :).

PR #507, opened 2019-07-25
fix formatting typos in Autoregression docs

trainorpj:
A space before triple tick marks ` ``` ` was causing some bad rendering.

See the [docs for `AutoregressiveStateSpaceModel`](https://www.tensorflow.org/probability/api_docs/python/tfp/sts/AutoregressiveStateSpaceModel) and [`Autoregressive`](https://www.tensorflow.org/probability/api_docs/python/tfp/sts/Autoregressive). Alternatively, see the images below, and note the triple tick mark.

`AutoregressiveStateSpaceModel`

typo in AutoregressiveStateSpaceModel docs causes bad rendering

`Autoregressive`

typo in Autoregressive docs causes bad rendering


davmre:
Thanks @trainorpj for catching this and making a PR! We really appreciate the feedback when mistakes like this slip through the cracks.

Issue #209, opened 2018-11-02
Feature request: PERT distribution


srhoades10:
I'd like to test out a beta-PERT distribution for my particular problem. However, I would also look for a way to use a Beta distribution and omit (a, b) < 1 to  avoid an arcsine-type of distribution (I'm modeling reaction activities, so I essentially want something more flexible than a Normal but on bounded support). As the calculations are very similar to a Beta, I wrote up a PERT distribution script, however I'm struggling to integrate it into my tf import (I get errors regarding input dtypes and tf.math.xlogy errors).

I'm happy to make a request for what I have made thus far, although I believe the the calculations for entropy and KL_div are only identical in the event of a lower bound of 0. Thanks!

axch:
It would probably be easier to get PERT as a scale+shift transformation of a standard Beta than coding the computations directly.  Something like this:
```
alpha = 6 * (mean - min) / (max - min)
beta = 6 * (max - mean) / (max - min)
std_beta = tfp.distributions.Beta(concentration1=alpha, concentration0=beta)
pert = tfp.distributions.TransformedDistribution(
  distribution=std_beta,
  bijector=tfp.bijectors.AffineScalar(shift=min, scale=(max - min)))
```

If you would like to contribute a PERT implementation along those lines (which we would definitely appreciate!), you can take a look at the multivariate Gaussian as an example: https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/distributions/mvn_linear_operator.py#L169-L193.  That's coded as a class that inherits from TransformedDistribution; it gets a nontrivial covariance by transforming an isotropic Gaussian with a linear operator bijector.

srhoades10:
Thanks for the suggestion, I've since gone through the tfp's paper detailing the utility of bijectors, and I think this is a much simpler and concise approach. If I tidy up the PERT implementation then I can move forward to contribute, but I'll see if this Beta transformation suits my objectives too

srvasude:
Closing this as PERT was added as a distribution.

PR #564, opened 2019-09-17
PERT distribution and basic test for issue #209

colemakdvorak:
Hello,

Issue #209 looked like a good first issue to learn about some building blocks of this project so I gave it a go. I held off on KL divergence implementation for now since I couldn't see a clean analytical solution right off the bat.

This PR implements the [modified PERT](https://en.wikipedia.org/wiki/PERT_distribution), which is a general case of standard PERT that includes an extra parameter to control the shape of the distribution. 

I used TensorFlow pylint for general style guide while referencing other implementations within this repo at my own discretion. So it might be a little off here and there - please feel free to point fingers. Whether it's about style, documentation, or implementation, I am open to any suggestions.

### DIFF
--- a/tensorflow_probability/python/distributions/pert.py
+++ b/tensorflow_probability/python/distributions/pert.py
@@ -21,7 +21,7 @@ class PERT(transformed_distribution.TransformedDistribution):
   """Modified PERT distribution for modeling expert predictions.
 
   PERT distribution is a loc-scale family of Beta distribution
-  fit onto an arbitrary real interval set between LOW and HIGH
+  fit onto an arbitrary real interval between LOW and HIGH
   values set by the user, along with PEAK to indicate the expert's
   most frequent prediction.
   [1](https://en.wikipedia.org/wiki/PERT_distribution).
--- a/tensorflow_probability/python/distributions/pert.py
+++ b/tensorflow_probability/python/distributions/pert.py
@@ -43,6 +43,8 @@ class PERT(transformed_distribution.TransformedDistribution):
   ```
 
   PERT distribution is obtained when g = 4.
+
+  For general reference: https://en.wikipedia.org/wiki/PERT_distribution
   """
   def __init__(self,
                mini,
--- a/tensorflow_probability/python/distributions/pert.py
+++ b/tensorflow_probability/python/distributions/pert.py
@@ -18,33 +18,64 @@ __all__ = [
 ]
 
 class PERT(transformed_distribution.TransformedDistribution):
-  """Modified PERT distribution:
-  PERT distribution is a loc-scale family of Beta distribution intended to
-  model expert prediction on arbitrary real interval bounded by MAX and MIN
+  """Modified PERT distribution for modeling expert predictions.
+
+  PERT distribution is a loc-scale family of Beta distribution
+  fit onto an arbitrary real interval set between MAX and MIN
   values set by the user, along with MODE to indicate the expert's
   most frequent prediction.
+  [1](https://en.wikipedia.org/wiki/PERT_distribution)
 
-  The shift and scale factor is given by:
+  #### Mathematical Details
 
+  In terms of a Beta distribution, PERT can be expressed as
+
+  ```none
+  PERT ~ loc + scale * Beta(concentration1, concentration0)
   ```
-  shift = MIN
+  where
+
+  ```none
+  loc = MIN
   scale = MAX - MIN
+                            MODE - MIN
+  concentration1 = 1 + g * ------------
+                            MAX - MIN
+
+                            MAX - MODE
+  concentration1 = 1 + g * ------------
+                            MAX - MIN
+  g > 0
   ```
+  over support [MIN, MAX]. And some MODE such that
+  MIN < MODE < MAX. `g` is a positive parameter that controls
+  the shape of the the distribution. Higher value forms a sharper
+  peak.
 
-  Modified PERT is a generalized version of PERT that has an additional
-  parameter, smoothness (g).
+  Standard PERT distribution is obtained when g = 4.
 
-  Smoothness determines the properties of Beta distribution:
+  #### Examples
+  ```python
+  import tensorflow_probability as tfp
+  tfd = tfp.distributions
 
+  # Single PERT distribution
+  dist = tfd.PERT(mini=1., mode=7., maxi=11., smoothness=4.)
+  dist.sample(10)
+  dist.prob(7.)
+  dist.prob(0.) # Throws nan when input out of support.
+
+  # Multiple PERT distributions with varying smoothness (broadcasted)
+  dist = tfd.PERT(mini=1., mode=7., maxi=11., smoothness=[1., 2., 3., 4.])
+  dist.sample(10)
+  dist.prob(7.)
+  dist.prob([[7.],[5.]])
+
+  # Multiple PERT distributions with varying mode
+  dist = tfd.PERT(mini=1., mode=[2., 5., 8.], maxi=11., smoothness=4.)
+  dist.sample(10)
+  dist.sample([10,5])
   ```
-  concentration1 = 1 + g * (MODE - MIN)/(MAX - MIN)
-  concentration0 = 1 + g * (MAX - MODE)/(MAX - MIN)
-  mean = (MIN + g * MODE + MAX)/(g + 2)
-  ```
-
-  PERT distribution is obtained when g = 4.
-
-  For general reference: https://en.wikipedia.org/wiki/PERT_distribution
   """
   def __init__(self,
                mini,
@@ -59,6 +90,7 @@ class PERT(transformed_distribution.TransformedDistribution):
       dtype = dtype_util.common_dtype([mini, maxi, mode], tf.float32)
       self._min = tensor_util.convert_nonref_to_tensor(
           mini, name='mini', dtype=dtype)
+      self._mode
       self._mod = tensor_util.convert_nonref_to_tensor(
           mode, name='mode', dtype=dtype)
       self._max = tensor_util.convert_nonref_to_tensor(
--- a/tensorflow_probability/python/distributions/pert.py
+++ b/tensorflow_probability/python/distributions/pert.py
@@ -1,0 +1,162 @@
+"""The PERT distribution class"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow.compat.v2 as tf
+from tensorflow_probability.python.distributions import Beta
+from tensorflow_probability.python.bijectors import AffineScalar
+from tensorflow_probability.python.distributions import transformed_distribution
+from tensorflow_probability.python.internal import assert_util
+from tensorflow_probability.python.internal import dtype_util
+from tensorflow_probability.python.internal import prefer_static
+from tensorflow_probability.python.internal import tensor_util # pylint: disable=g-direct-tensorflow-import
+
+__all__ = [
+    'PERT',
+]
+
+class PERT(transformed_distribution.TransformedDistribution):
+  """Modified PERT distribution:
+  PERT distribution is a loc-scale family of Beta distribution intended to
+  model expert prediction on arbitrary real interval bounded by MAX and MIN
+  values set by the user, along with MODE to indicate the expert's
+  most frequent prediction.
+
+  The shift and scale factor is given by:
+
+  ```
+  shift = MIN
+  scale = MAX - MIN
+  ```
+
+  Modified PERT is a generalized version of PERT that has an additional
+  parameter, smoothness (g).
+
+  Smoothness determines the properties of Beta distribution:
+
+  ```
+  concentration1 = 1 + g * (MODE - MIN)/(MAX - MIN)
+  concentration0 = 1 + g * (MAX - MODE)/(MAX - MIN)
+  mean = (MIN + g * MODE + MAX)/(g + 2)
+  ```
+
+  PERT distribution is obtained when g = 4.
+  """
+  def __init__(self,
+               mini,
+               mode,
+               maxi,
+               smoothness=4.,
+               validate_args=False,
+               allow_nan_stats=False,
+               name="PERT"):
+    parameters = dict(locals())
+    with tf.name_scope(name) as name:
+      dtype = dtype_util.common_dtype([mini, maxi, mode], tf.float32)
+      self._min = tensor_util.convert_nonref_to_tensor(
+          mini, name='mini', dtype=dtype)
+      self._mod = tensor_util.convert_nonref_to_tensor(
+          mode, name='mode', dtype=dtype)
+      self._max = tensor_util.convert_nonref_to_tensor(
+          maxi, name='maxi', dtype=dtype)
+      self._smoothness = tensor_util.convert_nonref_to_tensor(
+          smoothness, name='smoothness', dtype=dtype)
+      self._concentration1 = 1. + self._smoothness \
+        * (self._mod - self._min) / (self._max - self._min)
+      self._concentration0 = 1. + self._smoothness \
+        * (self._max - self._mod) / (self._max - self._min)
+
+      self._scale = self._max - self._min
+
+      # Broadcasting scale on affine bijector to match batch dimension.
+      # This prevents dimension mismatch for operations like cdf.
+      super(PERT, self).__init__(
+          distribution=Beta(
+              concentration1=self._concentration1,
+              concentration0=self._concentration0,
+              allow_nan_stats=allow_nan_stats),
+          bijector=AffineScalar(
+              shift=self._min,
+              scale=tf.broadcast_to(
+                  input=self._scale,
+                  shape=self._batch_shape_tensor(
+                      concentration1=self._concentration1, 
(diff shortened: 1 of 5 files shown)


tensorflow_probability/python/distributions/pert.py
```diff
@@ -0,0 +1,164 @@
...
+               smoothness=4.,
+               validate_args=False,
+               allow_nan_stats=False,
+               name="PERT"):
+    parameters = dict(locals())
+    with tf.name_scope(name) as name:
+      dtype = dtype_util.common_dtype([mini, maxi, mode], tf.float32)
+      self._min = tensor_util.convert_nonref_to_tensor(
+          mini, name='mini', dtype=dtype)
+      self._mod = tensor_util.convert_nonref_to_tensor(
```
brianwa84: should this be `_mode`?

tensorflow_probability/python/distributions/pert.py
```diff
@@ -0,0 +1,164 @@
...
+from tensorflow_probability.python.internal import dtype_util
+from tensorflow_probability.python.internal import prefer_static
+from tensorflow_probability.python.internal import tensor_util # pylint: disable=g-direct-tensorflow-import
+
+__all__ = [
+    'PERT',
+]
+
+class PERT(transformed_distribution.TransformedDistribution):
+  """Modified PERT distribution:
```
brianwa84: the first line should be a one-liner "sentence" e.g. "Modified PERT distribution for modeling expert predictions."
then an empty line.

can you add a "Mathematical details" section a la https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/distributions/gamma_gamma.py#L51
and optionally an example or two a la
https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/distributions/dirichlet.py#L103 ?

tensorflow_probability/python/distributions/pert.py
```diff
@@ -0,0 +1,164 @@
...
+  concentration0 = 1 + g * (MAX - MODE)/(MAX - MIN)
+  mean = (MIN + g * MODE + MAX)/(g + 2)
+  ```
+
+  PERT distribution is obtained when g = 4.
+
+  For general reference: https://en.wikipedia.org/wiki/PERT_distribution
+  """
+  def __init__(self,
+               mini,
```
brianwa84: let's rename these to match the triangular distribution (low, high, peak [though mode is kind of nice too])
https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/distributions/triangular.py#L95
you might even mention PERT being "like Triangular, but smoother" or something along those lines.

tensorflow_probability/python/distributions/pert.py
```diff
@@ -0,0 +1,164 @@
...
+      dtype = dtype_util.common_dtype([mini, maxi, mode], tf.float32)
+      self._min = tensor_util.convert_nonref_to_tensor(
+          mini, name='mini', dtype=dtype)
+      self._mod = tensor_util.convert_nonref_to_tensor(
+          mode, name='mode', dtype=dtype)
+      self._max = tensor_util.convert_nonref_to_tensor(
+          maxi, name='maxi', dtype=dtype)
+      self._smoothness = tensor_util.convert_nonref_to_tensor(
+          smoothness, name='smoothness', dtype=dtype)
+      self._concentration1 = 1. + self._smoothness \
```
brianwa84: for long lines, tfp style is to put an open close paren around the whole thing and omit the `\`

tensorflow_probability/python/distributions/pert.py
```diff
@@ -0,0 +1,164 @@
...
+              shift=self._min,
+              scale=tf.broadcast_to(
+                  input=self._scale,
+                  shape=self._batch_shape_tensor(
+                      concentration1=self._concentration1,
+                      concentration0=self._concentration0))),
+          validate_args=validate_args,
+          parameters=parameters,
+          name=name)
+
```
brianwa84: can you also implement _batch_shape (which is preferred when shapes are statically known)

tensorflow_probability/python/distributions/pert.py
```diff
@@ -0,0 +1,164 @@
...
+        ))
+    if is_init != tensor_util.is_ref(self._smoothness):
+      assertions.append(assert_util.assert_positive(
+          self._smoothness,
+          message="Smoothness parameter must be positive."
+      ))
+    return assertions
+
+  def _survivial_function(self, value, **kwargs):
+    return 1. - self.cdf(value, kwargs)
```
brianwa84: this should already be implemented as such in the base class (i.e. you can leave it out)

tensorflow_probability/python/distributions/pert.py
```diff
@@ -0,0 +1,164 @@
...
+          name=name)
+
+  def _batch_shape_tensor(self, concentration1=None, concentration0=None):
+    return prefer_static.broadcast_shape(
+        prefer_static.shape(
+            self.concentration1 if concentration1 is None else concentration1),
+        prefer_static.shape(
+            self.concentration0 if concentration0 is None else concentration0))
+
+  def _parameter_control_dependencies(self, is_init):
```
brianwa84: we prefer sticking this method at the very bottom, out of the way :)

tensorflow_probability/python/distributions/pert.py
```diff
@@ -0,0 +1,164 @@
...
+    if not self.validate_args:
+      return []
+    assertions = []
+    if is_init != tensor_util.is_ref(self._mod):
+      if is_init != tensor_util.is_ref(self._min):
+        assertions.append(assert_util.assert_greater(
+            self._mod,
+            self._min,
+            message="Mode must be greater than minimum."
```
brianwa84: minor nit: can we put these param names in backticks like "`peak` must be greater than `low`"

tensorflow_probability/python/distributions/pert_test.py
```diff
@@ -0,0 +1,205 @@
...
+    self.assertEqual(tf.TensorShape([8]), dist.batch_shape)
+    self.assertAllEqual([], self.evaluate(dist.event_shape_tensor()))
+    self.assertEqual(tf.TensorShape([]), dist.event_shape)
+
+  def testBroadcastingHigherDimParam(self):
+    dist = tfd.PERT(
+        mini=[1., 2.],
+        mode=[[2., 3.], [4., 5.]],
+        maxi=10.,
+        smoothness=4.,
```
brianwa84: I'm guessing we might want to rename this to `temperature`, but if you want you can leave it as is for now, I'll discuss w/ the team when we get ready to merge it.

tensorflow_probability/python/distributions/pert_test.py
```diff
@@ -0,0 +1,205 @@
...
+    maxi = tf.Variable([5., 5., 4.])
+    self.evaluate(maxi.initializer)
+    self.evaluate(mode.initializer)
+    with self.assertRaisesRegex(
+        tf.errors.InvalidArgumentError,
+        "Maximum must be greater than mode."):
+      dist = tfd.PERT(mini, mode, maxi, validate_args=True)
+      self.evaluate(dist.sample(1))
+
+  def testMaxModeInequalityAfterMutation(self):
```
brianwa84: this test looks great!

if you want a little more coverage, you can add a 
```python
@classmethod
def _params_event_ndims(cls):
  return dict(low=0, high=0, peak=0, temperature=0)
```
to the distribution (which will also make it batch-sliceable, i.e. `tfd.PERT(...)[..., 3:, -2]`), and then whitelist it in distribution_properties_test.py

tensorflow_probability/python/distributions/pert.py
colemakdvorak: I guess this won't matter anymore since I agree on changing these names to `low`, `peak`, and `high` but I am just curious about how to approach this kind of case in terms of general software engineering / design principle.

So I avoided calling this `_mode` because it seemed to be an abstract method in `distribution` class - and I didn't want to set a property on a name where a class usually expects a method. This is how I ended up setting it to `_mod` and implemented `_mode` method to return this property. Is this approach reasonable or is there a general good/alternative practice when it comes to these kind of things?

colemakdvorak:
Oops. In the process of fixing the suggestions and keeping my forked repo up to date I ended up including commits that don't include CLA.

tensorflow_probability/python/distributions/pert.py
colemakdvorak: Done.

tensorflow_probability/python/distributions/pert_test.py
colemakdvorak: To be honest either works for me intuitively, but I've seen the term `temperature` several times in context of statistics so I changed it.

tensorflow_probability/python/distributions/pert.py
colemakdvorak: Added a simple mathematical overview and couple of examples demonstrating broadcasting behaviors.

https://github.com/tensorflow/probability/blob/33ea65ee4be4ed05247357f7d5b374a38d3b3f58/tensorflow_probability/python/distributions/pert.py#L21

tensorflow_probability/python/distributions/pert_test.py
colemakdvorak: Cool, I haven't fully wrapped my head around details of `distribution_properties_test.py` but I get the gist of it.

https://github.com/tensorflow/probability/blob/ba0701f00cd1bed083caa61f392f4acd91f0fe20/tensorflow_probability/python/distributions/distribution_properties_test.py#L1048

I saw the test inputting some wild parameters, so I fixed it by adding something similar to `fix_triangular` with additional handling of `temperature`.

colemakdvorak:
Thank you for such an extensive feedback @brianwa84 . I went through all of your feedbacks and addressed the points you brought up.

Not sure what to do about the CLA part though, I assume it's got to do with me including a Googler's commit. Can I expect this to go away or do I juggle with git?

tensorflow_probability/python/distributions/pert_test.py
brianwa84: Great, I'll double check it when I bring the code in. I wonder if rebasing the PR instead of adding the commits into this PR would fix the CLA issue. It seems to have a problem with the "no-reply@google.com" commit (I think, the PlackettLuce one).

csuter:
are you merging from master or rebasing on your fork? using rebase would definitely leave a cleaner trail and avoid this issue. i generally keep a local branch called "upstream" which tracks the remote master, pull to that, then rebase my local master on top of it. i can help sort things out from this state if you need a hand.

csuter:
i'm working on creating a copy of your branch now without the merges; will update shortly. once it's in good shape you can just (force) pull it onto your master and the PR will automatically clean itself up.

csuter:
Ok, so I am a little confused about what happened before 20535835: "merge conflict resolution" (it looks like some commits contained unresolved merge conflict markers). So I've dropped everything before that since it looks like you tried to start over from scratch.

There's also been some annoyingness because PERT is being checked in simultaneously with PlackettLuce, and git is having a really hard time with all the edits on adjacent lines. Furthermore, PlackettLuce was submitted, then just now rolled back, and will shortly be resubmitted again, so we're probably not out of the woods with those merge conflicts 😭

Anyway, here is a branch on my fork that has everything you've done so far (aside from the stuff before 20535835: "merge conflict resolution", as noted above):

https://github.com/csuter/probability/tree/fix_colemakdvorak

Here's what you can do to clean up this PR:

```
# 1. First things first, create a branch that knows where your master is now,
# so you can get back to it later if need be.
git checkout master
git checkout -b master_backup

# 2. Add my fork as a remote, and fetch its contents
git remote add csuter https://github.com/csuter/probability
git fetch csuter

# 3. Hard reset your master to point to where my fixer upper branch is
git reset --hard csuter/fix_colemakdvorak

# 4. Update your remote on GitHub, which should result in this PR containing just your commits.
git push origin master
```

After that's all done your changes will be "rebased" onto the current HEAD of the TFP repo, the CLA  issues should disappear. Once Aditya rolls back the rollback of PluckettLuce, you'll probably hit some more merge issues. Maybe we can get your PR in before then. If not, I strongly advise using rebase locally to get things in order again. Assuming you have a local branch called `upstream` that tracks the TFP repo, this would look something like

```
# Switch to upstream branch and get the latest goodies
git checkout upstream
git pull

# Jump back to your master branch and rebase
git checkout master
git rebase upstream
```

If you hit conflicts there, git should give some nice instructions for resolving. Basically, edit the conflicting files (they'll be shown in red if you do `git status`), resolve the merge conflicts (marked by `<<<<<<<<`, `>>>>>>>`, `=======`), `git add` the fixed up files, then run `git rebase --continue` to proceed. Rebase is stepping through each of your commits to try to "replay" them on top of the upstream, so you may run into multiple of these, as say one of your commits collided with PlackettLuce in BUILD and another in __init__.py (or whatever).

Finally, this is a super useful command for visualizing the commit history in the form of a graph:

```
git -c color.diff=always log --graph --date-order --date=short --branches --pretty='%C(green bold)%h%Creset %C(blue bold)%d%Creset %C(black bold)%s%Creset
```

I have this aliased in my .gitconfig as `git gr`. When I'm doing git surgery, i basically alwasy keep an extra terminal open off to the side with the output of this command, and refresh it each time I make a change. I probably could not use git without this command :)

Please let me know if you have any questions -- sorry for the insanely long post!

colemakdvorak:
Sorry about the hassle here @csuter . I will start looking into your instructions. I've been fetching then rebasing from upstream/master to my local master, and at some point there has been a pull/merge conflict between my forked repo and local repo. Which I dealt with individually. And what I did is what you got :cry:.

csuter:
No hassle at all. For some reason I get a great deal of pleasure out of fixing git issues 🤷‍♂️

When you say "rebasing from upstream/master to local master" do you mean that the other way around? I think the jargon would have it "rebasing my local master onto the upstream master" which means "replaying some local commits, which were written on top of an old upstream state, as if they had been written on top of a newer upstream state". If that's what you meant, great! (Although then I'm not sure how the issues arose in the PR). If you meant something else, then that may be a source of the issues that arose here. If you run that graph command I shared on your repo in its current state, you'll see there are some branches that diverged and then rejoined. That implies there must've been some git merge commands run at some point (maybe in the GH UI?). Git merges are basically the root of all (VCS) evil IMO 😊. I wish GH didn't make them so easy to do with the click of a button....

colemakdvorak:
Well god bless, thanks for the git surgery instructions - the CLA sign is green now. And yes, I think I got terminology direction backwards. I will take note of the tricks and tips, and try to brush up on git. Thank you, it really means a lot to me.

csuter:
Woohoo! Glad we got things sorted out.

Looks like PlackettLuce is back in now, so there's a good chance to practice those git tricks, getting your master rebased onto upstream again :)

Feel free to reach back out with any questions or issues you run into!

colemakdvorak:
I checked few changes to commits done and I've got to say I learned more about this project than I contributed :) 

I know that the style guide already refers to calling `convert_to_tensor` on every argument after a name scope, and few implemented methods like `_mode` and `_variance` are called through a wrapper function that defines the scope and I missed that.

Quick question before closing:

If I am understanding `_transformed_beta` correctly, I assume invoking the method doesn't instantiate a transformed beta distribution every time it's called since local variables in that method are tied to specific properties defined at construction. Then what is the reasoning behind converting some arguments in `_variance` and passing it in as an argument again? What does "avoiding multiple conversions" in comment mean?

brianwa84:
> learned more about this project than I contributed
I don't think so! Without your work here, there would be no `tfd.PERT`. I learned some things too, e.g. about the statistical testing thing in the empirical cdf test during internal review.

Most of the changes I made while pulling it in were to make it more compliant with a "GradientTape-safety" effort we're working through. The specific goal would be to be able to do something like:

```python
d = tfd.PERT(tf.Variable(0.), tf.Variable(1.), tf.Variable(.7))
d.low.assign(.1)  # don't miss this assignment!
with tf.GradientTape() as tape:
  nll = -d.log_prob(.5)
grads = tape.gradient(nll, d.trainable_variables)  # not None, please!
```

The problem is that if we create concentration0/1 in the constructor, we will lose the linkage to those Variables: if those variables change the distribution will not reflect the changes; if we ask for gradients to the variables, they will be `None`. So we do indeed reconstruct the underlying transformed beta for each relevant method call, to ensure that variables are respected.

While we are trying to respect these variables, we are also trying to make sure we don't create too many reads (which require synchronization, and could be racy with variables assignments). This is what "multiple conversions" is referring to (`tf.convert_to_tensor(a_variable)`). I also made more rigorous the `distribution_properties_test`, to ensure we catch a couple more cases that were slipping through e.g. `fix_triangular` and `fix_pert`.

Hopefully that gives you a bit of background on the diff from your PR to head. Mostly it was moving your code around, removing sqrt(variance) because it's implemented in tfd.Distribution base class, little things that depend on some background knowledge.

Thanks for your PR, and lmk if you  have more questions; I'm happy to answer. Please keep sending more PRs!! Distributions & bijectors are quite valuable to increasing the breadth of the library, and we're happy to keep enriching diagnostics like tfp.stats, tfp.math, tfp.math.psd_kernels, etc.

PR #574, opened 2019-09-28
fix doc rendering for LocalLevelStateSpaceModel

trainorpj:
An additional whitespace was making it so that the code blocks in the docs weren't rendering properly. See [here](https://www.tensorflow.org/probability/api_docs/python/tfp/sts/LocalLevelStateSpaceModel)

Same issue as #507, which was merged into master, although the rendering is still bad on those pages. See [here](https://www.tensorflow.org/probability/api_docs/python/tfp/sts/Autoregressive)

### DIFF
--- a/tensorflow_probability/python/sts/local_level.py
+++ b/tensorflow_probability/python/sts/local_level.py
@@ -48,7 +48,7 @@ class LocalLevelStateSpaceModel(tfd.LinearGaussianStateSpaceModel):
 
   ```python
   level[t] = level[t-1] + Normal(0., level_scale)
-   ```
+  ```
 
   The latent state is `[level]` and `[level]` is observed (with noise) at each
   timestep.


axch:
Hey, thanks for the fix, and sorry for the slow response.  I don't think I can pull this in now as-is due to some internal tooling stuff; if you'd like the contribution to register, could you please rebase the branch onto a recent upstream master and update the PR?  Then we'll pull it in on our end.

If it's too much bother, I can also just fix the doc and close this, but then Github will not think you contributed that improvement.

Thanks!

trainorpj:
I'm having some git troubles which I don't have time to troubleshoot 😦it's okay with me if I don't get the contribution. Thanks for offering, though 😄

brianwa84:
This was merged in https://github.com/tensorflow/probability/commit/143806ecd704689ff34e0c85ccedd9064394ba68

PR #560, opened 2019-09-14
Update README.md

luiarthur:
Fixed broken link for `Upgrading_from_Edward_to_Edward2.md`

PR #577, opened 2019-09-30
R0.8 cherry-picks

brianwa84:


brianwa84:
CLAs are good per https://github.com/tensorflow/probability/pull/562 and https://github.com/tensorflow/probability/pull/560

PR #595, opened 2019-10-09
Update fitting.py for moved SeedStream

tarrou:
Changed `tfp_util.seedstream` to `tfp.util.SeedStream` based on deprecation error:

WARNING:tensorflow:From `...\tensorflow_probability\python\sts\fitting.py:185: SeedStream.__init__` (from `tensorflow_probability.python.util.seed_stream`) is deprecated and will be removed after 2019-10-01.
Instructions for updating:
SeedStream has moved to `tfp.util.SeedStream`.

### DIFF
--- a/tensorflow_probability/python/sts/fitting.py
+++ b/tensorflow_probability/python/sts/fitting.py
@@ -185,7 +185,7 @@ def build_factored_surrogate_posterior(
 
   """
   with tf.name_scope(name or 'build_factored_surrogate_posterior'):
-    seed = tfp_util.SeedStream(
+    seed = tfp.util.SeedStream(
         seed, salt='StructuralTimeSeries_build_factored_surrogate_posterior')
     variational_posterior = collections.OrderedDict()
     for param in model.parameters:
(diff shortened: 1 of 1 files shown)


tarrou:
@googlebot I fixed it.

tarrou:
@googlebot I fixed it.

tarrou:
Well, I can't figure out what the Travis CI build is failing. If anyone knows, please post a note of explanation.

Otherwise, I guess this should be closed.

brianwa84:
The deprecation warning is strange. https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/distributions/__init__.py#L152 is where it comes from. Yet as far as I can tell, we do not use this within TFP.

brianwa84:
Ah, it's actually used in tensorflow_probability/python/sts/structural_time_series.py

brianwa84:
Similar for local_level.py

axch:
Well, locally your problem is that you need to `import tensorflow_probabiliy as tfp` in fitting.py in order for the name `tfp` to be defined: https://travis-ci.org/tensorflow/probability/jobs/595681699#L206-L209.  The deprecation warning could perhaps have been more explicit about that.

However, I am surprised if this intervention actually helps -- what's deprecated is importing SeedStream via the tensorflow_probability.python.distributions namespace, but fitting.py is already getting it from tensorflow_probability.python.util, which is where it's supposed to come from.

axch:
We have since fixed some instances of this deprecation warning in other places in `sts/`.  Based on inactivity, I assume the problem has been fixed.

Feel free to reopen the PR (or open a new one) if you are still seeing these deprecation warnings.

PR #646, opened 2019-11-13
Change copula notebook for TF 2.0

sanket-kamthe:
I have changed notebook to be compatible with TF2.0  [tested on tfp-nightly]

List of changes:
1) Remove all sess.run() calls 
2) Organise the marginal plots in Grid rather than just one after other in single for loop
3) Change from **_normed_** keyword to **_density_** to account for matplotlib deprecation warning [The 'normed' kwarg was deprecated in Matplotlib 2.1 and will be removed in 3.1. Use 'density' instead.]

Question:
Is there a better way do contour plots than jet cmap ? sns kde plot ?


sanket-kamthe:
@googlebot  I signed it !

davmre:
Thanks for updating this and making a PR, Sanket! The changes look great to me. Just a couple of nits:

In the import block, can you replace 
`import tensorflow as tf` with `import tensorflow.compat.v2 as tf`,
and add a call to tf.enable_v2_behavior() at the end of the imports?
This will make sure it works with all recent versions of TF (including the 1.14 and 1.15 releases and Google's internal codebase).

Maybe change the name `copula_eval` (in cell 8) since it's no longer the result of any evaluation in the TF1 session sense. E.g., maybe just `copula_probs`?

FWIW, the custom `Concat` bijector could now be deleted and replaced with the built in [`tfb.Blockwise`](https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/Blockwise) (which didn't exist when this notebook was first written). Feel free not to do that, but if you want to take a stab at it, that'd also be awesome!

(also: don't worry about the failing Travis build. As long as the notebook runs for you and passes our internal CI system when we pull it in, it's good to go).

sanket-kamthe:
Uhh tfb.Blockwise looks nice! Didn't know about this one. I will add that as well.

SiegeLordEx:
> Is there a better way do contour plots than jet cmap ? sns kde plot ?

As a quick replacement, use contourf (note the f) and literally anything other than jet (viridis is alright)

davmre:
Thanks again Sanket!

During our internal review @csuter pointed out that the custom NormalCDF bijector in the notebook can also now be replaced by the built-in `tfb.NormalCDF()`. I went ahead and made that change to the checked-in version, so no action required on your part, but just fyi in case you're looking to build off of this example. :-)

Issue #662, opened 2019-11-29
`bijectors.RealNVP` should allow `num_masked` to be defined as proportions


hartikainen:
It would be nice if the RealNVP template allowed the number of masked elements to be defined as proportions instead of just integers. Right now the `num_masked` should be an integer in `[-D+1, D-1]`.  It should be unambiguous if we allowed the user to specify `num_masked` as a float between `[-1, 1]`, and the implementation would be quite trivial.

I'll open a PR for this, and the details can be discussed there. I can't see any reason why this couldn't be allowed, but if anyone has objections to it, please let me know.

SiegeLordEx:
Fixed by #663

PR #663, opened 2019-11-29
Real nvp floating num masked

hartikainen:
As mentioned in #662, it would be nice if the number of masked elements in RealNVP could be defined as proportions instead of just integers.

Happy to take feedback on this.

### DIFF
--- a/tensorflow_probability/python/bijectors/real_nvp.py
+++ b/tensorflow_probability/python/bijectors/real_nvp.py
@@ -299,7 +299,7 @@ class RealNVP(bijector_lib.Bijector):
 
     return self._bijector_fn(x0, self._bijector_input_units(),
                              **condition_kwargs).forward_log_det_jacobian(
-                               x1, event_ndims=1)
+                                 x1, event_ndims=1)
 
   def _inverse_log_det_jacobian(self, y, **condition_kwargs):
     self._cache_input_depth(y)
@@ -311,7 +311,7 @@ class RealNVP(bijector_lib.Bijector):
 
     return self._bijector_fn(y0, self._bijector_input_units(),
                              **condition_kwargs).inverse_log_det_jacobian(
-                               y1, event_ndims=1)
+                                 y1, event_ndims=1)
 
 
 def real_nvp_default_template(hidden_layers,
--- a/tensorflow_probability/python/bijectors/real_nvp.py
+++ b/tensorflow_probability/python/bijectors/real_nvp.py
@@ -136,7 +136,8 @@ class RealNVP(bijector_lib.Bijector):
   """
 
   def __init__(self,
-               num_masked,
+               num_masked=None,
+               proportion_masked=None,
                shift_and_log_scale_fn=None,
                bijector_fn=None,
                is_constant_jacobian=False,
@@ -145,14 +146,18 @@ class RealNVP(bijector_lib.Bijector):
     """Creates the Real NVP or NICE bijector.
 
     Args:
-      num_masked: Python `int` or `float` indicating the number of units of the
-        event that should should be masked. If `int`, must be in the closed
-        interval `[0, D-1]`, where `D` is the event size of the base
-        distribution. If the value is negative, then the last `d` units of the
-        event are masked instead. If `float`, must be in the closed interval
+      num_masked: Python `int`, indicating the number of units of the
+        event that should should be masked. Must be in the closed interval
+        `[0, D-1]`, where `D` is the event size of the base distribution.
+        If the value is negative, then the last `d` units of the event are
+        masked instead. Must be `None` if `proportion_masked` is defined.
+      proportion_masked: Python `float`, indicating the number of units of the
+        event that should should be masked. Must be in the closed interval
         `[-1, 1]`, and the value represents the proportion of the values to be
-        masked. If negative, then the last proportion of units will of the
-        are masked instead.
+        masked. The final number of values to be masked will be the input size
+        times the proportion, rounded to the the nearest integer towards zero.
+        If negative, then the last proportion of units will of the are masked
+        instead. Must be `None` if `num_masked` is defined.
       shift_and_log_scale_fn: Python `callable` which computes `shift` and
         `log_scale` from both the forward domain (`x`) and the inverse domain
         (`y`). Calculation must respect the 'autoregressive property' (see class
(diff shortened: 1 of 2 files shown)


tensorflow_probability/python/bijectors/real_nvp.py
```diff
@@ -206,24 +209,37 @@ def _bijector_fn(x0, input_depth, **condition_kwargs):
         validate_args=validate_args,
         name=name)
 
+  @property
+  def num_masked(self):
+    if isinstance(self._num_masked, (int, np.integer)):
+      return self._num_masked
+    elif if np.issubdtype(self._num_masked, np.floating):
+      return np.floor(self._input_depth * self._num_masked)
```
hartikainen: Should this be `floor` for both negative and positive values?

tensorflow_probability/python/bijectors/real_nvp.py
```diff
@@ -171,14 +176,12 @@ def __init__(self,
       name: Python `str`, name given to ops managed by this object.
 
     Raises:
-      ValueError: If num_masked < 0.
```
hartikainen: This is incorrect and should be removed even if this PR doens't get merged.

tensorflow_probability/python/bijectors/real_nvp.py
```diff
@@ -206,24 +209,37 @@ def _bijector_fn(x0, input_depth, **condition_kwargs):
         validate_args=validate_args,
         name=name)
 
+  @property
+  def num_masked(self):
+    if isinstance(self._num_masked, (int, np.integer)):
+      return self._num_masked
+    elif np.issubdtype(type(self._num_masked), np.floating):
+      return int(np.floor(self._input_depth * self._num_masked))
```
hartikainen: Should this be floor for both negative and positive values?

SiegeLordEx: I think we should round towards zero, so flipping the sign doesn't alter the number of units masked.

tensorflow_probability/python/bijectors/real_nvp.py
```diff
@@ -144,10 +145,14 @@ def __init__(self,
     """Creates the Real NVP or NICE bijector.
 
     Args:
-      num_masked: Python `int` indicating that the first `d` units of the event
-        should be masked. Must be in the closed interval `[0, D-1]`, where `D`
-        is the event size of the base distribution. If the value is negative,
-        then the last `d` units of the event are masked instead.
+      num_masked: Python `int` or `float` indicating the number of units of the
```
SiegeLordEx: I'm not super happy with reusing the argument for this purpose, can we add a second argument (e.g. `fraction_masked` and use that instead?).

tensorflow_probability/python/bijectors/real_nvp.py
```diff
@@ -144,10 +145,14 @@ def __init__(self,
...
+        event that should should be masked. If `int`, must be in the closed
+        interval `[0, D-1]`, where `D` is the event size of the base
+        distribution. If the value is negative, then the last `d` units of the
+        event are masked instead. If `float`, must be in the closed interval
+        `[-1, 1]`, and the value represents the proportion of the values to be
+        masked. If negative, then the last proportion of units will of the
+        are masked instead.
```
SiegeLordEx: This needs to say how we round it.

tensorflow_probability/python/bijectors/real_nvp.py
```diff
@@ -206,24 +209,37 @@ def _bijector_fn(x0, input_depth, **condition_kwargs):
         validate_args=validate_args,
         name=name)
 
+  @property
+  def num_masked(self):
```
SiegeLordEx: I don't think we need for this to be public, would you mind making this private instead? I'm particularly concerned that this will raise an error if `_input_depth` is `None`. We could document this, but if we don't need to have this public, then there's no need to worry about this.

Could make it a `_get_num_masked()` to avoid the name collision.

hartikainen:
Hey @SiegeLordEx, thanks a lot for the feedback! I've addressed your comments and split the `num_masked` into two different arguments. Let me know if there's anything else that you think could be improved.

tensorflow_probability/python/bijectors/real_nvp.py
```diff
@@ -171,20 +181,36 @@ def __init__(self,
...
+      self._reverse_mask = self._num_masked < 0
+    else:
+      assert np.issubdtype(type(fraction_masked), np.floating), (
+          "`fraction_masked` should be a float. Got: {} of type {}"
+          "".format(fraction_masked, type(fraction_masked)))
+      self._num_masked = None
+      self._fraction_masked = float(fraction_masked)
+      self._reverse_mask = self._fraction_masked < 0
+
+    if shift_and_log_scale_fn is not None and bijector_fn is not None:
```
hartikainen: Note that these changes are not strictly needed for this PR. But I think this way of checking the `None`s is better practice than what it used to be (`bool(shift_and_log_scale_fn) == bool(bijector_fn)`).

tensorflow_probability/python/bijectors/real_nvp_test.py
```diff
@@ -60,25 +64,39 @@ def testBijectorWithTrivialTransform(self):
...
+          num_masked
+          if num_masked is not None
+          else np.floor(input_depth * fraction_masked))
+
+      self.assertEqual(flip_nvp._masked_size, expected_num_masked)
+
+      _, x2_ = np.split(x_, [input_depth - abs(flip_nvp._masked_size)], axis=-1)
+
+      # Check latter half is the same after passing thru reversed mask RealNVP.
+      _, forward_x2 = tf.split(
```
hartikainen: Not sure about the linting here. Is there a command I can run to lint with the tfp styles?

SiegeLordEx: You can do `pylint --rcfile=testing/pylintrc <python files>`.

hartikainen:
The linting should be fixed now!

tensorflow_probability/python/bijectors/real_nvp_test.py
SiegeLordEx: It's still not happy about this line. I think you'll probably need to disable the lint here as it seems like a false positive. Just add a '  # pylint: disable=<lint name' here.

SiegeLordEx:
Thanks! Looks like our Travis build is kind of borked, so I'll take it from here. I'll send it to our internal review system and once that is merged, this'll get closed automatically.

hartikainen:
Nice, thanks @SiegeLordEx!

Issue #607, opened 2019-10-14
Any plans to update the examples to TensorFlow 2?


nbro:
The examples in https://github.com/tensorflow/probability/tree/master/tensorflow_probability/examples do not run with TensorFlow 2, but it's been a while the stable TF2 is available. So, any plans to update the examples to TensorFlow 2 (eager execution)? I am particularly interested in https://github.com/tensorflow/probability/blob/master/tensorflow_probability/examples/bayesian_neural_network.py.

Pyrsos:
I am interesting in giving this a go. Can I be assigned to work on it?

nbro:
@Pyrsos Well, if you have a go at it, share it with us, please! However, note that there are a few bugs, which may prevent you from updating certain examples to TF2. See https://github.com/tensorflow/probability/issues/620.

patelmiteshn:
Is there any timeline on this or has anyone played with any examples ?

brianwa84:
We have taken a pass over the examples to assess what work might need to be done (which might be a bit out of date, as @davmre has already made some updates). If anyone wants to take some of these on, please feel free to send PRs, esp for the easy-ish ones!


File | Purpose | Showcases | What needs work | Proposal
-- | -- | -- | -- | --
tensorflow_probability/examples/bayesian_neural_network.py | Trains a Bayesian neural network to classify MNIST digits. | tfp.layers.Convolution2DFlipouttfp.layers.DenseFlipoutinline in a keras modelCategorical output (but not tfpl.DistributionLambda) | TF2 (placeholder, session)tf.compat.v1.metrics.accuracytf.compat.v1.data.make_one_shot_iterator |  
tensorflow_probability/examples/cifar10_bnn.py | Trains a Bayesian neural network to classify CIFAR-10 images. | tfd.Categorical(+bayesian_{vgg,resnet}) | TF2 (placeholder, session)tf.compat.v1.metrics.accuracymigrate to colab?uses tensorflow_probability.examples.models.* | Easy to update
tensorflow_probability/examples/deep_exponential_family.py | Trains a sparse Gamma deep exponential family on NIPS 2011 conference papers. | tfp.edward2 (could be JDCoroutine) | tf.compat.v1 variables, summary, global stepuses edward2Looks like great opp to showcase JDCoroutineUses NIPS2011 papers dataset (migrate to TFDS?) | Delete or rewrite
[...]

Issue #620, opened 2019-10-25
tensorflow.python.eager.core._SymbolicException: Inputs to eager execution function cannot be Keras symbolic tensors, but found


nbro:
**System information**
- Have I written custom code: Yes.
- OS Platform and Distribution: Mac OS Catalina: 10.15 (19A602)
- TensorFlow installed from: binary
- TensorFlow version: 2.0.0
- Python version: 3.7.4
- GPU model and memory: Intel Iris Pro 1536 MB

**Describe the current behavior**

I am getting the error

> tensorflow.python.eager.core._SymbolicException: Inputs to eager execution function cannot be Keras symbolic tensors, but found [<tf.Tensor 'conv2d_flipout/divergence_kernel:0' shape=() dtype=float32>, <tf.Tensor 'conv2d_flipout_1/divergence_kernel:0' shape=() dtype=float32>, <tf.Tensor 'conv2d_flipout_2/divergence_kernel:0' shape=() dtype=float32>]

**Describe the expected behavior**

No error.

**Code to reproduce the issue**

```
from __future__ import print_function

import tensorflow as tf
import tensorflow_probability as tfp


def get_bayesian_model(input_shape=None, num_classes=10):
    model = tf.keras.Sequential()
    model.add(tf.keras.layers.Input(shape=input_shape))
    model.add(tfp.layers.Convolution2DFlipout(6, kernel_size=5, padding="SAME", activation=tf.nn.relu))
    model.add(tf.keras.layers.Flatten())
    model.add(tfp.layers.DenseFlipout(84, activation=tf.nn.relu))
    model.add(tfp.layers.DenseFlipout(num_classes))
    return model

def get_mnist_data(normalize=True):
    img_rows, img_cols = 28, 28
    (x_train, y_train), (x_test, y_test) = tf.keras.datasets.mnist.load_data()

    if tf.keras.backend.image_data_format() == 'channels_first':
        x_train = x_train.reshape(x_train.shape[0], 1, img_rows, img_cols)
        x_test = x_test.reshape(x_test.shape[0], 1, img_rows, img_cols)
        input_shape = (1, img_rows, img_cols)
    else:
        x_train = x_train.reshape(x_train.shape[0], img_rows, img_cols, 1)
        x_test = x_test.reshape(x_test.shape[0], img_rows, img_cols, 1)
        input_shape = (img_rows, img_cols, 1)

    x_train = x_train.astype('float32')
    x_test = x_test.astype('float32')

    if normalize:
[...]

PR #698, opened 2019-12-24
Update tensorflow probability examples

Pyrsos:
As addressed on issue #607, the tensorflow probability examples currently use the 'compat' tensorflow subpackage for TF2 compatibility. In this PR I have updated two of the examples ('logistic_regression ' and 'bayesian_neural_network.py') to be compatible with the TF2 paradigms, using the Keras high-level API. I have kept the same functionality for plots and retained the model architectures. On the other hand I have replaced the 'tf.data' input pipeline functions with the Keras sequence class, as these can be used very easily with the 'model.(fit|evaluate|predict)_generator' functions. Also on the 'keras_bayesian_neural_network.py' I have replaced the mnist datasets path (which appears to no longer be in that location) with the keras.datasets.mnist path.

One further issue I have noticed is when eager execution is enabled, the Keras model.fit_generator operation will return the loss metric as a Tensor object, while the accuracy will be returned as a scalar (float value). For this reason the 'keras_bayesian_neural_network.py' uses the 'model.train_on_batch' operation instead (which also allows for plotting of weights/images and performing monte carlo sampling on the validation set while training).


tensorflow_probability/examples/keras_bayesian_neural_network.py
```diff
@@ -0,0 +1,360 @@
...
+
+  model = create_model()
+
+  print(" ... Training convolutional neural network")
+  for epoch in range(FLAGS.num_epochs):
+    epoch_accuracy = []
+    for step, (batch_x, batch_y) in enumerate(train_seq):
+      # Eager mode returns a Tensor objest and not a scalar
+      # value for the loss, therefore only the mean accuracy
+      # is displayed.
```
davmre: Eager mode vs graph mode per se shouldn't change the shapes of Tensors; I'm guessing this is a property of keras (or tf2 keras). I'd maybe remove this comment since it doesn't add all that much information that wasn't already present in the fact that you're calling `np.mean` below.

tensorflow_probability/examples/keras_logistic_regression.py
```diff
@@ -0,0 +1,231 @@
...
+# Dependency imports
+from absl import flags, app
+from matplotlib import cm
+from matplotlib import figure
+from matplotlib.backends import backend_agg
+import numpy as np
+import tensorflow as tf
+import tensorflow_probability as tfp
+
+TFD = tfp.distributions
```
davmre: lower-case `tfd` (as in the BNN)

tensorflow_probability/examples/keras_logistic_regression.py
```diff
@@ -0,0 +1,231 @@
...
+      kernel_divergence_fn=kl_divergence_function)(input_layer)
+
+  # Model compilation.
+  model = tf.keras.Model(inputs=input_layer, outputs=dense_layer)
+  optimizer = tf.keras.optimizers.Adam(lr=FLAGS.learning_rate)
+  # We use the binary_crossentropy loss since this toy example contains
+  # two labels. The Keras API will then automatically add the
+  # Kullback-Leibler divergence (contained on the individual layers of
+  # the model), to the cross entropy loss, effectively
+  # calcuating the Evidence Lower Bound Loss (ELBO)
```
davmre: same nit as BNN: this is technically the negated ELBO

tensorflow_probability/examples/BUILD
```diff
@@ -27,6 +27,45 @@ package(
 
 exports_files(["LICENSE"])
 
+py_binary(
+    name = "keras_bayesian_neural_network",
```
davmre: Instead of adding new targets, can you go ahead and replace the old ones? (i.e., just overwrite `bayesian_neural_network.py` and `logistic_regression.py` with your new versions). I don't think there's a lot of value in us keeping around out-of-date examples (debatably there's negative value :).

tensorflow_probability/examples/keras_bayesian_neural_network.py
```diff
@@ -0,0 +1,360 @@
...
+import warnings
+
+# Dependency imports
+from absl import flags, app
+import matplotlib
+matplotlib.use("Agg")
+from matplotlib import figure  # pylint: disable=g-import-not-at-top
+from matplotlib.backends import backend_agg
+import numpy as np
+import tensorflow as tf
```
davmre: Can you do the import as

```python
tf.enable_v2_behavior()
```

That will help this work with recent versions of TF1 (and Google's internal TF, since we haven't fully moved to TF2 APIs by default yet).

(same for the logistic regression example)

tensorflow_probability/examples/keras_logistic_regression.py
```diff
@@ -0,0 +1,231 @@
...
+import os
+
+# Dependency imports
+from absl import flags, app
+from matplotlib import cm
+from matplotlib import figure
+from matplotlib.backends import backend_agg
+import numpy as np
+import tensorflow as tf
+import tensorflow_probability as tfp
```
davmre: see my comment re imports on the BNN

Pyrsos:
@davmre No worries about the delay, just had some spare time and thought it'd be good to give it a go. Thanks for the comments, I have addressed these and pushed the new changes. Because I have now renamed the files the git log might be a bit different, hopefully it will not be a massive issue. Let me know if there is anything else you would like me to adjust.

I will try and update some more examples in the coming weeks.

tensorflow_probability/examples/BUILD
```diff
@@ -91,10 +52,11 @@ py_library(
 py_test(
     name = "bayesian_neural_network_test",
     size = "small",
-    srcs = ["bayesian_neural_network.py"],
+    srcs = [bayesian_neural_network.py"],
```
davmre: Looks like this line dropped a quote - would you mind adding it back in? That'll help me pull this in to our internal repo.

Pyrsos: Apologies, completely missed this. Added this back in.

tensorflow_probability/examples/keras_bayesian_neural_network.py
```diff
@@ -0,0 +1,360 @@
...
+                                     FLAGS.model_dir,
+                                     "epoch{}_step{:05d}_weights.png".format(
+                                         epoch, step)))
+          plot_heldout_prediction(heldout_seq.images, probs,
+                                  fname=os.path.join(
+                                      FLAGS.model_dir,
+                                      "epoch{}_step{}_pred.png".format(
+                                          epoch, step)),
+                                  title="mean heldout logprob {:.2f}"
+                                  .format(heldout_log_prob))
```
nbro: Why don't use Keras' `fit` method? The code still looks unnecessarily verbose. I had actually updated this example to TF2, but I hadn't opened a pull request.

Pyrsos: The main reason for using the `train_on_batch` instead of the `fit` was because I am trying to preserve the operations of the original example, as I think these add some valuable insight on how bayesian networks are trained and their behaviour regarding uncertainty. I thought that this was the easier way to implement those instead of using some custom Keras callbacks, which I think would be more difficult to follow. When I first submitted the PR this example was working fine, still works for version 2.0. But since version 2.1 I am also getting the problem you described in issue #620. 

@davmre Is there any way you might be able to point me to a starting point for dealing with issue #620. I am making the naive assumption that the fix should not be too complicated, since as @nbro says the problem appears to be caused by the Convolution2DFlipout layers (specifically the kernel divergence) and not the DenseFlipout layers.

nbro: I don't agree with you. I think that, as a rule of thumb, people should use `fit` and callbacks are perfectly fine and intuitive. If you decide to use `fit`, you can solve that issue by following the instructions here: https://github.com/tensorflow/tensorflow/issues/33729#issuecomment-562656807.

Pyrsos: @nbro I am not particularly fond of this solution because I can similarly solve the issue by disabling eager execution. In both cases the underlying cause is still not treated. However, I will make the changes you suggest (both using the `fit` function and callbacks) if @davmre also agrees with that implementation.

nbro: "In both cases the underlying cause is still not treated.", because this is a bug in TensorFlow! Hopefully, someone is already taking care of this issue. `experimental_run_tf_function`, in fact, exists because there's an experimental feature related to distributed training.

brianwa84:
@davmre do you want to follow-up & import this, or need any more changes?

nbro:
@brianwa84, @davmre You should not merge this code. I have a better version than this one. In fact, I've noticed that, in this new version of the code, the Bayesian model is trained with the categorical cross-entropy, but this won't work well (the model should never converge). You'd better create a custom function that calculates the negative log-likelihood (which I did in my updated version). I may perform the pull request later.

Pyrsos:
@nbro I am testing this script and it is converging for me. Can you let me know what happens when you run it? As I stated before I am happy to make any changes that the reviewers @brianwa84 @davmre suggest.

nbro:
@Pyrsos What accuracy do you obtain after the first epoch? And what do you mean by converging? Do you mean that the loss decreases?

Pyrsos:
@nbro Test set accuracy after the first epoch is above 95%. And, yes when I mentioned that the model is converging I was referring to the loss decreasing.

nbro:
@Pyrsos I executed part of your code, and, as you say, the accuracy is "decent" at the end of the first epoch. I have an example where the accuracy remains at 10% throughout training (even though the loss decreases), but this was due to the fact that I was not using the softmax function in the last layer.

nbro:
@Pyrsos How do the distributions of the weights (mean and std) look like after the 1st, 5th and 10 epoch? Have you noticed any particular pattern? For example, the means do not change much and actually tend to concentrate around zero, while the variances tend to increase?

nbro:
@Pyrsos Anyway, given your concerns about the compatibility with the original example (and as I said above), you should use a loss function that uses the `tfd.Categorical` rather than the CCE.

Pyrsos:
@nbro I am glad that the code worked for you. With regard to the weights, I have noticed the same behaviour you are describing, but have not gone much deeper in experimenting with this. It is definitely an interesting research area to check, especially when considering different prior/posterior distribution settings.

With regard to the loss function, I think that the terms `negative log likelihood` and `cross entropy` are interchangeable. In the documentation for the DenseFlipout (https://www.tensorflow.org/probability/api_docs/python/tfp/layers/DenseFlipout), the negative log likelihood is defined as:
```
neg_log_likelihood = tf.nn.softmax_cross_entropy_with_logits(
    labels=labels, logits=logits)
```
Also, for the `tfd.Categorical` class, when the `log_prob` method is called, the `sparse_softmax_cross_entropy_with_logits` is returned. https://github.com/tensorflow/probability/blob/356cfddef026b3339b8f2a81e600acd2ff8e22b4/tensorflow_probability/python/distributions/categorical.py#L275-L283
In that case the sparse version is used, which explains why the inputs had to be integer values instead of one-hot (categorical) values. Notice also that the function returns the negative value, which is why the original example had to compensate by doing:
https://github.com/tensorflow/probability/blob/07b6fe40eee693cf7586e069707b18bc2ba63d7c/tensorflow_probability/examples/bayesian_neural_network.py#L261-L263

The keras `categorical_cross_entropy` string when passed to the `model.compile` method also calls the `softmax_cross_entropy_with_logits_v2` (I guess the v2 is a remnant from TF v1). 
https://github.com/tensorflow/tensorflow/blob/cf7fcf164c9846502b21cebb7d3d5ccf6cb626e8/tensorflow/python/keras/backend.py#L4487-L4504

So I think that is what is happening, but please correct me if I have this wrong and there is something else going on @brianwa84 @davmre

nbro:
@Pyrsos In any case, TFP provides abstractions for distributions, which should be used, no? I know that NLL is equivalent to CE.

nbro:
@Pyrsos I noticed that in your updated example https://github.com/Pyrsos/probability/blob/update_examples/tensorflow_probability/examples/bayesian_neural_network.py, you are using `compile(..., experimental_run_tf_function=False)`, but the option `experimental_run_tf_function=False` is only required if you use `fit` to train the model, which isn't the case, given that you are still using the hardcoded loop. Or do you also get the error described in the issue https://github.com/tensorflow/tensorflow/issues/33729 if you use `train_on_batch`? I haven't tried it, so it's possible this problem also occurs with `train_on_batch`.

nbro:
@brianwa84, @davmre Given that there are two models under the folder `models`, wouldn't it also be opportune to write the definition of the basic Bayesian CNN (and, actually, the definition of any Bayesian model) in a standalone module under this folder `models`, which would then be imported from the corresponding example?

Pyrsos:
@nbro Yes, the error you are describing also appears in `train_on_batch`, which is why as you suggested I used the `experimental_run_tf_function` on `model.compile`. As far as I can tell it is associated with the divergence kernels of the CNN layers. Dense layers appear to be unaffected.

nbro:
@Pyrsos No, you are confused: I hadn't suggested using `experimental_run_tf_function=False` because you were using `train_on_batch`. In fact, you were using `train_on_batch` before my suggestion (see https://github.com/Pyrsos/probability/blob/a945f418533b5827ff1966c8186e58604ceb43fc/tensorflow_probability/examples/bayesian_neural_network.py, where you are using no `experimental_run_tf_function=False`, but using `train_on_batch`), and you didn't say you were getting any error. 

And, yes, as I had already reported in several Github issues, yes, the issue is only apparently related to Bayesian convolution layers.

Pyrsos:
@nbro If you set `experimental_run_tf_function=True` with `train_on_batch` you will get the error you are describing here https://github.com/tensorflow/tensorflow/issues/33729. The error is not exclusive to the `fit` operation.

nbro:
@Pyrsos So, why did you create this pull request if, initially, you were not using `experimental_run_tf_function=False`, but were using `train_on_batch`, so you were getting an error?

Pyrsos:
@nbro Original version of the script was using v2.0 where `train_on_batch` did not produce an error. The error originated when I updated to v2.1, at which point the parameter `experimental_run_tf_function=False` was added to the `compile` function.

nbro:
@Pyrsos It's strange. You're the second person saying that `experimental_run_tf_function=False` was added in TF 2.1, but I had been using this option in TF 2.0 too.

davmre:
Sorry for the long delay on this; it fell through the cracks partly because I was on vacation for a while. We're putting this through internal review now; hopefully it will go in within the next couple of days.

A couple of notes:
- We usually ask that contributors run `pylint` with our [custom configuration](https://github.com/tensorflow/probability/blob/master/testing/pylintrc) and fix any errors reported. In this case, I went ahead and did that for you in our internal system, so you don't need to worry about it. (but it's helpful for future contributions!)
- The `model.build` workaround to #620 appears to work here. I've gone ahead and added that internally. We're following up with the Keras folks to try to understand the underlying issue.

nbro:
@davmre The merged examples now have both options `experimental_run_tf_function=False` and your suggestion `model.build(...)`. Maybe this was not intended.

Pyrsos:
Thank you very much @davmre ! I learned a lot from this process, and I am looking forward to contributing again!

Issue #712, opened 2020-01-08
Laplace distirbution lacks quantile method.


Dpananos:
## Issue

The Laplace distribution currently lacks a quantile method.  It would be useful to have this method.

## Example
The following code results in an error
```python
import tensorflow as tf
import tensorflow_probability as tfp

tfd = tfp.distributions

tfd.Laplace(loc=0, scale=1).quantile([0.25, 0.5, 0.75])
```
```

---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
<ipython-input-16-fe6e5b3aceb9> in <module>
      6 tfd.Normal(loc=0, scale=1).quantile([0.25, 0.5, 0.75])
      7 
----> 8 tfd.Laplace(loc=0, scale=1).quantile([0.25, 0.5, 0.75])

~/opt/miniconda3/envs/gymbo/lib/python3.7/site-packages/tensorflow_probability/python/distributions/distribution.py in quantile(self, value, name, **kwargs)
   1097         values of type `self.dtype`.
   1098     """
-> 1099     return self._call_quantile(value, name, **kwargs)
   1100 
   1101   def _variance(self, **kwargs):

~/opt/miniconda3/envs/gymbo/lib/python3.7/site-packages/tensorflow_probability/python/distributions/distribution.py in _call_quantile(self, value, name, **kwargs)
   1077       value = _convert_to_tensor(
   1078           value, name='value', dtype_hint=self.dtype)
-> 1079       return self._quantile(value, **kwargs)
   1080 
   1081   def quantile(self, value, name='quantile', **kwargs):

~/opt/miniconda3/envs/gymbo/lib/python3.7/site-packages/tensorflow_probability/python/distributions/distribution.py in _quantile(self, value, **kwargs)
   1071   def _quantile(self, value, **kwargs):
   1072     raise NotImplementedError('quantile is not implemented: {}'.format(
-> 1073         type(self).__name__))
   1074 
   1075   def _call_quantile(self, value, name, **kwargs):

NotImplementedError: quantile is not implemented: Laplace
```

Contrast this with `distributions.Normal`

```python
import tensorflow as tf
import tensorflow_probability as tfp

tfd = tfp.distributions

tfd.Normal(loc=0, scale=1).quantile([0.25, 0.5, 0.75])
```
`<tf.Tensor: id=711, shape=(3,), dtype=float32, numpy=array([-0.6744898,  0.       ,  0.6744898], dtype=float32)>`

PR #713, opened 2020-01-08
Add quantile method to distirbutions.Laplace

Dpananos:
Should solve #712 by giving the Laplace distribution a quantile method.

### DIFF
--- a/tensorflow_probability/python/distributions/laplace.py
+++ b/tensorflow_probability/python/distributions/laplace.py
@@ -203,9 +203,7 @@ class Laplace(distribution.Distribution):
     return self._mean()
 
   def _quantile(self, p):
-    return tf.where(p > 0.5, 
-            tf.where(p > 1, np.inf, self.loc - self.scale * tf.math.log(2 * (1 - p))), 
-            tf.where(p < 0, -np.inf, self.loc + self.scale * tf.math.log(2 * p)))
+    return tf.where(p > 0.5, self.loc - self.scale * tf.math.log(2 * (1 - p)), self.loc + self.scale * tf.math.log(2 * p))
 
   def _z(self, x):
     return (x - self.loc) / self.scale
--- a/tensorflow_probability/python/distributions/laplace.py
+++ b/tensorflow_probability/python/distributions/laplace.py
@@ -202,6 +202,9 @@ class Laplace(distribution.Distribution):
   def _mode(self):
     return self._mean()
 
+  def _quantile(self, x):
+    return tf.where(x > 0.5, -tf.math.log(2*(1-x)), tf.math.log(2*x))
+
   def _z(self, x):
     return (x - self.loc) / self.scale
 
--- a/tensorflow_probability/python/distributions/laplace.py
+++ b/tensorflow_probability/python/distributions/laplace.py
@@ -203,7 +203,9 @@ class Laplace(distribution.Distribution):
     return self._mean()
 
   def _quantile(self, x):
-    return tf.where(x > 0.5,self.loc - self.scale*tf.math.log(2*(1-x)), self.loc + self.scale*tf.math.log(2*x))
+    loc = tf.convert_to_tensor(self.loc)
+    scale = tf.convert_to_tensor(self.scale)
+    return tf.where(x > 0.5,loc - scale*tf.math.log(2*(1-x)), loc + scale*tf.math.log(2*x))
 
   def _z(self, x):
     return (x - self.loc) / self.scale
--- a/tensorflow_probability/python/distributions/laplace.py
+++ b/tensorflow_probability/python/distributions/laplace.py
@@ -203,7 +203,9 @@ class Laplace(distribution.Distribution):
     return self._mean()
 
   def _quantile(self, p):
-    return tf.where(p > 0.5, self.loc - self.scale*tf.math.log(2*(1-p)), self.loc + self.scale*tf.math.log(2*p))
+    return tf.where(p > 0.5, 
+            tf.where(p > 1, np.inf, self.loc - self.scale * tf.math.log(2 * (1 - p))), 
+            tf.where(p < 0, -np.inf, self.loc + self.scale * tf.math.log(2 * p)))
 
   def _z(self, x):
     return (x - self.loc) / self.scale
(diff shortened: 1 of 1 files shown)


Dpananos:
@googlebot I signed it!

tensorflow_probability/python/distributions/laplace.py
```diff
@@ -202,6 +202,11 @@ def _median(self):
   def _mode(self):
     return self._mean()
 
+  def _quantile(self, x):
+    loc = tf.convert_to_tensor(self.loc)
```
junpenglao: You shouldn't use the `convert_to_tensor` here, using `self.loc` and `self.scale` directly should be fine here.

tensorflow_probability/python/distributions/laplace.py
```diff
@@ -202,6 +202,11 @@ def _median(self):
...
+    scale = tf.convert_to_tensor(self.scale)
+    return tf.where(x > 0.5,loc - scale*tf.math.log(2*(1-x)), loc + scale*tf.math.log(2*x))
```
junpenglao: You might need the double where trick similar to https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/internal/special_math.py#L151-L154 - would it be possible to write the laplace quantile function as some scaling of the Normal quantile function? that would make it a bit easier to implement and test I think.

tensorflow_probability/python/distributions/laplace.py
Dpananos: Noted and will change.

tensorflow_probability/python/distributions/laplace.py
Dpananos: Not sure if I understand where the double where would be used.  As for writing the laplacian quantile function as a scaling of the Normal quantile function, a scaling *might* exist but I'm not immediately familiar with it.  I can look around for it.

tensorflow_probability/python/distributions/laplace.py
```diff
@@ -202,6 +202,9 @@ def _median(self):
   def _mode(self):
     return self._mean()
 
+  def _quantile(self, x):
```
junpenglao: You should rename input from `x` to `p` for consistency with the rest of the distributions.

Dpananos: Done.

Dpananos:
Instead of a double-where trick, would it make more sense to use if/else?  The implementation I provide is inspired by scipy's implementation [here](https://github.com/scipy/scipy/blob/v1.4.1/scipy/stats/_continuous_distns.py#L4061).  The percentile function is a piecewise defined function, so there are a few ways to get around the `tf.where`

junpenglao:
Using `tf.where` is perfered instead of `tf.cond`, as IIUC `tf.cond` communicate with CPU so you should avoid it if possible.

Dpananos:
Added a double where to avoid `nan` when quantiles above 1 or below 0 are passed.  Behavior is similar to Normal now.

junpenglao:
Sorry I might have confused you, I think the double where trick is probably not necessary in this case, but I will wait for review from @jvdillon or @brianwa84

csuter:
Is there a test that takes gradients of quantile that exercise edge cases of the `where`? The double-where trick is to work around propagation of NaNs from the discarded branches of where's in gradients, which are computed as `(cond) * grad(left) + (1 - cond) * grad(right)` which produces NaNs when *either* left or right is NaN (or `Inf`, since `0 * inf = NaN`), even if it's not the chosen branch.

csuter:
If there's no chance of Inf or NaN on either side of the branch I believe double where is unnecessary.

Dpananos:
There is a risk when p<0 or p>1.  If you call `Laplace(loc=0, scale=1).quantile(1.1)` then this returns nan.  Under the most recent commit, it returns inf, similar to normal.

csuter:
I think `p` outside those ranges should raise an exception; this may be already handled in the base Distribution class when `validate_args` flag is set to `True` (if not it should!). We don't want every Distribution that implements quantile to have to duplicate handling this logic.

If you're then just switching between left and right branches of the symmetric quantile function, i don't think you need a double where.

csuter:
Seems the base class doesn't have this assertion logic. I think for the time being you should not include the conditionals for `p` outside the "intuitive" domain of the quantile function. We might want to add an assertion to the base `Distribution` class later, which checks for valid `p` in the case of `validate_args=True`. But you don't need to handle this explicitly here (c.f. `Normal` quantile, which does not do this).

All that said, if it's possible for the *not chosen* branch of the remaining `tf.where` (checking `p<.5`) to return a `NaN` or `Inf` in the case that it's not chosen, then we do need the double where. If not, we don't. If you do need it, you can unit test by taking gradients in such a case and verifying that without the double-where trick you get NaN, and with the double-where you get the correct value.

Check out instructions here for running unit tests locally: https://github.com/tensorflow/probability/blob/master/CONTRIBUTING.md (it's not as simple as `python -m unittest ... -- you have to use `bazel`). There's a little wrinkle missing there that we've overlooked updating recently. You need to disable some tests that are (known) broken with the `--test_tag_filters=-notap` flag to `bazel`. The ultimate source of truth for bazel tests is here: https://github.com/tensorflow/probability/blob/master/testing/run_tests.sh#L120

Sorry this unit testing story is so messy! Please let us know if you have trouble getting it working. We'll make it a point to update the docs ASAP to reflect the above sublety..

Dpananos:
Hmm, I'm having some trouble getting the unit tests working.  I thought this would be an easy fix.  I suppose not.  I'll maybe try again later this week.

csuter:
Sorry to hear that. We very much welcome the contribution and would like to make it generally easier for people like you to contribute, so I hope you will pick it back up and we can learn from the troubles you are running into.

Dpananos:
Why was this merged?  I didn't run any tests on this PR yet.

brianwa84:
I added a test for you :)

Brian Patton | Software Engineer | bjp@google.com

brianwa84:
You can see it in here https://github.com/tensorflow/probability/commit/ee806c972d585837066674d8511d2aab4b413bb1

PR #723, opened 2020-01-11
New blog URL

amelio-vazquez-reina:
Apologies if you would prefer this request against a different branch. 

The new blog URL is https://blog.tensorflow.org/ and the README originally linked to https://medium.com/tensorflow.


amelio-vazquez-reina:
@googlebot I signed it!

PR #725, opened 2020-01-12
add assert p in [0, 1] in quantile method in distribution base class

### LINKED ISSUE #714: Distribution base class `quantile` should assert `p \in [0, 1]` when `validate_args=True`

blacksde:
Assert the prob/value in the quantile method in distribution base class when validate_args is true

Resolve #714


blacksde:
@googlebot I signed it!

Issue #731, opened 2020-01-14
tfp.util.DeferredTensor should allow access to the value of the tensor through .numpy()


hartikainen:
Hey,

I think it'd be useful if `tfp.util.DeferredTensor` values could be accessed in the same way as the non-deferred variable values. For example, it would be useful if the following was possible:
```python
import tensorflow as tf
import tensorflow_probability as tfp


def main():
    x = tf.Variable(0.5)
    y = tfp.util.DeferredTensor(x, tf.exp)
    print(y.numpy())  # y.transform_fn(y.variables).numpy() works.


if __name__ == '__main__':
    main()
```

However, the example currently fails with `AttributeError: 'DeferredTensor' object has no attribute 'numpy'`.

Is this intentional and there's a fundamental reason for not allowing this for `DeferredTensor`s or is this a feature that could exist? I'd be happy to contribute a fix for this if needed.

brianwa84:
This feature could exist, feel free to send a PR.

csuter:
Thanks, @hartikainen! You can also do tf.convert_to_tensor(y).numpy() (in eager mode), though that's not much shorter than the commented code you wrote :).

Also, check out TransformedVariable for an example like the one you've written. It let's you simply write

```
y = tfp.util.TransformedVariable(np.exp(.5), tf.exp)
```

This class creates the tf.Variable in the backgroundm and support assignments, which "do the right thing" w.r.t. the transformation. Note that it's initialized with the *post-transform* value.

PR #737, opened 2020-01-18
Add DeferredTensor.numpy

hartikainen:
Adds `DeferredTensor.numpy`. See #731.

### DIFF
--- a/tensorflow_probability/python/util/deferred_tensor_test.py
+++ b/tensorflow_probability/python/util/deferred_tensor_test.py
@@ -338,13 +338,13 @@ class DeferredTensorBehavesLikeTensorTest(test_util.TestCase):
   def testMethodNumpy(self):
     x_ = np.array([0., 1.])
     x = tfp.util.DeferredTensor(tf.Variable(x_), tf.math.exp)
+
     self.evaluate([v.initializer for v in x.trainable_variables])
     if tf.executing_eagerly():
       self.assertAllEqual(tf.math.exp(x_).numpy(), x.numpy())
     else:
-      with self.assertRaises(TypeError):
-        for _ in iter(x):
-          pass
+      with self.assertRaises(AttributeError):
+        x.numpy()
 
 
 if __name__ == '__main__':
--- a/tensorflow_probability/python/util/deferred_tensor_test.py
+++ b/tensorflow_probability/python/util/deferred_tensor_test.py
@@ -335,6 +335,17 @@ class DeferredTensorBehavesLikeTensorTest(test_util.TestCase):
         for _ in iter(x):
           pass
 
+  def testMethodNumpy(self):
+    x_ = np.array([0., 1.])
+    x = tfp.util.DeferredTensor(tf.Variable(x_), tf.math.exp)
+    self.evaluate([v.initializer for v in x.trainable_variables])
+    if tf.executing_eagerly():
+      self.assertAllEqual(tf.math.exp(x_).numpy(), x.numpy())
+    else:
+      with self.assertRaises(TypeError):
+        for _ in iter(x):
+          pass
+
 
 if __name__ == '__main__':
   tf.test.main()
--- a/tensorflow_probability/python/util/deferred_tensor_test.py
+++ b/tensorflow_probability/python/util/deferred_tensor_test.py
@@ -343,7 +343,7 @@ class DeferredTensorBehavesLikeTensorTest(test_util.TestCase):
     if tf.executing_eagerly():
       self.assertAllEqual(tf.math.exp(x_).numpy(), x.numpy())
     else:
-      with self.assertRaises(AttributeError):
+      with self.assertRaises(TypeError):
         x.numpy()
 
 
--- a/tensorflow_probability/python/util/deferred_tensor.py
+++ b/tensorflow_probability/python/util/deferred_tensor.py
@@ -290,7 +290,14 @@ class DeferredTensor(tf.Module):
 
   def numpy(self):
     """Returns (copy of) deferred values as a NumPy array or scalar."""
-    return self._value().numpy()
+    try:
+      return self._value().numpy()
+    except AttributeError:
+      if isinstance(self._value(), tf.Tensor):
+        raise TypeError(
+            "'{}.numpy()' only works in eager mode.".format(type(self).__name__))
+      else:
+        raise
 
   def set_shape(self, shape):
     """Updates the shape of this pretransformed_input.
--- a/tensorflow_probability/python/util/deferred_tensor.py
+++ b/tensorflow_probability/python/util/deferred_tensor.py
@@ -288,6 +288,10 @@ class DeferredTensor(tf.Module):
     """The string name of this object."""
     return self._name
 
+  def numpy(self):
+    """Returns (copy of) deferred values as a NumPy array or scalar."""
+    return self._value().numpy()
+
   def set_shape(self, shape):
     """Updates the shape of this pretransformed_input.
 
--- a/tensorflow_probability/python/util/deferred_tensor.py
+++ b/tensorflow_probability/python/util/deferred_tensor.py
@@ -290,14 +290,11 @@ class DeferredTensor(tf.Module):
 
   def numpy(self):
     """Returns (copy of) deferred values as a NumPy array or scalar."""
-    try:
-      return self._value().numpy()
-    except AttributeError:
-      if isinstance(self._value(), tf.Tensor):
-        raise TypeError(
-            "'{}.numpy()' only works in eager mode.".format(type(self).__name__))
-      else:
-        raise
+    value = self._value()
+    if not hasattr(value, 'numpy'):
+      raise TypeError(
+          "DeferredTensor.numpy() is only supported in eager execution mode.")
+    return value.numpy()
 
   def set_shape(self, shape):
     """Updates the shape of this pretransformed_input.


tensorflow_probability/python/util/deferred_tensor_test.py
```diff
@@ -335,6 +335,17 @@ def testOperatorIter(self):
         for _ in iter(x):
           pass
 
+  def testMethodNumpy(self):
```
hartikainen: Is this the right place for the test?

hartikainen:
What's the preferred way of running and debugging tests? I'm currently running `bazel test //tensorflow_probability/python/util:deferred_tensor_test`, and can inspect the logs from that, but I have no clue how to e.g. stop a debugger inside a test. Is there some documentation about this in tfp or tensorflow core?

tensorflow_probability/python/util/deferred_tensor_test.py
csuter: Yeah this is a good place. I'd suggest something like the following here:

```python
x_ = np.array([0., 1.])
x = tfp.util.DeferredTensor(tf.Variable(x_), tf.math.exp)

if tf.executing_eagerly():
  self.assertAllEqual(np.exp(x_), x.numpy())
else:
  with self.assertRaises(...):
    x.numpy()
```

As you probably figured out, we generate a few different test cases for each test method (this is due to the `@test_all_tf_execution_regimes` decorator):
  1. graph mode
  2. eager mode
  3. eager mode with all tf.function wrappers disabled (these can sometimes be generated by library code, and we want to ensure that a user debugging something sees consistent behavior with `tf.function`s enabled and disabled).

As for step-through debugging tests, there might be some helpful hints [here](https://stackoverflow.com/questions/51709590/can-i-use-python-debugger-in-bazel-test) but personally, after 20+ years, I still just use print statements for debugging 😝 ...

tensorflow_probability/python/util/deferred_tensor_test.py
```diff
@@ -335,6 +335,17 @@ def testOperatorIter(self):
...
+    x_ = np.array([0., 1.])
+    x = tfp.util.DeferredTensor(tf.Variable(x_), tf.math.exp)
+
+    self.evaluate([v.initializer for v in x.trainable_variables])
+    if tf.executing_eagerly():
+      self.assertAllEqual(tf.math.exp(x_).numpy(), x.numpy())
+    else:
+      with self.assertRaises(AttributeError):
```
hartikainen: This one raises:
```
  File "/Users/hartikainen/conda/envs/tensorflow-dev/lib/python3.7/site-packages/absl/testing/parameterized.py", line 263, in bound_param_test
    test_method(self, **testcase_params)
  File "/private/var/tmp/_bazel_hartikainen/4f63de4300ef6f519aa77402a0c94964/sandbox/darwin-sandbox/483/execroot/tensorflow_probability/bazel-out/darwin-fastbuild/bin/tensorflow_probability/python/util/deferred_tensor_test.runfiles/tensorflow_probability/tensorflow_probability/python/internal/test_combinations.py", line 312, in decorated
    execute_test_method()
  File "/private/var/tmp/_bazel_hartikainen/4f63de4300ef6f519aa77402a0c94964/sandbox/darwin-sandbox/483/execroot/tensorflow_probability/bazel-out/darwin-fastbuild/bin/tensorflow_probability/python/util/deferred_tensor_test.runfiles/tensorflow_probability/tensorflow_probability/python/internal/test_combinations.py", line 295, in execute_test_method
    test_method(**kwargs_to_pass)
  File "/private/var/tmp/_bazel_hartikainen/4f63de4300ef6f519aa77402a0c94964/sandbox/darwin-sandbox/483/execroot/tensorflow_probability/bazel-out/darwin-fastbuild/bin/tensorflow_probability/python/util/deferred_tensor_test.runfiles/tensorflow_probability/tensorflow_probability/python/util/deferred_tensor_test.py", line 352, in testMethodNumpy
    x.numpy()
  File "/private/var/tmp/_bazel_hartikainen/4f63de4300ef6f519aa77402a0c94964/sandbox/darwin-sandbox/483/execroot/tensorflow_probability/bazel-out/darwin-fastbuild/bin/tensorflow_probability/python/util/deferred_tensor_test.runfiles/tensorflow_probability/tensorflow_probability/python/util/deferred_tensor.py", line 293, in numpy
    return self._value().numpy()
AttributeError: 'Tensor' object has no attribute 'numpy'
```

tensorflow_probability/python/util/deferred_tensor_test.py
hartikainen: Thanks for the comments! I fixed the graph-mode part of the test. 

> I still just use print statements for debugging

Haha, that's definitely not a bad way to go 😄

hartikainen:
> What do you mean by "stop a debugger inside a test"?

@jvdillon What I meant by this was that it would be nice if bazel could stop the tests in the debugger using the `breakpoint()` (or `import ipdb; ipdb.set_trace()`). Not a big deal in this case as it was trivially debuggable with just print statements, but there's been a couple of cases in the past where I've tried debugging tensorflow tests and couldn't figure out how to let bazel to drop me in the debugger.

tensorflow_probability/python/util/deferred_tensor_test.py
csuter: Hm, not sure why it's getting a `Tensor` and not an `EagerTensor` (only the latter has a `.numpy()`)

tensorflow_probability/python/util/deferred_tensor_test.py
csuter: Just want to make sure, since it is a bit subtle in the UI, that the "here" in my comment was a link to this SO post, which may be helpful: https://stackoverflow.com/questions/51709590/can-i-use-python-debugger-in-bazel-test

tensorflow_probability/python/util/deferred_tensor_test.py
hartikainen: Isn't that expected because it's running in graph mode?

csuter: oh i was not looking at the x.numpy() below. i thought it was firing on the one inside the `if tf.executing_eagerly()` conditional. so is there a problem, or is it all good?

hartikainen: No, it looks good to me. Just wanted to make sure that `AttributeError` is what we'd expect here.

csuter: Ohh I see. We could  reasonably catch this and re-raise it with a more localized error message (eg, TypeError(?): .numpy() only works in eager mode)

tensorflow_probability/python/util/deferred_tensor_test.py
hartikainen: Oh yeah, I figured. I tried the `--run_under` examples (both with `--run_under=.../bin/ipdb3` and `--run_under=.../lib/python3.7/pdb.py`) from the SO post, but for some reason neither stops in to the debugger but it results in this weird behavior where the test passes no matter what I do to it (even if there's a syntax error and it fails without `--run_under`). Not sure if that's due to caching or something. I'll try to figure this out again next time I have to run the tests 🙂

tensorflow_probability/python/util/deferred_tensor_test.py
csuter: specifically,

```python
v = self._value()
if not hasattr(v, 'numpy'):
  raise TypeError("DeferredTensor.numpy() is only supported in eager execution mode")
return v.numpy()
```

This, rather than catching the exception, since it's best not to catch an re-raise (which can lose useful context if we accidentally catch something other than what we expected).

tensorflow_probability/python/util/deferred_tensor_test.py
csuter: Ah, sorry about that. 😬If you ever figure it out let us know and we can add some hints to the contributor docs!

tensorflow_probability/python/util/deferred_tensor_test.py
hartikainen: Sounds good! It now raises `TypeError`.

hartikainen: > specifically

Oh oops, I changed it before saw your comment. I changed it to:
```python
    try:
      return self._value().numpy()
    except AttributeError:
      if isinstance(self._value(), tf.Tensor):
        raise TypeError(
            "'{}.numpy()' only works in eager mode.".format(type(self).__name__))
      else:
        raise
```

Let me know if you prefer your version and I can change it.

csuter: Thanks, yeah, I prefer mine  as it's a bit simpler. I do appreciate your effort to catch the most specific error and re-raise anything else!

hartikainen: Cool, should be changed now. Thanks for the comments!

csuter:
d'oh. i think i borked the PR by trying to merge master.

Kristian, how is your git-fu? can you get your changes rebased onto the TFP mainline?

csuter:
Kristian, I think you can straighten this out as follows:

```bash
cd $(mktemp -d)
git clone git@github.com:hartikainen/probability
git remote add csuter https://github.com/csuter/probability
git fetch csuter
git checkout deferred-tensor-numpy-method
git reset --hard csuter/fix_hartikainen_pr
git push -f origin deferred-tensor-numpy-method
```

My copy of your PR branch (in csuter/fix_hartikainen_pr) has just your commits, but rebased on most recent master. Once you force push that to update your PR branch on GH, it should be updated here and ready to go. Sorry for the hassle; I've raised a question with some internal tooling folks, to see if we can avoid issues like this in the future (they stem from the fact that we have a different internal VCS and it gets confused about what repo state your PR is on top of, when it's not clearly rebased to the most recent master branch state).

hartikainen:
I just rebased and force-pushed the changed on top of the latest master 🙂

PR #738, opened 2020-01-21
Updated Tensorflow version to 2.1.0 

TheCaffeineDev:


srvasude:
Hi,
  What's the reason for this change? 

We generally try to support the latest TF release. So even if you have tf-nightly installed, the above should work. Hence, there shouldn't be a need to call out a specific TF version. Also we tend to generate this ipynb from the corresponding YAML file.

PR #747, opened 2020-01-24
Pixel cnn fixes 01

holli:
Training multichannel images (using log_probability) used real pixel values instead of supposed transformed values. This caused green/blue channels to be wrongly used with coeffs, which resulted weird red images.

There seemed to be no 3 channel image example in the docs, but below is one way to test:
```
# Build a small Pixel CNN++ model to train on colored FASHION_MNIST.

import tensorflow as tf
import tensorflow_datasets as tfds

tfd = tfp.distributions
tfk = tf.keras
tfkl = tf.keras.layers

# Load FASHION_MNIST from tensorflow_datasets
data = tfds.load('fashion_mnist')
train_data, test_data = data['train'], data['test']

# Make colorful dataset by shifting blue channel pixel to right
def image_preprocess(x):
    x = tf.cast(x['image'], tf.float32)
    x_b = tf.roll(x, shift=1, axis=1)
    x = tf.concat((x, x, x_b), -1)
    return (x,)  # (input, output) of the model

batch_size = 16
train_it = train_data.map(image_preprocess).batch(batch_size).shuffle(1000)

image_shape = (28, 28, 3)
# Define a Pixel CNN network
dist = tfd.PixelCNN(
  image_shape=image_shape,
  num_resnet=1,
  num_hierarchies=2,
  num_filters=32,
  num_logistic_mix=5,
  dropout_p=.3,
)

# Define the model input
image_input = tfkl.Input(shape=image_shape)

# Define the log likelihood for the loss fn
log_prob = dist.log_prob(image_input)

# Define the model
model = tfk.Model(inputs=image_input, outputs=log_prob)
model.add_loss(-tf.reduce_mean(log_prob))

# Compile and train the model
model.compile(
    optimizer=tfk.optimizers.Adam(.001),
    metrics=[])

model.fit(train_it, epochs=8, verbose=True)

# sample five images from the trained model
samples = dist.sample(2)


```

### DIFF
--- a/tensorflow_probability/python/distributions/pixel_cnn.py
+++ b/tensorflow_probability/python/distributions/pixel_cnn.py
@@ -122,9 +122,10 @@ class PixelCNN(distribution.Distribution):
   batch_size = 16
   train_it = train_data.map(image_preprocess).batch(batch_size).shuffle(1000)
 
+  image_shape = (28, 28, 1)
   # Define a Pixel CNN network
   dist = tfd.PixelCNN(
-      image_shape=(28, 28, 1),
+      image_shape=image_shape,
       num_resnet=1,
       num_hierarchies=2,
       num_filters=32,
@@ -133,7 +134,7 @@ class PixelCNN(distribution.Distribution):
   )
 
   # Define the model input
-  image_input = tfkl.Input(shape=input_shape)
+  image_input = tfkl.Input(shape=image_shape)
 
   # Define the log likelihood for the loss fn
   log_prob = dist.log_prob(image_input)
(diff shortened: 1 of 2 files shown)


holli:
@googlebot I signed it!

holli:
extra side note: seems like travis pylint didn't like the file because of previous code. That looks like the simple way of inheritance do do those tests so maybe disabling that linting error by adding ',assignment-from-none' to end of line https://github.com/tensorflow/probability/blob/master/testing/pylintrc#L54

emilyfertig:
Thank you! We had just noticed the red images also, glad to have this fixed.

PR #763, opened 2020-02-02
replace 'keep_dims' with 'keepdims' in quantiles.py

### LINKED ISSUE #557: tfp.stats.percentile uses keep_dims instead of keepdims
I think the rest of TF has converged on `keepdims` as the keyword argument name instead of `keep_dims`. It would be nice to switch over.

rushabh-v:
resolves #557

### DIFF
--- a/tensorflow_probability/python/stats/quantiles.py
+++ b/tensorflow_probability/python/stats/quantiles.py
@@ -22,6 +22,8 @@ from __future__ import print_function
 import numpy as np
 import tensorflow.compat.v2 as tf
 
+from tensorflow.python.util import deprecation
+
 from tensorflow_probability.python.internal import assert_util
 from tensorflow_probability.python.internal import distribution_util
 from tensorflow_probability.python.internal import dtype_util
--- a/tensorflow_probability/python/stats/quantiles.py
+++ b/tensorflow_probability/python/stats/quantiles.py
@@ -406,7 +406,7 @@ def percentile(x,
                q,
                axis=None,
                interpolation=None,
-               keep_dims=False,
+               keepdims=False,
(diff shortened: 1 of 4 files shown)


rushabh-v:
`continuous-integration/travis-ci/pr` files with `connection reset by peer` 
what should I do in this case?

tensorflow_probability/python/stats/quantiles.py
```diff
@@ -406,7 +406,7 @@ def percentile(x,
                q,
                axis=None,
                interpolation=None,
-               keep_dims=False,
+               keepdims=False,
```
brianwa84: I think we will need to keep the existing arg around as deprecated for 3 months.

Can you do something like this:
https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/distributions/mvn_diag.py#L138

Perhaps default it to `keep_dims=None`, and then on the first line, `keepdims = keepdims if keep_dims is None else keep_dims`

You can move it to the last position right before name, and add back the docstring but note it's deprecated in the docstring. Date should be +3m from now.

brianwa84: Also, this would only apply to public methods. Anything internal you can just change over.

rushabh-v:
SHARD 0 and SHARD 1 are still failing. It says `build did not complete successfully`.

brianwa84:
No worries, I can take a look at the failure when I merge. Comment above resolved now?

rushabh-v:
I have added deprecation with date 2020-05-03, Should I add a test for the deprecation?

brianwa84:
No test needed, thanks.

PR #767, opened 2020-02-04
 Wrong example for batch_interp_regular_nd_grid function

kovak1:
1. SyntaxError: positional argument follows keyword argument
2. Wrong x_ref_max value (1, instead of 10)
3. Missing 1 required positional argument: 'axis'


kovak1:
@googlebot I signed it!

SiegeLordEx:
Hi @kovak1, thanks for making this PR! Our documentation is automatically generated from docstrings, so that's where this should be fixed. Would you mind changing your commit to edit this file instead: https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/math/interpolation.py#L490 ?

Issue #764, opened 2020-02-03
`bfgs_minimize` errors with `@tf.function(experimental_compile=True)`


jeffpollock9:
Hi,

If an `initial_inverse_hessian_estimate` is supplied to `tfp.optimizer.bfgs_minimize` then some assert [operations are added](https://github.com/tensorflow/probability/blob/6fdaa6e60aa7a93a67055248ed5b81ce10b497d6/tensorflow_probability/python/optimizer/bfgs.py#L288). AFAIK this means that the function cannot work with `@tf.function(experimental_compile=True)` (see example below) and additionally the checks could be quite expensive for large problems as they involve a [cholesky decomposition](https://github.com/tensorflow/probability/blob/6fdaa6e60aa7a93a67055248ed5b81ce10b497d6/tensorflow_probability/python/optimizer/bfgs.py#L304).

I was wondering if there could be some way to skip these checks? Perhaps a new parameter like `tfp.optimizer.bfgs_minimize(..., validate_args=True)` in a similar fashion to the distributions? Happy to contribute this if it would be useful for anyone (it certainly would be for me).

Thanks!

```python
import tensorflow as tf
import tensorflow_probability as tfp

print(tfp.__version__)
### => '0.9.0'
print(tf.__version__)
### ==> '2.1.0'


def objective(x):
    return tf.reduce_sum(tf.math.squared_difference(x, [1.0, 10.0]))


def objective_value_and_grad(x):
    return tfp.math.value_and_gradient(objective, x)


@tf.function(autograph=False, experimental_compile=True)
def fit_works():
    return tfp.optimizer.bfgs_minimize(
        objective_value_and_grad, initial_position=tf.zeros(2),
    )


@tf.function(autograph=False, experimental_compile=True)
def fit_doesnt_work():
    return tfp.optimizer.bfgs_minimize(
        objective_value_and_grad,
        initial_position=tf.zeros(2),
        initial_inverse_hessian_estimate=tf.eye(2),
    )


optim = fit_works()

print(f"value: {optim.objective_value}, position: {optim.position}")
### => value: 0.0, position: [ 1. 10.]

fit_doesnt_work()
### =>
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jeff/workspace/misc/bfgs_compile.py", line 33, in <module>
    optim = fit_doesnt_work()
[...]

PR #771, opened 2020-02-05
Validate optimizer args

jeffpollock9:
As per #764 this adds `validate_args=True` (so by default the previous behaviour is maintained) to `bfgs_minimize` to allow the option of experimenting with `@tf.function(experimental_compile=True)`.

@csuter tagging you incase this is at all useful for you.

Thanks!


SiegeLordEx:
Looks great, thanks! We'll run this through an internal review and get this merged once that passes.

Issue #769, opened 2020-02-05
layers.VariationalGaussianProcess skips layer's 'name' property


alexander-e1off:
Hello,

it seems `layers.VariationalGaussianProcess` does not pass `name` property to the base class initializer.
It is accepted [here](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/layers/distribution_layer.py#L1814) but not passed to the base class initializer [here](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/layers/distribution_layer.py#L1854)

As a result layer's name remains default.

brianwa84:
Sounds like an easy fix, feel free to send a PR.

alexander-e1off:
Fixed in #772

PR #772, opened 2020-02-06
Pass kwargs to VariationalGaussianProcess initializer

alexander-e1off:
Addresses #769. 
For `VariationalGaussianProcess` class replace `name` argument with `
**kwargs` to allow passing  additional keyword arguments passed to `tf.keras.Layer` (including `name`).

### DIFF
--- a/tensorflow_probability/python/layers/distribution_layer.py
+++ b/tensorflow_probability/python/layers/distribution_layer.py
@@ -47,7 +47,7 @@ from tensorflow_probability.python.distributions import transformed_distribution
 from tensorflow_probability.python.distributions import variational_gaussian_process as variational_gaussian_process_lib
 from tensorflow_probability.python.internal import distribution_util as dist_util
 from tensorflow_probability.python.layers.internal import distribution_tensor_coercible as dtc
-from tensorflow_probability.python.layers.internal import tensor_tuple as tens_tuple
+from tensorflow_probability.python.layers.internal import tensor_tuple as tensor_tuple
 from tensorflow.python.keras.utils import tf_utils as keras_tf_utils  # pylint: disable=g-direct-tensorflow-import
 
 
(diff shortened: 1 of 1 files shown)


alexander-e1off:
@googlebot I signed it!

tensorflow_probability/python/layers/distribution_layer.py
```diff
@@ -47,7 +47,7 @@
 from tensorflow_probability.python.distributions import variational_gaussian_process as variational_gaussian_process_lib
 from tensorflow_probability.python.internal import distribution_util as dist_util
 from tensorflow_probability.python.layers.internal import distribution_tensor_coercible as dtc
-from tensorflow_probability.python.layers.internal import tensor_tuple as tensor_tuple
+from tensorflow_probability.python.layers.internal import tensor_tuple as tens_tuple
```
csuter: please leave this as it was

tensorflow_probability/python/layers/distribution_layer.py
```diff
@@ -1811,7 +1811,7 @@ def __init__(
       mean_fn=None,
       jitter=1e-6,
       convert_to_tensor_fn=tfd.Distribution.sample,
-      name=None):
+      **kwargs):
```
csuter: We generally avoid using catchall **kwargs this way in TFP. Please instead keep the explicit `name` argument here, and propagate it explicitly to the later call to construct the distribution instance.

The reason we avoid kwargs is because it makes things like docstrings and tab-completion generally less useful.

tensorflow_probability/python/layers/distribution_layer.py
alexander-e1off: Rolled back this change, although it caused pylint error.

tensorflow_probability/python/layers/distribution_layer.py
alexander-e1off: Explicitly passed `name` argument

alexander-e1off:
@csuter thanks for your review

PR #777, opened 2020-02-08
Fix code examples in sample and dual_averaging_step

Sayam753:
Fixed code examples in - 
 - [mcmc/sample.py](https://github.com/tensorflow/probability/blob/v0.9.0/tensorflow_probability/python/mcmc/sample.py#L296-L299)
 - [mcmc/dual_averaging_step_size_adaptation.py](https://github.com/tensorflow/probability/blob/v0.9.0/tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py#L169)


Sayam753:
@googlebot I signed it!

PR #781, opened 2020-02-10
Fixed the example

liamlio:
Example didn't run previously with the higher dimension and non-matrix input to .log_prob()


ColCarroll:
Thanks for spotting this! 

What do you think about using `tf.zeros(dims)` instead, to more clearly show the intention?

srvasude:
Thanks! Could you make the change to  the corresponding .py file? We generate the markdown from the py files.

liamlio:
@srvasude Created a new request https://github.com/tensorflow/probability/pull/783
And good idea @ColCarroll! I changed it to your suggestion

PR #790, opened 2020-02-13
Update math module reference in API documentation

gameshamilton:


srvasude:
Hi,
  Thanks for the change! Do you mind making it instead to the corresponding python file? We generate our markdown files from the python files.

srvasude:
Can you remove the markdown change?

gameshamilton:
It seems there is still an ``` change on the .md file. Should I address that? Apologies.

srvasude:
Yes please. Thanks!

Issue #791, opened 2020-02-15
Add bijectors: FillDiagonal and FillScaleDiagonal


fdtomasi:
While I'd guess the use for it it would be minimal, I found myself implementing a `FillDiagonal` and a `FillScaleDiagonal` bijectors, in particular to deal with the `variational_inducing_observations_scale` parameter of `tfd.VariationalGaussianProcess`, that reads as

```
variational_inducing_observations_scale: `float` `Tensor`; the scale
        matrix of the (full-rank Gaussian) variational posterior over function
        values at the inducing points, conditional on observed data. Shape has
        the form `[b1, ..., bB, e2, e2]`, where `b1, ..., bB` is broadcast
        compatible with other parameters and `e2` is the number of inducing
        points.
```

I like the general framework of `VariationalGaussianProcess`, even though it would be good to be able to specify a `variational_inducing_observations_scale_diag` parameter to avoid materialising a scale matrix while only interested in the diagonal. Side note: I briefly tried to look into the code of VariationalGaussianProcess but I was not sure how to deal with `variational_gaussian_process._solve_cholesky_factored_system(cholesky_factor, rhs, name=None)` when `rhs` is a vector representing a diagonal matrix.

While I guess this is probably a workaround to deal with the `VariationalGaussianProcess` case, it is probably useful for other use cases, hence if it makes sense I can submit a PR with the FillDiagonal and FillScaleDiagonal bijectors.
The bijectors are modifications of `FillTriangular` and `FillScaleTril`. 

```python
class FillDiagonal(bijector.Bijector):
    """Transforms vectors to diagonal matrices.

  Given input with shape `batch_shape + [d]`, produces output with
  shape `batch_shape + [d, d]`.
  [...]
```
```python
  b = tfb.FillDiagonal()
  b.forward([1, 2, 3])
  # ==> [[1, 0, 0],
  #      [0, 2, 0],
  #      [0, 0, 3]]
  ```

and 

```python
class FillScaleDiagonal(chain.Chain):
    """Transforms unconstrained vectors to Diagonal matrices with positive diagonal.
  This is implemented as a simple `tfb.Chain` of `tfb.FillDiagonal`
  followed by `tfb.TransformDiagonal`, and provided mostly as a
  convenience. The default setup is somewhat opinionated, using a
  Softplus transformation followed by a small shift (`1e-5`) which
  attempts to avoid numerical issues from zeros on the diagonal.
  [...]
  ```

```python
[...]

PR #792, opened 2020-02-15
Add bijectors: FillDiagonal and FillScaleDiagonal

fdtomasi:
Following #791, this implements the `FillDiagonal` and `FillScaleDiagonal` bijectors.
I also added some tests, similar to those for `FillTriangular` and `FillScaleTriL`.

### DIFF
--- a/tensorflow_probability/python/bijectors/fill_scale_diagonal.py
+++ b/tensorflow_probability/python/bijectors/fill_scale_diagonal.py
@@ -1,0 +1,99 @@
+"""FillScaleDiagonal bijector."""
+
+from __future__ import absolute_import, division, print_function
+
+import tensorflow.compat.v2 as tf
+from tensorflow_probability.python.bijectors import chain
+from tensorflow_probability.python.bijectors import fill_diagonal
+from tensorflow_probability.python.bijectors import shift
+from tensorflow_probability.python.bijectors import softplus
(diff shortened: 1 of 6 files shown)


tensorflow_probability/python/bijectors/fill_scale_diagonal.py
```diff
@@ -0,0 +1,117 @@
...
+        dtype = dtype_util.common_dtype([diag_bijector, diag_shift],
+                                        tf.float32)
+        diag_shift = tensor_util.convert_nonref_to_tensor(
+            diag_shift, name="diag_shift", dtype=dtype)
+        diag_bijector = chain.Chain(
+            [shift.Shift(shift=diag_shift), diag_bijector])
+
+      super(FillScaleDiagonal, self).__init__(
+          [
+              transform_diagonal.TransformDiagonal(
```
brianwa84, line 111: I think for the FillScale we would want to flip the order so that we apply the transform prior to filling the diagonal. Otherwise it's `diag(transform(diag_part(diag(x))))`, which is wasteful.

brianwa84:
Sorry this has languished so long. Looks pretty good, small tweak requested.

fdtomasi:
Fixed! I Now applied `diag_bijector` before Filldiagonal. I also removed the dependency from `transform_diagonal` module.

Issue #785, opened 2020-02-11
example missing variable declaration


bourov:
Tried to run [Bayesian_Gaussian_Mixture_Model.ipynb](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/examples/jupyter_notebooks/Bayesian_Gaussian_Mixture_Model.ipynb)

get error in cell below, 
`NameError: name 'step_sizes' is not defined
`

cell in question:
```@tf.function(autograph=False)
def sample():
  return tfp.mcmc.sample_chain(
    num_results=2000,
    num_burnin_steps=500,
    current_state=initial_state,
    kernel=tfp.mcmc.SimpleStepSizeAdaptation(
        tfp.mcmc.TransformedTransitionKernel(
            inner_kernel=tfp.mcmc.HamiltonianMonteCarlo(
                target_log_prob_fn=unnormalized_posterior_log_prob,
                 step_size=step_sizes,
                 num_leapfrog_steps=5),
            bijector=unconstraining_bijectors),
         num_adaptation_steps=400),
    trace_fn=lambda _, pkr: pkr.inner_results.inner_results.is_accepted)

[mix_probs, loc, chol_precision], is_accepted = sample()```

manda-creator:
I believe I can fix this and would like to try it.

csuter:
@manda-creator go for it!

PR #805, opened 2020-02-20
Updating variable declaration

manda-creator:
* Issue Reference: #785
Fixed error (NameError: name 'step_sizes' is not defined) you get when you run def sample().


csuter:
Thanks! We'll pull this in shortly.

PR #806, opened 2020-02-21
Removed unnecessary 'output' variable in _test_grad_finite()

Iqrar99:
Hello,
The assignment to `output` is unnecessary as it is redefined in line 193 before the value is used. This makes the first assignment useless.

https://github.com/tensorflow/probability/blob/ab414b238654db5fe7749a6cb5ad2e0855e908e4/tensorflow_probability/python/internal/special_math_test.py#L186-L197

This is a common weakness enumeration based on https://cwe.mitre.org/data/definitions/563.html


Iqrar99:
Can you review this PR, please? @jvdillon

Issue #739, opened 2020-01-22
Missing diagonal shift in VariationalKernel


mrksr:
Hi all,

The `VariationalGaussianProcess` was reformulated using a `VariationalKernel` similar to the `SchurComplement`-Kernel. This kernel is defined with respect to the inducing points `Z` and contains a cholesky decomposition of the `kernel.matrix(Z, Z)` matrix. It is computed here:
https://github.com/tensorflow/probability/blob/e0346910ecca45f6ee83a7deccb3fe7d690e296a/tensorflow_probability/python/distributions/variational_gaussian_process.py#L127

This decomposition does not add a diagonal jitter, which sometimes leads to the decomposition failing, for example for initializations which happen to be suboptimal.
Could the jitter be added using `_add_diagonal_shift` or is there a specific reason not to?

Thanks!

brianwa84:
Chris: thoughts?

Markus: consider filing an issue on github to track the request?

Brian Patton | Software Engineer | bjp@google.com

csuter:
@brianwa84 this is a github issue :)

@mrksr while adding a jitter here might allow the cholesky to proceed, I worry a bit about sprinkling jitter in too many places, especially in a non-user-configurable manner (and I'd also be concerned about demanding separate jitter parameters for each of the various places we take cholesky's).

I wonder if a viable workaround would be to try to ensure that the inducing points aren't too close together (in the kernel-induced metric). Often in my experience, ensuring they're initialized sufficiently far apart can help here. I.e., one might consider using QMC (e.g. [Halton sequence](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/mcmc/sample_halton_sequence.py)) to initialize the inducing points, rather than something like uniform random over some hypercube.

csuter:
As an addendum: you could consider adding a regularization term to the overall loss that discourages inducing points from getting too close.  Whereas adding diagonal jitter amounts to a change in the model (hence my trepidation), such a regularization leaves the model fixed but changes the optimization procedure. It's also an easier lever for the user to wield freely, without any change to the VGP APIs.

Thoughts welcome!

brianwa84:
Lol woops I thought it was on the tfprobability mailing list. :-)

mrksr:
I have now found some time to play with the model and generate a small example. Thank you for the pointer to the halton sequence, that's definitely a more principled way to initialize inducing points.

# An example model
[...]

PR #813, opened 2020-02-24
Add diagonal shift in VariationalKernel

mrksr:
# Motivation
See #739 for previous discussion.

The `VariationalGaussianProcess` was reformulated using a `VariationalKernel` similar to the `SchurComplement`-Kernel. This kernel is defined with respect to the inducing points `Z` and contains a cholesky decomposition of the `kernel.matrix(Z, Z)` matrix. It is computed here:
https://github.com/tensorflow/probability/blob/e0346910ecca45f6ee83a7deccb3fe7d690e296a/tensorflow_probability/python/distributions/variational_gaussian_process.py#L127

# An example model
Let's take a look at this relatively simple example using `tfp.layers` for interop with keras. This is meant as a relatively real-worldy example, I'm aware it is not quite minimal.
```python
import numpy as np
import tensorflow as tf


class RBFLayer(tf.keras.layers.Layer):
    def __init__(self, feature_dimensions, **kwargs):
        super().__init__(**kwargs)     
        
        softplus_inverse_of_one = tfp.bijectors.Softplus().inverse(1.).numpy()
        
        self.amplitude = self.add_weight(
            name='amplitude',
            shape=[1],
            initializer=tf.keras.initializers.Constant(softplus_inverse_of_one),            
        )
        self.length_scale = self.add_weight(
            name='length_scale',
            shape=feature_dimensions,
            initializer=tf.keras.initializers.Constant(softplus_inverse_of_one),
        )
    
    @property
    def kernel(self):
        return tfp.math.psd_kernels.ExponentiatedQuadratic(
                amplitude=tf.math.softplus(self.amplitude),
                length_scale=tf.math.softplus(self.length_scale),
            )

    
num_inducing_points = 50
halton_sequence = tfp.mcmc.sample_halton_sequence(1, num_inducing_points).numpy()
halton_sequence = (-2) + (2 - (-2)) * halton_sequence

tf.keras.backend.set_floatx('float64')
model = tf.keras.Sequential([
    tfp.layers.VariationalGaussianProcess(
        num_inducing_points=num_inducing_points,
        kernel_provider=RBFLayer([1]),
        inducing_index_points_initializer=tf.keras.initializers.Constant(halton_sequence),
    ), 
])

model(np.array([[0.]]))
# Cholesky decomposition unsuccessful most of the time.
```
In this example, we formulate a 1D GP both on the input and output side using an RBF kernel. Keras assumes roughly standardized data and we initialize the RBF parameters with `1`. The lengthscale is slightly on the longer side, but I would call it _the_ standard initialization for GPs. It's a bit hard for me to reproduce exactly when the initialization is bad, but the problem definitely gets more pronounced for low dimensions. Not very surprising, after all, as there is more chance of inducing points being close together. Also note that 50 inducing points is not very much. It would be good if 500 also worked.

# Changes
This PR adds a jitter to the cholesky decomposition in the `_VariationalKernel`. We re-use the jitter of the `VariationalGaussianProcess` model which is passed through in a way analogous to the `VariationalGaussianProcess` model itself. Adding the default jitter `1e-6` for `tf.float64` fixes the problems described above and also works for 500 inducing points. Feedback welcome!

### DIFF
--- a/tensorflow_probability/python/distributions/variational_gaussian_process.py
+++ b/tensorflow_probability/python/distributions/variational_gaussian_process.py
@@ -90,6 +90,7 @@ class _VariationalKernel(tfpk.PositiveSemidefiniteKernel):
                base_kernel,
                inducing_index_points,
                variational_scale,
+               jitter=1e-6,
                name='VariationalKernel'):
     """Construct a _VariationalKernel instance.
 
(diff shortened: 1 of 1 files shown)


mrksr:
@googlebot I signed it!

mrksr:
@csuter Sorry to keep bugging you. Any thoughts?

PR #817, opened 2020-02-27
In Linear_Mixed_Effects_Model notebook, add more clarifications on how to attach tf.Module submodules to JointDistributionSequential

ksachdeva:
When I first read this notebook, the statement `model._to_track = self` gave me an impression that `_to_track` must be a private property. However, on further reading and understanding of how tf.Module works (i.e. it does reflection on properties attached to it and if the property is of type tf.Module then it includes its variables in the graph) cleared the picture.

That said, I think usage of '_to_track' could mislead others (as it did to me) because of two reasons -
* First it gives an impression that a private property is being used because of underscore in front of it
* and then usage of 'track' in the name further solidifies ones reasoning that it must be a private property meant for tracking. Reason it could become problem is when some one wants to attach two tf.Modules to JointDistributionSequential, then they would start to wonder if they should use list etc. I hope you see how naming of a variable opens up confusion for someone who may not be an expert in the underlying mechanism. 

This revision simply add more comments on this aspect and even clarification on the name of the property being used to attach the tf.Module.  

> A side comment that has nothing to do with this revision - IMHO, overall I believe the design of tf.Module to allow someone to do this (i.e reliance on attaching properties to object) is not so great even though python (the dynamic language) allows it. It may be a good idea for TFP team to expose a proper construct/api to add submodules to help mitigate the impact of the design choice made by tf.Module.


tensorflow_probability/examples/jupyter_notebooks/Linear_Mixed_Effects_Models.ipynb
```diff
@@ -590,11 +588,17 @@
...
     "\n",
     "    # To enable tracking of the trainable variables via the created distribution,\n",
-    "    # we attach a reference to `self`. Since all TFP objects sub-class\n",
-    "    # `tf.Module`, this means that the following is possible:\n",
-    "    # LinearMixedEffectModel()(features_train).trainable_variables\n",
-    "    # ==> tuple of all tf.Variables created by LinearMixedEffectModel.\n",
-    "    model._to_track = self\n",
+    "    # we attach a reference to `self`. Since all TFP objects (e.g. JointDistributionSequential in this case )\n",
+    "    # sub-class `tf.Module`, this means that if attach an object of \n",
+    "    # type tf.Module (i.e. LinearMixedEffectModel in this case) to it then the variables (i.e tf.Variables created\n",
```
ColCarroll, line 593: change "it" to something like "the return value", and no need to use the parenthetical here. What do you think of something like 

```
# Since all TFP objects (e.g., `JointDistributionSequential` here) sub-class 
# `tf.Module`, this means that if we attach an object of type `tf.Module` 
# (the `LinearMixedEffectModel` in this case) to the return value, then the 
# `tf.Variables` created by `LinearMixedEffectModel` will be automatically 
# tracked. Practically this next line allows:
#
# LinearMixedEffectModel()(features_train).trainable_variables
# ==> tuple of all tf.Variables created by LinearMixedEffectModel
```

tensorflow_probability/examples/jupyter_notebooks/Linear_Mixed_Effects_Models.ipynb
```diff
@@ -625,14 +629,16 @@
...
-      "tags": []
-     },
+     "execution_count": 11,
+     "metadata": {},
      "output_type": "execute_result"
     }
    ],
    "source": [
+    "# lmm_train is of type JointDistributionSequential which implictly inherits from tf.Module\n",
+    "# and since in above cell we attached the LinearMixedEffectModel Module to it you would\n",
```
ColCarroll, line 639: missing: "in _the_ above cell"

also `back_ticks` around `LinearMixedEffectModel` and `tf.Module` (and some inline code below)

tensorflow_probability/examples/jupyter_notebooks/Linear_Mixed_Effects_Models.ipynb
```diff
@@ -590,11 +588,17 @@
     "    ])\n",
     "\n",
     "    # To enable tracking of the trainable variables via the created distribution,\n",
-    "    # we attach a reference to `self`. Since all TFP objects sub-class\n",
-    "    # `tf.Module`, this means that the following is possible:\n",
-    "    # LinearMixedEffectModel()(features_train).trainable_variables\n",
-    "    # ==> tuple of all tf.Variables created by LinearMixedEffectModel.\n",
-    "    model._to_track = self\n",
+    "    # we attach a reference to `self`. Since all TFP objects (e.g. JointDistributionSequential in this case )\n",
+    "    # sub-class `tf.Module`, this means that if attach an object of \n",
```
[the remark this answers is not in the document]
ksachdeva: ok

tensorflow_probability/examples/jupyter_notebooks/Linear_Mixed_Effects_Models.ipynb
ksachdeva: ok. looks better.

tensorflow_probability/examples/jupyter_notebooks/Linear_Mixed_Effects_Models.ipynb
ksachdeva: ok

ksachdeva:
Hi @ColCarroll, have incorporated your feedback. Please review.

Regards

tensorflow_probability/examples/jupyter_notebooks/Linear_Mixed_Effects_Models.ipynb
```diff
@@ -590,11 +588,17 @@
...
-    "    # `tf.Module`, this means that the following is possible:\n",
-    "    # LinearMixedEffectModel()(features_train).trainable_variables\n",
-    "    # ==> tuple of all tf.Variables created by LinearMixedEffectModel.\n",
-    "    model._to_track = self\n",
+    "    # we attach a reference to `self`. Since all TFP objects (e.g. `JointDistributionSequential` in this case )\n",
+    "    # sub-class `tf.Module`, this means that if we attach an object of \n",
+    "    # type tf.Module (i.e. `LinearMixedEffectModel` in this case) to the return value then the variables created\n",
+    "    # by `LinearMixedEffectModel` will be automatically tracked\n",
+    "    #        \n",
+    "    # In the below statement, we have used `lmm_model` as the name of the property but we could have\n",
```
ColCarroll, line 596: I still think that including the example that was here previously is helpful, and can be called back to in the second comment you added.

I would also remove the part of this comment after "In the below statement" -- I think your renaming makes it clear this is not a private flag, and you already mentioned that `tf.Module` tracks trainable variables above.

tensorflow_probability/examples/jupyter_notebooks/Linear_Mixed_Effects_Models.ipynb
```diff
@@ -625,14 +629,16 @@
...
-     "metadata": {
-      "tags": []
-     },
+     "execution_count": 11,
+     "metadata": {},
      "output_type": "execute_result"
     }
    ],
    "source": [
+    "# lmm_train is of type `JointDistributionSequential` which implictly inherits from `tf.Module`\n",
```
ColCarroll, line 638: Maybe just  "We can see the trainable variables because of the reference to `self` we included earlier."?

tensorflow_probability/examples/jupyter_notebooks/Linear_Mixed_Effects_Models.ipynb
ksachdeva: If you come from other languages & domains, it is very unnatural to attach a property to an object at runtime and hence I believe that having an explicit explanation of what is happening and the fact that it could have been named anything else is appropriate here. 

As I say in my extended comment in the pull request (which will not be part of the notebook itself) that one should not be dynamically attaching properties in the first place.  It is a bad design of `tf.Module` that allows it (or rather mandates it) and hence at a minimum, we should explain it.

ColCarroll: Two things:
1. I was asking about keeping in the text:
```
# Practically this next line allows:
#
# LinearMixedEffectModel()(features_train).trainable_variables
# ==> tuple of all tf.Variables created by LinearMixedEffectModel
```

2. If you feel strongly about it, I won't insist (improving examples is great!). I don't think an example notebook is the right place to teach how Python classes work, though.

ksachdeva:
Going to add that example line. This example was not there in the original code and was added after my pull request (I think !)

Since merging is challenging with Jupyter notebook how about - I create a different pull request and incorporate all your feedback in that and then cancel this one. This way there is no issue of the merge.

PR #818, opened 2020-02-28
Fix typos in documentation for smoothing

mjul:
Fix typos by adding the missing 1


jeffpollock9:
Hi @mjul. AFAIK the markdown files you've edited are automatically generated from the docstrings in the corresponding .py files - so better to make the changes there instead.

mjul:
@jeffpollock9 Thanks for the advice, I have updated the PR accordingly.

srvasude:
Hi,
  Would you please remove the changes to the .md files? PR looks good otherwise.

Also, we will be removing the .md files soon, since we have nightly docs generated from the docstrings now.

mjul:
OK, @srvasude, I have rebased and dropped the commit to the .md files as requested.

Issue #834, opened 2020-03-07
doc string error in weibull_cdf.py


bourov:
in class [tensorflow_probability/python/bijectors/weibull_cdf.py](https://github.com/tensorflow/probability/blob/425395da0472b91400655ba3aec8f69cf0249673/tensorflow_probability/python/bijectors/weibull_cdf.py)

doc text is
```python 
"""Compute `Y = g(X) = 1 - exp((-X / scale) ** concentration), X >= 0`.
```

while actual calculation is (note `(-x)**power` vs `-(x**power)`)
```python
  def _forward(self, x):
    with tf.control_dependencies(self._maybe_assert_valid_x(x)):
      return -tf.math.expm1(-((x / self.scale) ** self.concentration))
```

if no-one objects, I can try to fix the doc string in several places in `weibull_cdf.py `

srvasude:
Hi, 
   Feel free to send a PR for this! Thanks for the catch!

bourov:
> Hi,
> Feel free to send a PR for this! Thanks for the catch!

Will do!

PR #835, opened 2020-03-07
update doc string formula to match actual computation

bourov:
fixed doc string in `weibull_cdf.py` per [issue 834](https://github.com/tensorflow/probability/issues/834) to match actual calculation


bourov:
@googlebot I signed it!

srvasude:
@csuter Hey Chris, 
 I pulled this request in and submitted above, but it seems to have not auto-closed this. Is the recommendation that I manually close this (I'm assuming that the PR is not synced to latest which is why it generated the merge conflict).

csuter:
Yeah let's just manually close. Looks like the commit is in and properly attributed to @bourov.

Issue #448, opened 2019-06-03
TFP and TF2.0 incompatibility


awav:
After using TFP with TF2.0, I've realized that I'm spending a lot of time on understanding why it doesn't work rather than making any progress. The TFP relies too much on TF1.0, even not realizing it. Some conceptions of the TF2.0 and the TFP are just incompatible, e.g. `tf.Module`.

```python
class Model(tf.Module):
    def __init__(self):
        pass  # Create variables, set the prior and etc.

data = ... setup data
model = Model()

def run_chain():
    def posterior_log_prob_fn(*parameters):
        model.assign_parameters(*parameters)
        log_prob = -model.neg_log_marginal_likelihood(data)
        return log_prob
    
    # state_gradients_are_stopped=True
    variables = model.trainable_variables
    hmc = mcmc.HamiltonianMonteCarlo(target_log_prob_fn=posterior_log_prob_fn,
                                     num_leapfrog_steps=2,
                                     step_size=1.0,
                                     state_gradients_are_stopped=True)
    adaptive_hmc = mcmc.SimpleStepSizeAdaptation(hmc, num_adaptation_steps=2)
    return mcmc.sample_chain(num_results=2,
                             num_burnin_steps=0,
                             current_state=variables,
                             kernel=adaptive_hmc)
```

The HMC will not be able to compute the gradients as there are no links in a graph between states and the model (with or without tf.function).

Other examples you can find here: https://github.com/tensorflow/probability/issues/333, https://github.com/tensorflow/probability/issues/348, https://github.com/tensorflow/probability/issues/47

There is no easy fix for that and hope I'm wrong), as the roots of these problems are in TF2.0 itself. I've created another github issue at TF2.0: https://github.com/tensorflow/tensorflow/issues/29367, describing how and what makes TFP framework less effective.

csuter:
Hi @awav, as mentioned just now on the parallel TF issue, we have a candidate solution for this, which we're working on getting checked in now. It's only a template, and we'll have to do a substantial refactoring of all bijectors and distributions to bring them up to date with the new approach. Stay tuned.

awav:
@csuter, fantastic! Do you have any time estimates when it will be finished?

[...]

PR #841, opened 2020-03-14
Remove deprecated affine_scalar

giovp:
Replace `tfb.AffineScalar` with `tfb.Shift(shift=...)(tfb.Scale(scale=...))` from `tfb.MaskedAutoregressiveFlow` as suggested in #448 .
Running `pytest masked_autoregressive_test.py` returns a lot of tests failing but I might be doing it wrong.  
If nonetheless is useful I will go on an replace it also from `tfb.RealNVP`


giovp:
@googlebot I signed it!

giovp:
@googlebot I fixed it.

Issue #839, opened 2020-03-13
bijector_fn argument does not work for bijector examples


giovp:
```python
print('tensorflow: ', tf.__version__)
print('tensorflow-probability: ', tfp.__version__)
```
```
tensorflow:  2.1.0. 
tensorflow-probability:  0.9.0
```
I want to make a flow with an autoregressive network. I understand  from #448 #683 that I would have to pass the values of shift and log_scale to the bijector_fn, and use `tfb.Shift` and `tfb.Scale` as suggested because `tfb.AffineScalar` is deprecated. This is they way I though it works (but clearly does not)...

```python
NN = tfb.AutoregressiveNetwork(
    params=2,
    event_shape=[10],
    hidden_units=[20,20],
)

shift, log_scale = tf.unstack(NN(tf.random.normal((30,10))), axis=-1)
bij = tfb.Shift(shift)(tfb.Scale(log_scale))

flow = tfb.MaskedAutoregressiveFlow(
    bijector_fn=bij,
    validate_args = True,
)
```

<details>
<summary>Error</summary>
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-33-a69ba0ab2a1d> in <module>
     12     validate_args = True,
     13 )
---> 14 flow(tf.random.normal((30,10)))

/tungstenfs/groups/gliberal/Users/pallgiov/prg/miniconda3/envs/tfp_scanpy/lib/python3.6/site-packages/tensorflow_probability/python/bijectors/bijector.py in __call__(self, value, name, **kwargs)
    863       return chain.Chain([self, value], name=name, **kwargs)
    864 
--> 865     return self.forward(value, name=name or 'forward', **kwargs)
    866 
    867   def _forward_event_shape_tensor(self, input_shape):

/tungstenfs/groups/gliberal/Users/pallgiov/prg/miniconda3/envs/tfp_scanpy/lib/python3.6/site-packages/tensorflow_probability/python/bijectors/bijector.py in forward(self, x, name, **kwargs)
   1001       NotImplementedError: if `_forward` is not implemented.
   1002     """
-> 1003     return self._call_forward(x, name, **kwargs)
   1004 
   1005   @classmethod

/tungstenfs/groups/gliberal/Users/pallgiov/prg/miniconda3/envs/tfp_scanpy/lib/python3.6/site-packages/tensorflow_probability/python/bijectors/bijector.py in _call_forward(self, x, name, **kwargs)
[...]

PR #842, opened 2020-03-14
Remove deprecated affine_scalar

giovp:
Replace deprecated `tfb.AffineScalar` with `tfb.Shift(shift=...)(tfb.Scale(scale=...))` for `tfb.MaskedAutoregressiveFlow` as suggested in #448 . 
Pytest is failing but I can't figure out why.

Sorry for the previous PR, had an issue with mismatching users w/ google CLA...


tensorflow_probability/python/bijectors/masked_autoregressive.py
```diff
@@ -25,7 +25,7 @@
 import tensorflow.compat.v1 as tf1
 import tensorflow.compat.v2 as tf
 
-from tensorflow_probability.python.bijectors import affine_scalar
+from tensorflow_probability.python.bijectors import shift, scale
```
srvasude: Note that shift is also a name used below, which is what is making Travis fail. 
maybe:
`from tensorflow_probability.python.bijectors import shift as shift_bijector`
Also please put a single import per line.

giovp: Sorry for late reply. Seems like tests are still failing.

giovp:
Hi @srvasude, I removed `affine_scalar` also from real_nvp but tests are failing there as well. 
Just want to mention that using the `Shift` and `Scale` bijectors as per #839 still seems to not have trainable variables. Any suggestion on this?
Thanks a lot in advance.

srvasude:
Hi apologies for the delay on this,
   I added an option for `Scale` to take in `log_scale` now, so please try that (hoping that will fix the tests).

giovp:
Eventually I re implemented by subclassing the bijector class, but this I believe should do the job. I see that it was changed already in master, so this PR can be closed I guess, right?

Thank you for having a look anyway!

brianwa84:
Can you merge masked_autoregressive and rebase? The change to real_nvp is still relevant, if you want to go ahead w/ it.

brianwa84:
You could also update masked_autoregressive to use the `log_scale` parameter to `tfb.Scale`, if you don't mind.

giovp:
Hi @brianwa84 , added log_scale to both autoregressive and realnvp but some tests are still failing :( sorry for that but not sure how to address them.

Thank you for looking into this.

brianwa84 (approved):
will get this pulled in. thanks!

PR #843, opened 2020-03-15
Implement loglogistic distribution

npfp:
Formula from the
Handbook of Differential Entropy, J.V. Michalowicsz, J.M. Nichols, F. Bucholtz
p. 113, 114

### DIFF
--- a/tensorflow_probability/python/distributions/loglogistic.py
+++ b/tensorflow_probability/python/distributions/loglogistic.py
@@ -130,17 +130,16 @@ class LogLogistic(transformed_distribution.TransformedDistribution):
       return (tf.math.log(x) - self.loc) / self.scale
 
   def _log_prob(self, x):
-    return (-tf.math.log(self.scale) - self.loc
+    return (-tf.math.log(self.scale)
+            - self.loc
             + (1. - self.scale) * self._log_z(x)
-            - 2 * tf.math.log1p(
-              tf.math.exp(self._log_z(x)))
-            )
+            - 2 * tf.math.softplus(self._log_z(x)))
 
   def _log_cdf(self, x):
-    return -tf.math.log1p(tf.math.exp(-self._log_z(x)))
+    return -tf.math.softplus(-self._log_z(x))
 
   def _log_survival_function(self, x):
-    return - self._log_z(x) + self._log_cdf(x)
+    return self._log_cdf(x) - self._log_z(x)
 
   def _sample_control_dependencies(self, x):
     assertions = []
--- a/tensorflow_probability/python/distributions/loglogistic.py
+++ b/tensorflow_probability/python/distributions/loglogistic.py
@@ -99,19 +99,43 @@ class LogLogistic(transformed_distribution.TransformedDistribution):
     return self._concentration
 
   def _mean(self):
-    b = 1. / self.concentration
-    mean = self.scale / sinc(b)
-    return tf.where(tf.greater(self.concentration, 1.), mean, np.nan)
+    concentration = tf.convert_to_tensor(self.concentration)
+    with tf.control_dependencies(
+        [] if self.allow_nan_stats else [  # pylint: disable=g-long-ternary
+          assert_util.assert_less(
+            tf.ones([], dtype=self.dtype),
+            concentration,
+            message='Mean undefined for concentration <= 1.'),
+        ]):
+      b = 1. / self.concentration
+      mean = self.scale / sinc(b)
+      return tf.where(self.concentration > 1., mean, np.nan)
 
   def _variance(self):
-    b = 1. / self.concentration
-    variance = self.scale ** 2 * (1. / sinc(2*b) - 1. / sinc(b)**2)
-    return tf.where(tf.greater(self.concentration, 2.), variance, np.nan)
+    concentration = tf.convert_to_tensor(self.concentration)
+    with tf.control_dependencies(
+        [] if self.allow_nan_stats else [  # pylint: disable=g-long-ternary
+          assert_util.assert_less(
+            2. * tf.ones([], dtype=self.dtype),
+            concentration,
+            message='Variance undefined for concentration <= 2.'),
+        ]):
+      b = 1. / self.concentration
+      variance = self.scale ** 2 * (1. / sinc(2*b) - 1. / sinc(b)**2)
+      return tf.where(self.concentration > 2., variance, np.nan)
 
   def _mode(self):
-    mode = self.scale * ((self.concentration - 1.)/(self.concentration + 1.)
-                         )**(1./self.concentration)
-    return tf.where(tf.greater(self.concentration, 1.), mode, np.nan)
+    concentration = tf.convert_to_tensor(self.concentration)
+    with tf.control_dependencies(
+        [] if self.allow_nan_stats else [  # pylint: disable=g-long-ternary
+          assert_util.assert_less(
+            tf.ones([], dtype=self.dtype),
+            concentration,
+            message='Mode undefined for concentration <= 1.'),
+        ]):
+      mode = self.scale * ((self.concentration - 1.)/(self.concentration + 1.)
+                           )**(1./self.concentration)
+      return tf.where(self.concentration > 1., mode, np.nan)
 
   def _entropy(self):
     return (tf.math.log(self.scale / self.concentration) + 2.) / tf.math.log(2.)
--- a/tensorflow_probability/python/distributions/loglogistic.py
+++ b/tensorflow_probability/python/distributions/loglogistic.py
@@ -116,11 +116,10 @@ class LogLogistic(transformed_distribution.TransformedDistribution):
       return tf.where(scale > 0.5, np.nan, variance)
 
   def _mode(self):
-    mode = tf.math.exp(
-      self.loc
-      + self.scale * (tf.math.log(1. - self.scale) - tf.math.log(self.scale)
-                      - tf.math.log1p(1./self.scale)))
-    return tf.where(self.scale > 1., 0.,  mode)
+    log_mode = self.loc + self.scale * (
+        tf.math.log1p(- self.scale) - tf.math.log1p(self.scale)
+    )
+    return tf.where(self.scale > 1., 0., tf.math.exp(log_mode))
 
   def _entropy(self):
     return 2. + self.loc + tf.math.log(self.scale)
--- a/tensorflow_probability/python/distributions/loglogistic.py
+++ b/tensorflow_probability/python/distributions/loglogistic.py
@@ -91,27 +91,29 @@ class LogLogistic(transformed_distribution.TransformedDistribution):
     return self.distribution.scale
 
   def _mean(self):
+    scale = tf.convert_to_tensor(self.scale)
     with tf.control_dependencies(
         [] if self.allow_nan_stats else [  # pylint: disable=g-long-ternary
           assert_util.assert_less(
-            self.scale,
+            scale,
             tf.ones([], dtype=self.dtype),
             message='Mean undefined for scale > 1.'),
         ]):
-      mean = tf.math.exp(self.loc) / sinc(self.scale)
-      return tf.where(self.scale > 1., np.nan, mean)
+      mean = tf.math.exp(self.loc) / sinc(scale)
+      return tf.where(scale > 1., np.nan, mean)
 
   def _variance(self):
+    scale = tf.convert_to_tensor(self.scale)
     with tf.control_dependencies(
-        [] if self.allow_nan_stats else [  # pylint: disable=g-long-ternary
+    [] if self.allow_nan_stats else [  # pylint: disable=g-long-ternary
           assert_util.assert_less(
-            self.scale,
+            scale,
             0.5 * tf.ones([], dtype=self.dtype),
             message='Variance undefined for scale > 1/2.'),
         ]):
       variance = tf.math.exp(2 * self.loc) * (
-          1. / sinc(2 * self.scale) - 1. / sinc(self.scale) ** 2)
-      return tf.where(self.scale > 0.5, np.nan, variance)
+          1. / sinc(2 * scale) - 1. / sinc(scale) ** 2)
+      return tf.where(scale > 0.5, np.nan, variance)
 
   def _mode(self):
     mode = tf.math.exp(
--- a/tensorflow_probability/python/distributions/loglogistic.py
+++ b/tensorflow_probability/python/distributions/loglogistic.py
@@ -51,9 +51,9 @@ class LogLogistic(transformed_distribution.TransformedDistribution):
 
     Args:
       scale: Floating-point `Tensor`;
-        the scale of the log-Logistic distribution(s).
+        the scale of the log-logistic distribution(s).
       concentration: Floating-point `Tensor`;
-        the shape of the log- Logistic distribution(s).
+        the shape of the log-logistic distribution(s).
       validate_args: Python `bool`, default `False`. Whether to validate input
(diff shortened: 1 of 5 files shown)


tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,153 @@
...
+__all__ = [
+    'LogLogistic',
+]
+
+
+class LogLogistic(transformed_distribution.TransformedDistribution):
+  """The log-logistic distribution."""
+
+  def __init__(self,
+               alpha,
```
brianwa84: We typically use terms like `scale` and `concentration` for these, and avoid greek.

srvasude:
+1 I might try to rephrase this as a TransformedDistribution.

npfp:
Yes, the main point is to use the (relatively simple) closed formula for the various quantities and gain some numerical stability.

tensorflow_probability/python/distributions/loglogistic.py
npfp: Regarding the greeks, I'll change that, thx

npfp:
@srvasude I'm not sure to understand. It's already inheriting from TransformedDistribution

srvasude:
Sorry I meant to use as much as TransformedDistribution as you can. But you already answered why you are overriding methods, so you can ignore my comment :).

npfp:
Hi there, Hope you're safe.
 I just wonder if there was anything pending on my side on this PR?

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,151 @@
...
+    return self._concentration
+
+  def _mean(self):
+    b = 1. / self.concentration
+    mean = self.scale / sinc(b)
+    return tf.where(tf.greater(self.concentration, 1.), mean, np.nan)
+
+  def _variance(self):
+    b = 1. / self.concentration
+    variance = self.scale ** 2 * (1. / sinc(2*b) - 1. / sinc(b)**2)
```
srvasude, line 108: Do you know how numerically stable this calculation is?

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,151 @@
...
+    return tf.where(tf.greater(self.concentration, 2.), variance, np.nan)
+
+  def _mode(self):
+    mode = self.scale * ((self.concentration - 1.)/(self.concentration + 1.)
```
srvasude, line 112: Can this be done in log space (using `expm1` and `log1p` for better numerics)?

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,151 @@
...
+                         )**(1./self.concentration)
+    return tf.where(tf.greater(self.concentration, 1.), mode, np.nan)
+
+  def _entropy(self):
+    return (tf.math.log(self.scale / self.concentration) + 2.) / tf.math.log(2.)
```
srvasude, line 117: Prefer `np.log(2)`

tensorflow_probability/python/distributions/loglogistic_test.py
```diff
@@ -0,0 +1,110 @@
+# Copyright 2018 The TensorFlow Probability Authors.
```
srvasude, line 1: 2020

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,151 @@
...
+               scale,
+               concentration,
+               validate_args=False,
+               allow_nan_stats=True,
+               name='LogLogistic'):
+    """Construct a log-logistic distribution.
+
+    The LogLogistic distribution models positive-valued random variables
+    whose logarithm is logisticly distributed with mean `log(scale)` and
+    scale `1/concentration`. It is constructed as the exponential
```
srvasude, line 49:
```suggestion
    scale `1 / concentration`. It is constructed as the exponential
```

tensorflow_probability/python/distributions/loglogistic_test.py
```diff
@@ -0,0 +1,110 @@
...
+from tensorflow_probability.python.internal import test_util
+
+
+@test_util.test_all_tf_execution_regimes
+class LogLogiticTest(test_util.TestCase):
+
+  def setUp(self):
+    self._rng = np.random.RandomState(123)
+
+  def testLogLogisticStats(self):
```
srvasude, line 35: Separate out tests?

tensorflow_probability/python/distributions/loglogistic.py
npfp: I'm having a look

tensorflow_probability/python/distributions/loglogistic.py
npfp: I'm having a look

npfp:
@googlebot I fixed it.

tensorflow_probability/python/distributions/loglogistic.py
npfp: I'm not sure it will improve numerical stability:
it's only computed for `c >1`, so

```
[(c-1)/(c+1)]**(1/c) = exp{[ln(c-1) - ln(c+1)]/c}
```

The numerical issue is when `c~1` where `exp{- ln(c+1)/c} ~1/2` and therefore won't benefit from `log1p`.

What do you think ?

tensorflow_probability/python/distributions/loglogistic.py
npfp: There will be some numerical issues when `c -> 2`, and the only way to solve them would be to work out a numerical approximation  when `c~2` and use it with something like:

```
tf.where(c < 2.+ tol, approximation, real_value)
```
But I'm no sure if it would be a good thing to introduce a `tf.where` and a `tol`. What do you think ?

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,175 @@
...
+    with tf.name_scope(name) as name:
+      dtype = dtype_util.common_dtype([scale, concentration],
+                                      dtype_hint=tf.float32)
+      self._scale = tensor_util.convert_nonref_to_tensor(
+          scale, name='scale', dtype=dtype)
+      self._concentration = tensor_util.convert_nonref_to_tensor(
+          concentration, name='concentration', dtype=dtype)
+      super(LogLogistic, self).__init__(
+          distribution=logistic.Logistic(loc=tf.math.log(self.scale),
+                                         scale=1./self.concentration),
```
srvasude, line 76:
```suggestion
                                         scale=tf.math.reciprocal(self.concentration)),
```

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,175 @@
...
+    parameters = dict(locals())
+    with tf.name_scope(name) as name:
+      dtype = dtype_util.common_dtype([scale, concentration],
+                                      dtype_hint=tf.float32)
+      self._scale = tensor_util.convert_nonref_to_tensor(
+          scale, name='scale', dtype=dtype)
+      self._concentration = tensor_util.convert_nonref_to_tensor(
+          concentration, name='concentration', dtype=dtype)
+      super(LogLogistic, self).__init__(
+          distribution=logistic.Logistic(loc=tf.math.log(self.scale),
```
srvasude, line 75: This can be problematic (this and below). This is because if these are `tf.Variable`, this won't be gradient tape safe.

I wonder if it's better to accept `loc` and `scale` as parameters. We do this for the `LogNormal` distribution and use the parametrization, and so would be consistent with it. WDYT?

npfp: Why is it not gradient tape safe? Is it because the operations are defined at the initialisation ?

Regarding your suggestion, I'm not sure. I agree it would good to be consistent with the `LogNormal` but in the meantime, by doing so we lose the possibility to introduce a `loc` parameter later on  (and it makes me realise that if we keep this option I should remove:

```
  @property
  def loc(self):
    """Distribution parameter for the pre-transformed mean."""
    return self.distribution.loc
```

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,175 @@
...
+                                         scale=1./self.concentration),
+          bijector=exp_bijector.Exp(),
```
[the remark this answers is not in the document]
npfp: I implemented them as well as log_cdf and log_survival_function as both of them can also gain some numerical stability.

tensorflow_probability/python/distributions/loglogistic.py
srvasude: Yep that's exactly it.

I would probably try to be consistent with LogNormal. Also IIRC, GeneralizedLogLogistic has a slightly different parameterization so we might want to have that as a separate implementation anyways.

npfp: I see, thx.

I juste pushed a change but for some reasons I got nans for a couple of numpy tests only. Would you have any clue on why the numpy tests behave differently than the others?

npfp:
> You need to sign the CLA

I did sign it.

I believe the CLA check is failing because I committed @srvasude suggested changes. See https://github.com/tensorflow/probability/pull/843#issuecomment-620167926

srvasude:
Hrmm this came up before. For some reason it is using an @gmail.com account. I think you can change my email for the suggestion to @google.com and things should work?

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,161 @@
...
+
+  def _mean(self):
+    with tf.control_dependencies(
+        [] if self.allow_nan_stats else [  # pylint: disable=g-long-ternary
+          assert_util.assert_less(
+            self.scale,
+            tf.ones([], dtype=self.dtype),
+            message='Mean undefined for scale > 1.'),
+        ]):
+      mean = tf.math.exp(self.loc) / sinc(self.scale)
```
srvasude, line 101: For passing distribution properties test you will need to do something like:

`scale = tf.convert_to_tensor(self.scale)`

and then use scale.

Multiple uses of a `self....` property will fail the test. This will need to be changed here and below

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,161 @@
...
+            message='Variance undefined for scale > 1/2.'),
+        ]):
+      variance = tf.math.exp(2 * self.loc) * (
+          1. / sinc(2 * self.scale) - 1. / sinc(self.scale) ** 2)
+      return tf.where(self.scale > 0.5, np.nan, variance)
+
+  def _mode(self):
+    mode = tf.math.exp(
+      self.loc
+      + self.scale * (tf.math.log(1. - self.scale) - tf.math.log(self.scale)
```
srvasude, line 119: `scale * (tf.math.log1p(-scale) - tf.math.log1p(scale))`

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,161 @@
...
+            self.scale,
+            0.5 * tf.ones([], dtype=self.dtype),
+            message='Variance undefined for scale > 1/2.'),
+        ]):
+      variance = tf.math.exp(2 * self.loc) * (
+          1. / sinc(2 * self.scale) - 1. / sinc(self.scale) ** 2)
+      return tf.where(self.scale > 0.5, np.nan, variance)
+
+  def _mode(self):
+    mode = tf.math.exp(
```
srvasude, line 117: Break this expression in to log_mode, and then exponentiate for the mode?

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,161 @@
...
+  def _log_z(self, x):
+    """Returns log of the standardized input."""
+    with tf.name_scope('log_standardize'):
+      return (tf.math.log(x) - self.loc) / self.scale
+
+  def _log_prob(self, x):
+    return (-tf.math.log(self.scale) - self.loc
+            + (1. - self.scale) * self._log_z(x)
+            - 2 * tf.math.log1p(
+              tf.math.exp(self._log_z(x)))
```
srvasude, line 135: Prefer using tf.math.softplus.

`tf.math.softplus(x) == tf.math.log1p(tf.math.exp(x))`

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,161 @@
...
+            )
+
+  def _log_cdf(self, x):
+    return -tf.math.log1p(tf.math.exp(-self._log_z(x)))
```
srvasude, line 139: softplus here as well

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,161 @@
...
+
+  def _log_survival_function(self, x):
+    return - self._log_z(x) + self._log_cdf(x)
```
srvasude, line 142: nit: remove leading - sign

tensorflow_probability/python/distributions/loglogistic.py
```diff
@@ -0,0 +1,161 @@
...
+        ]):
+      variance = tf.math.exp(2 * self.loc) * (
+          1. / sinc(2 * scale) - 1. / sinc(scale) ** 2)
+      return tf.where(scale > 0.5, np.nan, variance)
+
+  def _mode(self):
+    log_mode = self.loc + self.scale * (
+        tf.math.log1p(- self.scale) - tf.math.log1p(self.scale)
+    )
+    return tf.where(self.scale > 1., 0., tf.math.exp(log_mode))
```
srvasude, line 122: You'll need to cast `0.` and `np.nan` in these wheres unfortunately (this is for dtype=tf.float64).

tensorflow_probability/python/distributions/loglogistic_test.py
```diff
@@ -0,0 +1,209 @@
...
+
+    self.assertAllClose(
+      self.evaluate(pdf),
+      stats.fisk.pdf(x, loc=0., scale=log_logistic_scale, c=1./scale)
+    )
+
+  def testLogLogisticLogPDF(self):
+    log_logistic_scale = 1.5
+    loc = np.log(log_logistic_scale)
+    scale = 1./0.4
```
srvasude, line 141: Write as 2.5? Here and elsewhere

tensorflow_probability/python/distributions/loglogistic_test.py
```diff
@@ -0,0 +1,209 @@
...
+                        self.evaluate(dist.mean()),
+                        atol=0.1)
+    self.assertAllClose(np.std(samples),
+                        self.evaluate(dist.stddev()),
+                        atol=0.5)
+
+  def testLogLogisticPDF(self):
+    log_logistic_scale = 1.5
+    loc = np.log(log_logistic_scale)
+    scale = 1./0.4
```
srvasude, line 126: Use batches and broadcasting for these pdf/cdf related tests? Could also add separate tests where you have the parameters loc and scale broadcast, so you have a non-trivial batch of LogLogistic distributions.

srvasude:
Can you address the linter issues mentioned in Travis? Once those are fixed, I can approve and pull in. Thanks again for this PR!

npfp:
> Can you address the linter issues mentioned in Travis? Once those are fixed, I can approve and pull in. Thanks again for this PR!

Done

Many thanks for your review, I learnt a lot!

PR #844, opened 2020-03-17
Add the Frechet cdf

npfp:
Implement the Fréchet cdf (https://en.wikipedia.org/wiki/Fréchet_distribution).

### DIFF
--- a/tensorflow_probability/python/bijectors/frechet_cdf.py
+++ b/tensorflow_probability/python/bijectors/frechet_cdf.py
@@ -32,7 +32,8 @@ __all__ = [
 
 
 class FrechetCDF(bijector.Bijector):
-  """Compute `Y = g(X) = exp(-((X - loc) / scale) ** - alpha)`, the Frechet CDF.
+  """Compute `Y = g(X) = exp(-((X - loc) / scale) ** - concentration)`,
+  the Frechet CDF.
 
   This bijector maps inputs from `[loc, inf]` to `[0, 1]`. The inverse of the
   bijector applied to a uniform random variable `X ~ U(0, 1)` gives back a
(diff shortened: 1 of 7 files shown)


npfp:
Purpose is to add the Frechet distribution eventually. But the discussion on PR#843 made me think than maybe I could go with a TransformedDistribution using the property.

If X∼U(0,1) (Uniform distribution (continuous)) then m+s(−log⁡(X))^(−1/α) ∼ Frechet
https://en.wikipedia.org/wiki/Fréchet_distribution

tensorflow_probability/python/bijectors/frechet_cdf.py
```diff
@@ -0,0 +1,169 @@
...
+    return self._concentration
+
+  @classmethod
+  def _is_increasing(cls):
+    return True
+
+  def _forward(self, x):
+    with tf.control_dependencies(self._maybe_assert_valid_x(x)):
+      z = (x - self.loc) / self.scale
+      return tf.exp(-z ** (- self.concentration))
```
srvasude: We generally prefer to tf.convert_to_tensor(...) quantities that are used multiple times before using them below. The test I mention will catch this.

npfp:
I added the Fréchet constraint. Please let me know if there is anything wrong.
Thx

npfp:
Hi @brianwa84, thx!

Looking forward to having it merged. Please let me know if I can be of any help.

Also I'm not sure about the tests that are failing, are they related to my PR?

brianwa84:
While pulling this in I've added a TODO to test a few more extreme values. If you have a moment once this merges, please send a PR to do so and drop the TODO.

npfp:
I'll do. Thx a lot

Issue #819, opened 2020-02-28
Conditional Covariance


missmagnum:
Is the line below calculated regarding conditional covariance with Schur complement?
If yes, should it be:

```none
new_Sigma[q] :=  I -  inv(Sigma[q])
```

https://github.com/tensorflow/probability/blob/dc29e851e7bb4df76b3c355e80621a188aa0951c/tensorflow_probability/examples/vae.py#L82

PR #858, opened 2020-03-29
Changed formula in vae.py

manda-creator:
Issue: #819


tensorflow_probability/examples/vae.py
```diff
@@ -79,7 +79,7 @@
     new_mu[p] := 0
     new_Sigma[p] := eye(d)
     new_mu[q] := inv(chol(Sigma[p])) @ (mu[p] - mu[q])
-    new_Sigma[q] := inv(Sigma[q]) @ Sigma[p]
+    new_Sigma[q] :=  I -  inv(Sigma[q])
```
srvasude, line 82: What's the rationale for this change? Doing a change of variables, one should expect the covariance to transform with the inverse covariance (rather than a complement).

PR #859, opened 2020-04-01
Conditional MAF

tomcharnock:
As part of pydelfi (https://github.com/justinalsing/pydelfi) we need conditional distributions. We also wanted to make it completely integrated with tfp, but this needed a few modifications. Basically we now allow for a "conditional" kwarg to be passed around all over the place (so where there was no option for kwarg we've added it - in chain.py, normal.py, etc.). The main change is in the MAF, but for use in sampling we needed to add extra "conditional" arguments to the transformed distribution call (and distribution.py). I think there is also now a truncated multivariate tril, which is just a mix between a blockwise uniform and a tril. 

What do you think?

P.S. - sorry this is on v0.8.0 and not a newer version - we needed a tfp which worked with a stable tf2 rather than a nightly - most users of pydelfi are not seasoed tf2/tfp and so the less debugging they have due to nightly versions, the better.


tomcharnock:
@googlebot I signed it!

tomcharnock:
@googlebot I consent

tomcharnock:
@googlebot I consent.

junpenglao:
Hi Tom, I think you will need to rebase to master branch otherwise it will be quite difficult to review.

PR #881, opened 2020-04-16
ENH: Added binomial distribution to GLM

fearghuskeeble:
This distribution needs a parameter (tensor of total_count) unlike any of the current ones so I subclassed the `__init__` to allow for this. The use would look like:

`tfp.glm.fit(model_matrix=x, response=y, model=tfp.glm.Binomial(n))`
`tfp.glm.Binomial(n).log_prob(y, r)`

There are other ways to do this like passing kwargs through `fit` and the `ExponentialFamily` functions. I'm not sure what is preferred but this is my proposal and I think it generalises fairly well. I have also done the test for it.


fearghuskeeble:
@jvdillon I see you have been doing some larger changes in the glm so I updated the style of this to fit with [your more recent work](https://github.com/tensorflow/probability/commit/efbb4286873d70e1ae00a50af6cbbe5730c9363a).

jvdillon:
Hi fearghuskeeble. Thanks for your contribution. I actually wasnt aware of your PR otherwise I would have checked with you first. That said, your changes look great to me. Thanks for the contiribution!

PR #890, opened 2020-04-21
Add a cross_entropy_with_quantiles method

npfp:
- Make use of log_cdf and log_survival_function to comput the cross entropy  and avoid numerical overflow.

- Test it with the exponential distribution

- Improve log_survival_function of the exponential distribution

### DIFF
--- a/tensorflow_probability/python/distributions/distribution.py
+++ b/tensorflow_probability/python/distributions/distribution.py
@@ -1374,6 +1374,34 @@ class Distribution(_BaseDistribution):
     # must ensure that assertions are applied for both `self` and `other`.
     return self._kl_divergence(other)
 
+  def cross_entropy_with_quantiles(self, pred_true, quantiles,
+                                   name="cross_entropy_with_quantiles"):
+    """Computes the cross entropy given quantiles.
+
+    Args:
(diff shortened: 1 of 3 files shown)


brianwa84:
This looks like a neat approach. Glancing at the code it seems like it's maybe limited to cases where `pred_true` is in the set `{0, 1}`? (Is that accurate?)

Is there a paper you're implementing, or could you work out a bit of the math in the docstring? Is this a function that meaningfully applies to every scalar-real distribution?

npfp:
Thx @brianwa84 

I just improved the doctoring

 I was not able to find any paper documenting the generalization implemented (from logic/probit models to any distribution model), but hopefully the docstring should be self-explanatory.

You're right, this indeed meaningfully applies to every scalar-real distribution. The use-case (`pred_true`  in `{0,1}`) is really the same than https://www.tensorflow.org/api_docs/python/tf/nn/sigmoid_cross_entropy_with_logits
but for other distributions than the logistic one.
I kept the implementation choice of `sigmoid_cross_entropy_with_logits` to not impose `pred_true` to be in `{0,1}` though.

npfp:
Hi there,

Is there anything pending on my side? Please let me know if this needs more detailed explanation.

Thx

PR #900, opened 2020-04-24
Added KL divergence calculation between two Cauchy distributions.

gergely-flamich:
KL calculation is based on [this paper](https://arxiv.org/pdf/1905.10965.pdf).

### DIFF
--- a/tensorflow_probability/python/distributions/cauchy_test.py
+++ b/tensorflow_probability/python/distributions/cauchy_test.py
@@ -432,7 +432,10 @@ class CauchyTest(test_util.TestCase):
       with tf.control_dependencies([scale.assign([1., 2., -3.])]):
         self.evaluate(d.mean())
 
-  def testCauchyCauchyKL(self):
+  def testCauchyCauchyKLWithAnalytic(self):
+    """
+    Test the KL by comparing it against the analytic formula for it
+    """
     batch_size = 6
     mu_a = np.array([3.0] * batch_size)
     gamma_a = np.array([1.0, 2.0, 3.0, 1.5, 2.5, 3.5])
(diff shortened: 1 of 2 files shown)


tensorflow_probability/python/distributions/cauchy_test.py
```diff
@@ -432,6 +432,33 @@ def testAssertsPositiveScaleAfterMutation(self):
...
+    c_b = tfd.Cauchy(loc=mu_b, scale=gamma_b, validate_args=True)
+
+    kl = tfd.kl_divergence(c_a, c_b)
+    reverse_kl = tfd.kl_divergence(c_b, c_a)
+    kl_val = self.evaluate(kl)
+    reverse_kl_val = self.evaluate(reverse_kl)
+
+    kl_expected = np.log((gamma_a + gamma_b)**2 + (mu_a - mu_b)**2) - np.log(4. * gamma_a * gamma_b)
+
+    x = c_a.sample(int(1e5), seed=test_util.test_seed())
```
srvasude, line 452: Let's split these tests up (testing against numpy vs. sampling based test.)

gergely-flamich: I split the test into two functions, one with the "WithAnalytic" and other with the "WithMC" postfix, and moved the numpy and sampling-based tests to them respectively.

Also note: I based the tolerances for the MC tests on the ones for tfd.Normal

gergely-flamich: Oh also, the list of registered KLs is reported here:
https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/kl_divergence

Should I also update this, or is this done automatically / internally after the merge?

srvasude: This will be done automatically.

Issue #874, opened 2020-04-14
Generalized extreme value distribution


jedisom:
I see that the Gumbel distribution has been created based on this link: https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Gumbel
This is the xi =0 case per https://en.wikipedia.org/wiki/Generalized_extreme_value_distribution

I would like to use a more generalized version of the extreme value distrubutions allowing xi to be non-zero; Gumbel xi =0, Frechet xi > 0, and/or Weibull xi < 0.

brianwa84:
Does the FrechetCDF bijector help here?
tfb.Invert(tfb.FrechetCDF(...))(tfd.Uniform(0,1))

jedisom:
@brianwa84, That appears to be a way to get a random variable from a Frechet distribution.  The application I'm looking for is implementing these distributions within a Neural Network output layer allowing the network to learn the location, scale _AND_ shape parameters.  Something like:

`tfp.layers.DistributionLambda(lambda t: tfd.GeneralizedExtremeDistribution(loc=t[..., 0], scale=tf.nn.softplus(t[..., 1]), shape=t[..., 2]))`

The above obviously doesn't exist yet, but there might be a clever way to create the same effect with bijectors as you've suggested in your response above.

srvasude:
There might be a way of chaining a `PowerTransform` bijector here, since essentially the difference between GEV and Gumbel is replacing the exponential in the exponent with a power transform.

Probably something like this would work (the forward of this would give you the CDF of a GEV):
`Exp(Reciprocal(PowerTransform(power=xi)(Scale(1/sigma)(Shift(-mu)))))`
but I'd probably recommend implementing from scratch (since I think the numerics of the above might not be so great). Contributions welcome here.

jedisom:
I may take a stab at building the GeneralizedExtremeValue distribution class.  This would be my first contribution to the repo.  My initial thought was that I should be able to use the Gumbel class as a template and then just augment it with the shape/concentration parameter the generalized version needs, and then re-use the existing gumbel, frechet and weibull bijectors.  Thoughts or feedback on this approach?

blacksde:
Hi @srvasude @brianwa84 @jedisom I am raising a pr of Generalized extreme value distribution cdf bijector. It could be a partial solution of this issue. A future gev distribution could be added based on this. Could you help review the code?

jedisom:
[...]

PR #901, opened 2020-04-25
add gev_cdf bijector

blacksde:
This PR is a partial solution of the issue #874 . 

Here I implement the generalized extreme value bijector with the following cdf fucntion:

`Y = g(X) = exp(-t(X))` where `t(x) = (1 + shape * (x - loc) / scale) ) ^ (-1 / shape)` and shape is not 0.

I followed the notation of in the wiki [Generalized extreme value distribution](
  https://https://en.wikipedia.org/wiki/Generalized_extreme_value_distribution), and the cdf has three parameters: loc, scale, and shape. Especially, when shape is 0, the gev dist has a special form, which is in fact the gumbel distribution. Thus I have the parameter control here that when the user tried to set shape = 0, an error is thrown and the error message leads the user to use the gumbel cdf.

For the test case, I use the `scipy.stats.genextreme` as the benchmark. The notation is slightly different: c in `scipy.stats.genextreme` is -shape in my notation. The rest are the same.

In the future, we can add the gev distribution base on this bijector/cdf function.

Update on Apr 26: Now I support the case shape = 0. Also, I change the name of the parameter and bijector and some math inside the function as suggested by @srvasude


jedisom:
@blacksde, For the application I'm interested in on issue #874, we will need the bijector to be able to handle shape parameters equal to 0.  This is because as the model "learns" the shape parameter, it will need to be allowed to try a shape parameter of 0 (it might be the best fit), then wander around (positive and negative) until it finds the shape parameter that fits the data best.  Essentially the domain for the shape parameter needs to be allowed to be anything in (-inf, inf)

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,178 @@
+# Copyright 2018 The TensorFlow Probability Authors.
```
srvasude, line 1: 2020

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,178 @@
...
+from tensorflow_probability.python.internal import dtype_util
+from tensorflow_probability.python.internal import tensor_util
+
+
+__all__ = [
+    'GEVCDF',
+]
+
+
+class GEVCDF(bijector.Bijector):
```
srvasude, line 34: I think unabbreviating this to GeneralizedExtremeValueCDF is fine.

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,178 @@
...
+  t(y; loc, scale, shape) = (1 + shape * (y - loc) / scale) ) ^ (-1 / shape)
+  pdf(y; loc, scale, shape) = t(y; loc, scale, shape) ^ (1 + shape) * exp(
+  - t(y; loc, scale, shape) ) / scale
+  ```
+  """
+
+  def __init__(self,
+               loc=0.,
+               scale=1.,
+               shape=0.5,
```
srvasude, line 54: Is there intuition why this should be the default value? For `loc` and `scale` I understand, but not quite for `shape`.

srvasude, line 54: Can we change the name of this parameter to something like `concentration`? We generally avoid naming parameters `shape` since it can easily be conflated with `shape` as a TF concept.

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,178 @@
...
+    return self._shape
+
+  @classmethod
+  def _is_increasing(cls):
+    return True
+
+  def _forward(self, x):
+    with tf.control_dependencies(self._maybe_assert_valid_x(x)):
+      z = (x - self.loc) / self.scale
+      t = (tf.ones_like(z) + self.shape * z) ** (-1 / self.shape)
```
srvasude, line 123:
```suggestion
      t = (1. + self.shape * z) ** (-1 / self.shape)
```

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,178 @@
...
+      return tf.exp(-t)
+
+  def _inverse(self, y):
+    with tf.control_dependencies(self._maybe_assert_valid_y(y)):
+      t = - tf.math.log(y)
+      z = tf.ones_like(y) - t ** (-self.shape)
```
srvasude, line 129:
```suggestion
      z = 1. - t ** (-self.shape)
```

tensorflow_probability/python/bijectors/gev_cdf_test.py
```diff
@@ -0,0 +1,92 @@
+# Copyright 2018 The TensorFlow Probability Authors.
```
srvasude, line 1: 2020

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,178 @@
...
+__all__ = [
+    'GEVCDF',
+]
+
+
+class GEVCDF(bijector.Bijector):
+  """Compute `Y = g(X) = exp(-t(X))`, the GEV CDF,
+  where `t(x) = (1 + shape * (x - loc) / scale) ) ^ (-1 / shape)`.
+
+  This bijector maps inputs from `[-inf, inf]` to `[0, 1]`. The inverse of the
```
srvasude, line 38: In this case, the support is constrained by the parameters?

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,178 @@
...
+               shape=0.5,
+               validate_args=False,
+               name='gev_cdf'):
+    """Instantiates the `GEVCDF` bijector.
+
+    Args:
+      loc: Float-like `Tensor` that is the same dtype and is
+        broadcastable with `scale` and `shape`.
+        This is `loc` in `Y = exp(-t(X))`
+        where `t(x) = (1 + shape * (x - loc) / scale) ) ^ (-1 / shape)`.
```
srvasude, line 63: We usually use pythonic math, so something like `**` instead of `^`.

tensorflow_probability/python/bijectors/gev_cdf.py
blacksde: Here I can't find a trivial default value. In the [wiki](https://en.wikipedia.org/wiki/Generalized_extreme_value_distribution), it demonstrates +/-0.5 and 0. As I didn't include 0 here so I picked 0.5. Maybe I should include the case of zero shape/concentration and set the default value to be 0?

blacksde: Sure. That's a better name compared to `shape`.

tensorflow_probability/python/bijectors/gev_cdf.py
blacksde: Yes, I will make a more accurate statement here.

tensorflow_probability/python/bijectors/gev_cdf.py
blacksde: OK~

blacksde:
> @blacksde, For the application I'm interested in on issue #874, we will need the bijector to be able to handle shape parameters equal to 0. This is because as the model "learns" the shape parameter, it will need to be allowed to try a shape parameter of 0 (it might be the best fit), then wander around (positive and negative) until it finds the shape parameter that fits the data best. Essentially the domain for the shape parameter needs to be allowed to be anything in (-inf, inf)
> 
> Perhaps adding an epsilon (tolerance) variable where the behavior would switch between the Gumbel behavior and the behavior you have coded would resolve the issue. The default epsilon could be arbitrarily small (e.g. 1e-5 or something similar). If the shape parameter is within [-epsilon, epsilon] then the Gumbel bijector would be used, otherwise the GEV code would be used.
> 
> What are your thoughts?

That's a good idea. Yes, this is a nice way to cover all shapes from -inf to inf. I will try to implement it.

blacksde:
> * `PowerTransform`

Hi @srvasude Thanks for all your comments. Just notice there are some similar math in powertransform, will improve math here using that as the reference.

npfp:
I'm also really interested into this family of distribution (so my PR https://github.com/tensorflow/probability/pull/844).

Considering the fact, that the Fréchet distribution (`concentration > 0`) and the Gumbel (`concentration= 0`) are already available, shouldn't we take advantage of those ?

Wouldn't it be preferable to implement to the Reversed Weibull (`concentration > 0`)  bijector, and then make the GEV points to Gumbel, Fréchet or the Reversed Weibull depending on the `concentration` value ?

npfp:
Also, in case you're interested, I'm working on a PR for the Fréchet distribution. 

A draft can be found here: https://github.com/npfp/probability/tree/frechet_2

blacksde:
> I'm also really interested into this family of distribution (so my PR #844).
> 
> Considering the fact, that the Fréchet distribution (`concentration > 0`) and the Gumbel (`concentration= 0`) are already available, shouldn't we take advantage of those ?
> 
> Wouldn't it be preferable to implement to the Reversed Weibull (`concentration > 0`) bijector, and then make the GEV points to Gumbel, Fréchet or the Reversed Weibull depending on the `concentration` value ?

Hi @npfp thanks for your comment. The math behind the case conc > 0 and conc < 0 are almost the same so maybe putting them together is better. Also, the difficulty is that in an array of conc some are positive, some are negative and some are zeros, it is hard to map each of them to a separate bijector.

blacksde:
Hi @srvasude , I have already updated the code, could you help review the updated codes? Thx!

Also, I considered the suggestion of @jedisom , but I checked `power_transform` which has the similar situation while it switches on power = 0 and power != 0, so I decided to use this strategy. We can consider further update this if it has the issue of dividing a extreme small number in pratice.

npfp:
> > I'm also really interested into this family of distribution (so my PR #844).
> > Considering the fact, that the Fréchet distribution (`concentration > 0`) and the Gumbel (`concentration= 0`) are already available, shouldn't we take advantage of those ?
> > Wouldn't it be preferable to implement to the Reversed Weibull (`concentration > 0`) bijector, and then make the GEV points to Gumbel, Fréchet or the Reversed Weibull depending on the `concentration` value ?
> 
> Hi @npfp thanks for your comment. The math behind the case conc > 0 and conc < 0 are almost the same so maybe putting them together is better. Also, the difficulty is that in an array of conc some are positive, some are negative and some are zeros, it is hard to map each of them to a separate bijector.


Sorry I was not clear. I was thinking to something like:

  ```
def __init__(self,
                loc=0.,
                scale=1.,
                concentration=0,
                tol=1e-5,
                validate_args=False,
                name='generalizedextremevalue_cdf'):
    parameters = dict(locals())
     with tf.name_scope(name) as name:
       dtype = dtype_util.common_dtype(
           [loc, scale, concentration], dtype_hint=tf.float32)
       self._loc = tensor_util.convert_nonref_to_tensor(
           loc, dtype=dtype, name='loc')
       self._scale = tensor_util.convert_nonref_to_tensor(
           scale, dtype=dtype, name='scale')
       self._concentration = tensor_util.convert_nonref_to_tensor(
           concentration, dtype=dtype, name='concentration')
       self._gumbel_bijector = gumbel_cdf_bijector(loc, scale, allow_nan_stats=True)
       self._frechet_bijector = frechet_cdf_bijector(loc, scale, concentration, allow_nan_stats=True)
       self._weibull_bijector = weibull_cdf_bijector(loc, scale, concentration, allow_nan_stats=True)
       super(GeneralizedExtremeValueCDF, self).__init__(
           validate_args=validate_args,
           forward_min_event_ndims=0,
           parameters=parameters,
           name=name)

@property
def is_gumbel(self):
   return tf.where(tf.logical_and(self.concentration >= -self._tol, self.concentration <= self._tol

@property
def is_frechet(self):
  return self.concentration > self._tol

@property
def is_weibull(self):
  return self.concentration < - self._tol
```

And then for example:

```
  def _forward(self, x):
     with tf.control_dependencies(self._maybe_assert_valid_x(x)):
       forward = tf.where(self.is_frechet, self._frechet_bijector(x),  self._weibull_bijector(x))
       return tf.where(self.is_gumbel, self._gumbel_bijector(x), forward
``` 

And so forth so on.

This way, it avoids any code duplication, any change in the Gumbel, the Frechet or the Weibull will benefit to the GEV.


What do you think?

**Comment by blacksde** [comment]
> > > I'm also really interested into this family of distribution (so my PR #844).
> > > Considering the fact, that the Fréchet distribution (`concentration > 0`) and the Gumbel (`concentration= 0`) are already available, shouldn't we take advantage of those ?
> > > Wouldn't it be preferable to implement to the Reversed Weibull (`concentration > 0`) bijector, and then make the GEV points to Gumbel, Fréchet or the Reversed Weibull depending on the `concentration` value ?
> > 
> > 
> > Hi @npfp thanks for your comment. The math behind the case conc > 0 and conc < 0 are almost the same so maybe putting them together is better. Also, the difficulty is that in an array of conc some are positive, some are negative and some are zeros, it is hard to map each of them to a separate bijector.
> 
> Sorry I was not clear. I was thinking to something like:
> 
> ```
> def __init__(self,
>               loc=0.,
>               scale=1.,
>               concentration=0,
>               tol=1e-5,
>               validate_args=False,
>               name='generalizedextremevalue_cdf'):
>   parameters = dict(locals())
>    with tf.name_scope(name) as name:
>      dtype = dtype_util.common_dtype(
>          [loc, scale, concentration], dtype_hint=tf.float32)
>      self._loc = tensor_util.convert_nonref_to_tensor(
>          loc, dtype=dtype, name='loc')
>      self._scale = tensor_util.convert_nonref_to_tensor(
>          scale, dtype=dtype, name='scale')
>      self._concentration = tensor_util.convert_nonref_to_tensor(
>          concentration, dtype=dtype, name='concentration')
>      self._gumbel_bijector = gumbel_cdf_bijector(loc, scale, allow_nan_stats=True)
>      self._frechet_bijector = frechet_cdf_bijector(loc, scale, concentration, allow_nan_stats=True)
>      self._weibull_bijector = weibull_cdf_bijector(loc, scale, concentration, allow_nan_stats=True)
>      super(GeneralizedExtremeValueCDF, self).__init__(
>          validate_args=validate_args,
>          forward_min_event_ndims=0,
>          parameters=parameters,
>          name=name)
> 
> @property
> def is_gumbel(self):
>  return tf.where(tf.logical_and(self.concentration >= -self._tol, self.concentration <= self._tol
> 
> @property
> def is_frechet(self):
> return self.concentration > self._tol
> 
> @property
> def is_weibull(self):
> return self.concentration < - self._tol
> ```
> 
> And then for example:
> 
> ```
>   def _forward(self, x):
>      with tf.control_dependencies(self._maybe_assert_valid_x(x)):
>        forward = tf.where(self.is_frechet, self._frechet_bijector(x),  self._weibull_bijector(x))
>        return tf.where(self.is_gumbel, self._gumbel_bijector(x), forward)
> ```
> 
> And so forth so on.
> 
> This way, it avoids code duplication, and any change in the Gumbel, the Frechet or the Weibull bijectors will benefit to the GEV.
> 
> What do you think?

I am not sure whether using three bijectors inside one is a good idea. At least it takes a longer time to call a forward/inverse function. Also, to implement GEV with conc < 0, we need to reverse current Weibull with some affine transform, it is against our first motivation to add this cdf from scratch without any chaining.

**Comment by npfp** [comment]
> > Sorry I was not clear. I was thinking to something like:
> > ```
> > def __init__(self,
> >               loc=0.,
> >               scale=1.,
> >               concentration=0,
> >               tol=1e-5,
> >               validate_args=False,
> >               name='generalizedextremevalue_cdf'):
> >   parameters = dict(locals())
> >    with tf.name_scope(name) as name:
> >      dtype = dtype_util.common_dtype(
> >          [loc, scale, concentration], dtype_hint=tf.float32)
> >      self._loc = tensor_util.convert_nonref_to_tensor(
> >          loc, dtype=dtype, name='loc')
> >      self._scale = tensor_util.convert_nonref_to_tensor(
> >          scale, dtype=dtype, name='scale')
> >      self._concentration = tensor_util.convert_nonref_to_tensor(
> >          concentration, dtype=dtype, name='concentration')
> >      self._gumbel_bijector = gumbel_cdf_bijector(loc, scale, allow_nan_stats=True)
> >      self._frechet_bijector = frechet_cdf_bijector(loc, scale, concentration, allow_nan_stats=True)
> >      self._weibull_bijector = weibull_cdf_bijector(loc, scale, concentration, allow_nan_stats=True)
> >      super(GeneralizedExtremeValueCDF, self).__init__(
> >          validate_args=validate_args,
> >          forward_min_event_ndims=0,
> >          parameters=parameters,
> >          name=name)
> > 
> > @property
> > def is_gumbel(self):
> >  return tf.where(tf.logical_and(self.concentration >= -self._tol, self.concentration <= self._tol
> > 
> > @property
> > def is_frechet(self):
> > return self.concentration > self._tol
> > 
> > @property
> > def is_weibull(self):
> > return self.concentration < - self._tol
> > ```
> > 
> > 
> > And then for example:
> > ```
> >   def _forward(self, x):
> >      with tf.control_dependencies(self._maybe_assert_valid_x(x)):
> >        forward = tf.where(self.is_frechet, self._frechet_bijector(x),  self._weibull_bijector(x))
> >        return tf.where(self.is_gumbel, self._gumbel_bijector(x), forward)
> > ```
> > 
> > 
> > And so forth so on.
> > This way, it avoids code duplication, and any change in the Gumbel, the Frechet or the Weibull bijectors will benefit to the GEV.
> > What do you think?
> 
> I am not sure whether using three bijectors inside one is a good idea. At least it takes a longer time to call a forward/inverse function. Also, to implement GEV with conc < 0, we need to reverse current Weibull with some affine transform, it is against our first motivation to add this cdf from scratch without any chaining.

Ok, I was not sure if using inside bijectors could impact performance.

Regarding, the reversed Weibull, maybe it would make sense to implement it into its own class in case, someone i only interested into a GEV III. I'll work on it once, I'll finish the Fréchet implementation.

File: tensorflow_probability/python/bijectors/gev_cdf.py [hunk 9]
```diff
@@ -0,0 +1,197 @@
...
+    return self._concentration
+
+  @classmethod
+  def _is_increasing(cls):
+    return True
+
+  def _forward(self, x):
+    with tf.control_dependencies(self._maybe_assert_valid_x(x)):
+      conc = tf.convert_to_tensor(self.concentration)
+      condition = tf.equal(conc, 0.)
```
srvasude, line 129: Inline this.

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,197 @@
...
+
+      return tf.math.log(self.scale / y) + log_dt
+
+  def _forward_log_det_jacobian(self, x):
+    with tf.control_dependencies(self._maybe_assert_valid_x(x)):
+      scale = tf.convert_to_tensor(self.scale)
+      z = (x - self.loc) / scale
+
+      conc = tf.convert_to_tensor(self.concentration)
+      condition = tf.equal(conc, 0.)
```
srvasude, line 164: Inline

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,197 @@
...
+
+      z = (x - self.loc) / self.scale
+      t = tf.where(condition, tf.exp(-z),
+                   tf.exp(-tf.math.log1p(z * conc) / conc))
+      return tf.exp(-t)
+
+  def _inverse(self, y):
+    with tf.control_dependencies(self._maybe_assert_valid_y(y)):
+      t = - tf.math.log(y)
```
srvasude, line 138: Nit: No space between negative sign and `tf.math.log`

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,197 @@
...
+
+      return self.loc + self.scale * z
+
+  def _inverse_log_det_jacobian(self, y):
+    with tf.control_dependencies(self._maybe_assert_valid_y(y)):
+      t = - tf.math.log(y)
+
+      conc = tf.convert_to_tensor(self.concentration)
+      condition = tf.equal(conc, 0.)
+      log_dt = tf.where(condition, -tf.math.log(t),
```
srvasude, line 153: Can this just be (-conc -1.) * tf.math.log(t) without a tf.where?

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,197 @@
...
+      return tf.exp(-t)
+
+  def _inverse(self, y):
+    with tf.control_dependencies(self._maybe_assert_valid_y(y)):
+      t = - tf.math.log(y)
+
+      conc = tf.convert_to_tensor(self.concentration)
+      condition = tf.equal(conc, 0.)
+      z = tf.where(condition, -tf.math.log(t),
+                   tf.math.expm1(-tf.math.log(t) * conc) / conc)
```
srvasude, line 143: I wonder if Taylor expanding for small conc is the right thing to do here. That would guarantee stability of this expression

tensorflow_probability/python/bijectors/gev_cdf.py
blacksde: Yes we can do this~

tensorflow_probability/python/bijectors/gev_cdf.py
blacksde: I can add a Taylor expansion here. In fact, if we add Taylor expansion, we can add this to all similar places. Also, we can check that using Taylor expansion is in fact using the case conc=0, which is Gumbel bijector as suggested by @jedisom . Thus in my last update, I add a `tol` to control when the bijector converges to the Gumbel bijector.

blacksde:
Hi @srvasude , thx a lot for your review comments. I add a new par to control when the bijector converges to the Gumbel bijector as explained on the inline comments. Let me know whether this makes sense. I can rollback if this would lead to any possible problems I could not see.

tensorflow_probability/python/bijectors/gev_cdf.py
srvasude: I would avoid a tolerance parameter. I was thinking that there is a radius R where if log(t) and conc < R, then using a taylor expansion would be the right thing to do.
Happy to punt on this for now since it is a bit non-trivial.

blacksde:
> Realized this was marked still as review required (I responded to relevant comments inline).

Hi @srvasude Thx a lot for your comments. To solve the problem of the radius as you commented inline, I have the following proposal:

Now we have two formula to estimate
- ` (exp(- log(t) * conc)  - 1 )/ conc  \approx -log(t)`
- ` exp( -log(z * conc - 1)) / conc) \approx exp(-z) `

The first one is relatively easy. Let `-log(t) = a`. Then use the Taylor expansion around `conc = 0` and we have
```
(exp(a conc) - 1) / conc - a = a^2 conc / 2! + a^3 conc^2 / 3! + ....
```
Thus the dominating term  `a^2 conc / 2!` and `a^2 conc / 2! <= a^2 R / 2` when `conc < R` inside the disc. If we need to meet the precision `atol` and `rtol`, we have `a^2 R / 2 < atol` and `abs(a) R / 2 < rtol` (here we use `(exp(a conc) - 1) / conc \approx a` to simplify the calculation). Thus we have
```
R < min(2 * atol / a^2, 2 * rtol / abs(a), const).
```
Here we add a `const` such as `1.0` here to avoid the case that `abs(a)` is too small.

The second formula is relatively complex and we can use some less strict way to calculate. We start from the index. Use the Taylor expansion around `conc = 0` and then
```
-log(z * conc - 1) / conc = - z + z^2 conc / 2 - z^3 conc^2 /3 .... = -z (1 - z conc / 2 + z^2 conc^2 / 3 )
```

Then we can estimate the error by
```
exp( -log(z * conc - 1)) / conc) - exp(-z) =  exp(-z) (exp(1 - z conc / 2 + z^2 conc^2 / 3 ... ) - 1) =   exp(-z) * ( - z conc / 2 + z^2 conc^2 / 3 ... )
```
The (absolute value of) dominating term is `exp(-z) * abs(z) conc / 2 < exp(-z) * abs(z) R / 2` inside the disc.  If we need to meet the precision `atol` and `rtol`, we have `exp(-z) * abs(z) R / 2 < atol` and `abs(z) R / 2 < rtol`. Thus we have
```
R < min(2 * atol / (exp(-z) * abs(z)), 2 * rtol / abs(z), const).
```

Both of the above estimations can be more strict with the Taylor formula with the Lagrange remainder, but I believe the order of the above estimation is right. I will update the codes based on the above formula first. Let me know whether you had any comments on this proposal.

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,218 @@
...
+      super(GeneralizedExtremeValueCDF, self).__init__(
+          validate_args=validate_args,
+          forward_min_event_ndims=0,
+          parameters=parameters,
+          name=name)
+
+  @property
+  def loc(self):
+    """
+    The `loc` in `Y = exp(-t(X))`
```
srvasude, line 98: python docs have a 1 line doc for the first line. Can you try to condense the description to a one line doc (or a one line doc, followed by more explanation)? Here and below

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,218 @@
...
+      # when `abs(conc) < tol`, gev converges to the Gumbel dist.
+      # check the following link for the estimation
+      # of the threshold (also for _inverse and _forward_log_det_jacobian)
+      # https://github.com/tensorflow/probability/pull/901#issuecomment-629857187
+      tol_target = 1e-5
+      rtol_target = 2. * tol_target / tf.math.abs(z)
+      atol_target = 2. * tol_target / (tf.math.exp(-z) * tf.math.abs(z))
+      tol = tf.math.minimum(tf.math.minimum(rtol_target, atol_target),
+                            tf.constant(1., dtype=z.dtype))
+      t = tf.where(tf.math.less_equal(tf.math.abs(conc), tol),
```
srvasude, line 142: tf.math.abs(conc) <= tol

Here and below

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,218 @@
...
+      conc = tf.convert_to_tensor(self.concentration)
+
+      z = (x - self.loc) / self.scale
+
+      # for the numerical stability, we set a threshold `tol` that
+      # when `abs(conc) < tol`, gev converges to the Gumbel dist.
+      # check the following link for the estimation
+      # of the threshold (also for _inverse and _forward_log_det_jacobian)
+      # https://github.com/tensorflow/probability/pull/901#issuecomment-629857187
+      tol_target = 1e-5
```
srvasude, line 137: Note this tolerance needs to be dtype dependent (float64 vs float32). Do you think you can change that? I am also fine with dropping the tolerance stuff at the moment to get this checked in and later iterate on stability near 0.

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,218 @@
...
+  def _inverse_log_det_jacobian(self, y):
+    with tf.control_dependencies(self._maybe_assert_valid_y(y)):
+      t = -tf.math.log(y)
+
+      conc = tf.convert_to_tensor(self.concentration)
+      log_dt = (-conc - 1.) * tf.math.log(t)
+
+      return tf.math.log(self.scale / y) + log_dt
+
+  def _forward_log_det_jacobian(self, x):
```
srvasude, line 173: We could omit this and rely on inverse_log_det_jacobian since the calculation looks more numerically stable. WDYT?

tensorflow_probability/python/bijectors/gev_cdf.py
```diff
@@ -0,0 +1,218 @@
...
+                   -tf.math.log(t),
+                   tf.math.expm1(-tf.math.log(t) * conc) / conc)
+
+      return self.loc + self.scale * z
+
+  def _inverse_log_det_jacobian(self, y):
+    with tf.control_dependencies(self._maybe_assert_valid_y(y)):
+      t = -tf.math.log(y)
+
+      conc = tf.convert_to_tensor(self.concentration)
```
srvasude, line 168: Since this is used once, you can just use self.concentration inline

tensorflow_probability/python/bijectors/gev_cdf.py
blacksde: Yes, that's a good idea.  `_inverse_log_det_jacobian` has a much more numerically stable calculation formula. Will use  `_inverse_log_det_jacobian`  instead.

blacksde:
@srvasude thx for your review all these days. Right now I dropped the `tol` calculation. Also, I add a test case with small `conc` to check the numerical stability, and it looks good. I kept the commits with `tol` calculation in case we need this in the future.

blacksde:
@srvasude thx for pointing this out. I have some difficulty in running this bijector properties test locally even with a plain master branch. Let me know any further update I need to make this test work.

tensorflow_probability/python/bijectors/hypothesis_testlib.py
```diff
@@ -398,3 +402,19 @@ def frechet_constraint(loc):
...
+
+def gev_constraint(loc, scale, conc):
+  """
+  Maps `s` to:
+    1. [loc - scale / conc, inf) if conc > 0;
+    2. (-inf, inf) if conc = 0;
+    3. (-inf, loc - scale / conc] if conc < 0.
+  """
+  def constrain(x):
+    if conc == 0:
```
srvasude, line 414: Given that these are tensors, you'll need to use tf.where semantics here:

```suggestion
    tf.where(conc > 0., tf.math.softplus(x) + loc - scale / conc, tf.where(tf.equal(0., conc), x, loc - scale / conc - tf.math.softplus(x))
```

blacksde: Thx for your suggestion! Updated~

PR #906, opened 2020-04-30
Update RealNVP bijector docstrings

johannespitz:
Update the presumably outdated reference from `real_nvp_default_nvp` to `real_nvp_default_template`.


johannespitz:
@googlebot I signed it!

PR #912, opened 2020-05-03
fix: Use ascending sort for percentile

### LINKED ISSUE #864: tfp.stats.percentile sort order differs from docstring (and differs in behavior from NumPy)
# Description

The docstring for `v0.9.0` of `tfp.stats.percentile` states that

https://github.com/tensorflow/probability/blob/356cfddef026b3339b8f2a81e600acd2ff8e22b4/tensorflow_probability/python/stats/quantiles.py#L415-L416

From this description (specifically the "**from the minimum to the maximum**" part) I would assume that this means that for

```python
>>> import tensorflow as tf
>>> a = tf.constant([[10.0, 7.0, 4.0], [3.0, 2.0, 1.0]])
>>> a
<tf.Tensor: shape=(2, 3), dtype=float32, numpy=
array([[10.,  7.,  4.],
       [ 3.,  2.,  1.]], dtype=float32)>
```
the sorted version of it that `percentile` evaluates on would be in _ascending_ order

```python
>>> y = tf.reshape(a, [-1])
>>> tf.sort(y)
<tf.Tensor: shape=(6,), dtype=float32, numpy=array([ 1.,  2.,  3.,  4.,  7., 10.], dtype=float32)>
```

However, as I learned in this Stack Overflow question ([Different results with NumPy percentile and TensorFlow percentile for “nearest” interpolation method](https://stackoverflow.com/questions/61024629/different-results-with-numpy-percentile-and-tensorflow-percentile-for-nearest))

the sort order used is actually _descending_

https://github.com/tensorflow/probability/blob/356cfddef026b3339b8f2a81e600acd2ff8e22b4/tensorflow_probability/python/stats/quantiles.py#L884

which gives

```python
>>> sorted_y, _ = tf.math.top_k(y, k=tf.shape(y)[-1])
>>> sorted_y
<tf.Tensor: shape=(6,), dtype=float32, numpy=array([10.,  7.,  4.,  3.,  2.,  1.], dtype=float32)>
```

# Problem

The above means that when the "nearest" interpolation method is used for [`tfp.stats.percentile`](https://www.tensorflow.org/probability/api_docs/python/tfp/stats/percentile) this example gives

```python
>>> import tensorflow as tf
>>> import tensorflow_probability as tfp
>>> a = tf.constant([[10.0, 7.0, 4.0], [3.0, 2.0, 1.0]])
>>> q = 50
>>> tfp.stats.percentile(a, q, interpolation="nearest")
<tf.Tensor: shape=(), dtype=float32, numpy=4.0>
```

where the docstring would make one think that the result would be

```python
<tf.Tensor: shape=(), dtype=float32, numpy=3.0>
```

Additionally, as is outlined in [my Stack Overflow question](https://stackoverflow.com/questions/61024629/different-results-with-numpy-percentile-and-tensorflow-percentile-for-nearest), this behavior gives different results than NumPy's [`numpy.percentile`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.percentile.html), which is confusing given that `tfp.stats.percentile`'s docstring directly compares itself to `numpy.percentile`

https://github.com/tensorflow/probability/blob/356cfddef026b3339b8f2a81e600acd2ff8e22b4/tensorflow_probability/python/stats/quantiles.py#L428

# Proposed Change

Given that the it seems that `tfp.stats.percentile` intentionally gives congruent behavior to `numpy.percentile`, it would be good to [change the sort order used from _descending_ to _ascending_](https://stackoverflow.com/a/61035376/8931942).

So maybe something like changing

https://github.com/tensorflow/probability/blob/356cfddef026b3339b8f2a81e600acd2ff8e22b4/tensorflow_probability/python/stats/quantiles.py#L884

to

```python
sorted_, _ = tf.math.top_k(tensor, k=tf.shape(tensor)[-1])
sorted_ = tf.reverse(sorted_, [-1])
```

This would also make the code base more readable, as @langmore [notes in the comments of `_get_indices`](https://github.com/tensorflow/probability/blame/356cfddef026b3339b8f2a81e600acd2ff8e22b4/tensorflow_probability/python/stats/quantiles.py#L550-L552)

https://github.com/tensorflow/probability/blob/356cfddef026b3339b8f2a81e600acd2ff8e22b4/tensorflow_probability/python/stats/quantiles.py#L550-L552

# Additional Notes

This affects [`pyhf` Issue 815](https://github.com/scikit-hep/pyhf/issues/815) (I'm on the dev team).

cc @lukasheinrich @kratsg

matthewfeickert:
Resolves #864

This changes the behavior of `percentile` and `quantiles`, as previously the sort order had been descending. This is being changed as it differs in behavior from [NumPy's `percentile` function](https://numpy.org/doc/1.18/reference/generated/numpy.percentile.html) and the docstring describes behavior that is consistent with NumPy's implementation.

The main change occurs replacing `_sort_tensor` with a single call to [`tf.sort`](https://www.tensorflow.org/versions/r2.1/api_docs/python/tf/sort), which is then used in `percentile`. In `_sort_tensor` [`tf.math.top_k`](https://www.tensorflow.org/versions/r2.1/api_docs/python/tf/math/top_k) (which results in descending sort) was used. When swapped for [`tf.sort`](https://www.tensorflow.org/versions/r2.1/api_docs/python/tf/sort) (with the default values of `k=-1, direction='ASCENDING'` explicitly passed to make the behavior more clear to a reader) there is no longer a need to ensure that the output is reshaped properly. As `_sort_tensor` is now just a single function call it is then removed in favor of `tf.sort`. This then necessitates the change in definition of `frac_at_q_or_above` such that it is renamed to `frac_at_q_or_above` and the swap of `tf.math.ceil` and `tf.math.floor`.

This could have also been achieved with

```python
sorted_, _ = tf.math.top_k(tensor, k=tf.shape(tensor)[-1])
sorted_ = tf.reverse(sorted_, [-1])
```

but as there doesn't seem to be a clear reason to use `tf.math.top_k` over `tf.sort`, one function call seemed preferable to two.

A test is also added that validates the edge case discussed in Issue #864 that prompted this PR.

cc @lukasheinrich @kratsg

### DIFF
--- a/tensorflow_probability/python/stats/quantiles.py
+++ b/tensorflow_probability/python/stats/quantiles.py
@@ -552,7 +552,7 @@ def percentile(x,
 
     # Sort (in ascending order) everything which allows multiple calls to sort
     # only once (under the hood) and use CSE.
-    sorted_y = _sort_tensor(y)
+    sorted_y = tf.sort(y, axis=-1, direction='ASCENDING')
 
     d = tf.cast(tf.shape(y)[-1], tf.float64)
 
@@ -893,11 +893,3 @@ def _move_dims_to_flat_end(x, axis, x_ndims, right_end=True):
     full_shape = tf.concat(
         [other_shape, [-1]] if right_end else [[-1], other_shape], axis=0)
   return tf.reshape(x_permed, shape=full_shape)
-
-
-def _sort_tensor(tensor):
-  """Use `sort` to sort a `Tensor` in ascending order along
-     the last dimension."""
-  sorted_ = tf.sort(tensor, axis=-1, direction='ASCENDING')
-  tensorshape_util.set_shape(sorted_, tensor.shape)
-  return sorted_
--- a/tensorflow_probability/python/stats/quantiles.py
+++ b/tensorflow_probability/python/stats/quantiles.py
@@ -548,25 +548,22 @@ def percentile(x,
       axis = _make_static_axis_non_negative_list(axis, x_ndims)
       y = _move_dims_to_flat_end(x, axis, x_ndims, right_end=True)
 
-    frac_at_q_or_above = 1. - q / 100.
+    frac_at_q_or_below = q / 100.
 
-    # Sort everything, not just the top 'k' entries, which allows multiple calls
-    # to sort only once (under the hood) and use CSE.
+    # Sort (in ascending order) everything which allows multiple calls to sort
+    # only once (under the hood) and use CSE.
     sorted_y = _sort_tensor(y)
 
     d = tf.cast(tf.shape(y)[-1], tf.float64)
 
     def _get_indices(interp_type):
       """Get values of y at the indices implied by interp_type."""
-      # Note `lower` <--> ceiling.  Confusing, huh?  Due to the fact that
-      # _sort_tensor sorts highest to lowest, tf.ceil corresponds to the higher
-      # index, but the lower value of y!
       if interp_type == 'lower':
-        indices = tf.math.ceil((d - 1) * frac_at_q_or_above)
+        indices = tf.math.floor((d - 1) * frac_at_q_or_below)
       elif interp_type == 'higher':
-        indices = tf.floor((d - 1) * frac_at_q_or_above)
+        indices = tf.math.ceil((d - 1) * frac_at_q_or_below)
       elif interp_type == 'nearest':
-        indices = tf.round((d - 1) * frac_at_q_or_above)
+        indices = tf.round((d - 1) * frac_at_q_or_below)
       # d - 1 will be distinct from d in int32, but not necessarily double.
       # So clip to avoid out of bounds errors.
       return tf.clip_by_value(
@@ -583,8 +580,8 @@ def percentile(x,
       # Copy-paste of docstring on interpolation:
       # linear: i + (j - i) * fraction, where fraction is the fractional part
       # of the index surrounded by i and j.
-      larger_y_idx = _get_indices('lower')
-      exact_idx = (d - 1) * frac_at_q_or_above
+      larger_y_idx = _get_indices('higher')
+      exact_idx = (d - 1) * frac_at_q_or_below
       if preserve_gradients:
         # If q corresponds to a point in x, we will initially have
         # larger_y_idx == smaller_y_idx.
@@ -596,8 +593,8 @@ def percentile(x,
         larger_y_idx = tf.minimum(smaller_y_idx + 1, tf.shape(y)[-1] - 1)
         fraction = tf.cast(larger_y_idx, tf.float64) - exact_idx
       else:
-        smaller_y_idx = _get_indices('higher')
-        fraction = tf.math.ceil((d - 1) * frac_at_q_or_above) - exact_idx
+        smaller_y_idx = _get_indices('lower')
+        fraction = tf.math.ceil((d - 1) * frac_at_q_or_below) - exact_idx
 
       fraction = tf.cast(fraction, y.dtype)
       gathered_y = (
@@ -899,7 +896,8 @@ def _move_dims_to_flat_end(x, axis, x_ndims, right_end=True):
 
 
 def _sort_tensor(tensor):
-  """Use `top_k` to sort a `Tensor` along the last dimension."""
-  sorted_, _ = tf.math.top_k(tensor, k=tf.shape(tensor)[-1])
+  """Use `sort` to sort a `Tensor` in ascending order along
+     the last dimension."""
+  sorted_ = tf.sort(tensor, axis=-1, direction='ASCENDING')
   tensorshape_util.set_shape(sorted_, tensor.shape)
   return sorted_
--- a/tensorflow_probability/python/stats/quantiles_test.py
+++ b/tensorflow_probability/python/stats/quantiles_test.py
@@ -720,6 +720,15 @@ class PercentileTestWithNearestInterpolation(test_util.TestCase):
       self.assertAllEqual((), pct.shape)
       self.assertAllClose(expected_percentile, self.evaluate(pct))
 
+  def test_one_dim_numpy_docs_example(self):
+    x = [[10.0, 7.0, 4.0], [3.0, 2.0, 1.0]]
+    for q in [0, 10.1, 25.1, 49.9, 50.0, 50.1, 50.01, 89, 100]:
+      expected_percentile = np.percentile(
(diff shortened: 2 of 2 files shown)


matthewfeickert:
@jvdillon @csuter As you discussed this issue with me in Issue #864 I thought I would tag you here in the event that it might be easier for you two to review this PR. I have no idea how the TFP dev team does PR review so my apologies if there is a designated review shifter that will get around to this at some point.

I also don't see any CI getting kicked off(?) so if it would help to provide the test script that I've been running locally as well to check all the docstring examples for `percentile` and `quantiles` just let me know and I'm happy to provide that as well in the PR discussion.

tensorflow_probability/python/stats/quantiles.py
```diff
@@ -548,23 +548,20 @@ def percentile(x,
       axis = _make_static_axis_non_negative_list(axis, x_ndims)
       y = _move_dims_to_flat_end(x, axis, x_ndims, right_end=True)
 
-    frac_at_q_or_above = 1. - q / 100.
+    frac_at_q_or_above = q / 100.
 
-    # Sort everything, not just the top 'k' entries, which allows multiple calls
-    # to sort only once (under the hood) and use CSE.
+    # Sort (in ascending order) everything, not just the top 'k' entries, which
```
ColCarroll, line 553: you can delete `not just the top k entries`, since that was referring to the `_sort_tensor` implementation.

tensorflow_probability/python/stats/quantiles.py
```diff
@@ -548,23 +548,20 @@ def percentile(x,
       axis = _make_static_axis_non_negative_list(axis, x_ndims)
       y = _move_dims_to_flat_end(x, axis, x_ndims, right_end=True)
 
-    frac_at_q_or_above = 1. - q / 100.
+    frac_at_q_or_above = q / 100.
```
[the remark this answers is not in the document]
matthewfeickert: Good catch!

matthewfeickert:
@ColCarroll Thanks very much for the review! I've gone ahead and implemented those changes (also thanks for catching them!) and rebased them into the first commit when I rebased against upstream.

(My apologies if this is not your preferred method of changes during code review. I just figured since TFP doesn't do squash and merge PRs this would be cleaner to track the relevant changes for whenever it gets moved into a release than if it was spread across a few extra cleaning commits.)

tensorflow_probability/python/stats/quantiles.py
matthewfeickert: Done.

tensorflow_probability/python/stats/quantiles.py
```diff
@@ -899,7 +896,8 @@ def _move_dims_to_flat_end(x, axis, x_ndims, right_end=True):
 
 
 def _sort_tensor(tensor):
-  """Use `top_k` to sort a `Tensor` along the last dimension."""
-  sorted_, _ = tf.math.top_k(tensor, k=tf.shape(tensor)[-1])
+  """Use `sort` to sort a `Tensor` in ascending order along
+     the last dimension."""
+  sorted_ = tf.sort(tensor, axis=-1, direction='ASCENDING')
   tensorshape_util.set_shape(sorted_, tensor.shape)
```
csuter, line 902: i think the set_shape is no longer needed, since tf.sort will not change the shape (top_k would have the potential to, although we were abusing it slightly by setting k to the size of the input).

matthewfeickert: @csuter As this can then be done in 1 line now
```python
return tf.sort(tensor, axis=-1, direction='ASCENDING')
```
and there is only 1 function call to `_sort_tensor` should I just remove `_sort_tensor` altogether and instead have
```python
sorted_y = tf.sort(y, axis=-1, direction='ASCENDING')
```
?

csuter: yes, that sounds great. thanks!

matthewfeickert: Done.

matthewfeickert:
@srvasude Now that this has been reviewed and approved, can you explain what are the conditions that are required before this can get merged? I hope that I'm not coming across as impatient or demanding, I'm just unfamiliar with the how contributions to TFP slot in with development from the core dev team.

srvasude:
Hi, apologies for the slowness. The process is for us to pull in the request, and let it go through internal review before pushing. I was in the process of this, but was coming across some issues in our internal testing. In particular the percentile tests (specifically the test class PercentileTestWithLinearInterpolation in quantiles_test) is failing. Do you mind checking if any changes need to be made to quantile / quantile_test now that the sort order has been changed?

matthewfeickert:
> Hi, apologies for the slowness. The process is for us to pull in the request, and let it go through internal review before pushing. 

Nothing to apologize about, and this information is actually very helpful for understanding how things work in general. Thank you for sharing it.

> I was in the process of this, but was coming across some issues in our internal testing. In particular the percentile tests (specifically the test class PercentileTestWithLinearInterpolation in quantiles_test) is failing. Do you mind checking if any changes need to be made to quantile / quantile_test now that the sort order has been changed?

Sure. Luckily the changes that I've made are incredibly minor and only show up in the test case that I've added in the test that I added to `quantile_test`. I have a test script that I've been running locally as well that runs that test and additionally checks all the docstring examples for `percentile` and `quantiles`. I've run it each time I rebase this PR of `master` and things are passing. If this would be helpful I'm more than happy to provide it.

matthewfeickert:
@srvasude Now that the CI is back up and running I see that there is indeed a deviation between NumPy and my branch for `axis=None`. I've moved this PR into draft mode until I fix this. I'll ping you for review once I get this fixed and fully passing CI. I apologize for not catching this before I requested a review in the first place.

matthewfeickert:
@srvasude This is passing CI and is ready for review now. 

What was causing the errors in the tests was that I had missed that given my change in how 'lower' and 'upper' were defined, I forgot to swap their usage as well &mdash; sloppy mistake on my part. I'm sorry about that error, but I've rebased this in against `master` and with that fixed and this passing CI I think it is ready for final review.

matthewfeickert:
Excellent! Thank you very much @srvasude for the review and helping me get this PR done &mdash; I really appreciate it! Also thank you @csuter and @ColCarroll for your reviews and input which made all of this easier. :)

csuter:
Thanks for your contribution, Matthew. It's been a pleasure to work with you! :)

PR #922, opened 2020-05-11
Handle app flags in pytest session hook to fix CI

0x0badc0de:
Not sure how flags were passed under pytest before, I got `unrecognized argument` error even on master. So added `--absl-flag` option to `pytest` which can be used as
```
pytest tensorflow_probability/python/distributions/mixture_same_family_test.py --absl-flag='--vary_seed'
```

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -18,3 +18,9 @@ collect_ignore = [
     "setup.py",
     "tensorflow_probability/python/experimental/substrates/"
 ]
+
+def pytest_sessionstart(session):
+  from absl import app
+  # Unlike bazel, `pytest` doesn't invoke `tf.test.run()` (which parses flags), so
+  # for external developers using pytest we just parse the flags directly.
+  app._register_and_parse_flags_with_usage()  # pylint: disable=protected-access
--- a/conftest.py
+++ b/conftest.py
@@ -13,14 +13,26 @@
 # limitations under the License.
 # ============================================================================
 """Blacklist for pytest."""
+
+from absl import app
+
+
 collect_ignore = [
     "discussion/",
     "setup.py",
     "tensorflow_probability/python/experimental/substrates/"
 ]
 
-def pytest_sessionstart(session):
-  from absl import app
-  # Unlike bazel, `pytest` doesn't invoke `tf.test.run()` (which parses flags), so
-  # for external developers using pytest we just parse the flags directly.
-  app._register_and_parse_flags_with_usage()  # pylint: disable=protected-access
+def pytest_addoption(parser):
+  parser.addoption(
+      "--absl-flag",
+      action="append",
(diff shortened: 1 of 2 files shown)


conftest.py
```diff
@@ -13,8 +13,26 @@
...
+from absl import app
+
+
 collect_ignore = [
     "discussion/",
     "setup.py",
     "tensorflow_probability/python/experimental/substrates/"
 ]
+
+def pytest_addoption(parser):
```
srvasude, line 26: I'm not too familiar with pytest, so pardon the questions: --absl-flag doesn't seem to be used anywhere in our code base. Is this for a user running a test? And we need this because by default there isn't any passing of flags from pytest to the actual test itself?

tensorflow_probability/python/internal/test_util.py
```diff
@@ -154,7 +147,7 @@ def _one_part(*structure):
     if overall_exception:
       final_msg += str(overall_exception)
       raise AssertionError(final_msg)
-    elif exceptions_with_paths:
+    if exceptions_with_paths:
```
srvasude, line 150: Not needed?

conftest.py
0x0badc0de: Yes, this is for someone who wants to run pytest and pass some flags to absl. Relying on `sys.argv` doesn't seem to be an option since pytest errors on unknown flags. So I provided it a flag `--absl_flag` which can be used to pass flags down to absl.

tensorflow_probability/python/internal/test_util.py
0x0badc0de: Pylint was not happy with 'elif after raise', here some details https://stackoverflow.com/a/55632864

conftest.py
srvasude: Gotcha, this sounds like a good change then. This doesn't immediately change anything with CI though, correct?

0x0badc0de: I believe conftest doesn't affect CI at all. What fixed CI is removing code from test_util.

srvasude: Was it the elif or the pytest check (or both)?

0x0badc0de: Just pytest check. Basically just found first failing commit https://github.com/tensorflow/probability/commit/ec955788a3a5fdd31e4fccee9f9106a46ebcaf0c#diff-daa5b463df9f46faa41ebafd13aa3058 and reverted it. `pytest` was in `sys.modules` because hypothesis lib imported it so it's not really good approach to test if code is run within pytest. In this case pytest guys recommend using hooks like here https://docs.pytest.org/en/latest/example/simple.html#detect-if-running-from-within-a-pytest-run So I just found a hook which is invoked when all modules are imported hence all flags are defined and can be registered.

PR #928, opened 2020-05-13
lbfgs: support passing in `previous_optimizer_results`

### LINKED ISSUE #924: tfp.optimizer.lbfgs_minimize calls function for converged models
As far as I can tell, `lbfgs_minimize` continues to call the passed in `value_and_gradients`  function with the fullet set of parameters, even if some of the models in the batch have converged. When passing in a `tf.function` as `value_and_gradients`, I don't think this can be optimised away.

Would it be possible to have `value_and_gradients` only be called with the subset of the models that haven't converged yet?

We've previously implemented a batch optimizer that worked on a `(num_models, num_parameters)` tensor and call the underlying function on a slice of that tensor which included only the unconverged models.

Since unconverged models may not necessarily be contiguous, this worked by swapping out a newly converged model's parameters with the last one in the tensor and then decreased the slice by 1. As you do this, you can keep track of the permutation you've applied so far, and at the end unpermute the parameters.

I can see a few issues with meshing this with the current implementation of LBFGS. First, the `lbfgs_minimize` function accepts any number of batching dimensions, while slicing the way I described only works for a single batching dimension. Secondly, it would require the `value_and_gradients` function to accept a dynamic shape tensor (for the first dimension).

I don't know what the best way to implement this while keeping the curent API is. We could just have a boolean that controls this behaviour (and asserts the shape of the parameters is a matrix). 

Is this a good idea? Is there any desire for a PR implementing this?

**Edit** I submitted a PR #928 that proposes a minimal change to the API of `lbfgs_minimize` allowing optimization to be resumed after a single model converges. In between calls to `lbfgs_minimize` the user can do the swapping out of model as appropriate.

cristicbz:
This is a very small change that adds support for resuming lbfgs minimization if previously stopped by the stopping condition. If this design isn't the preferred one, happy to change the PR---I could also change the name of the argument to `experimental_initial_state` if you think it's not necessarily the way the API should look.

This means you can run the optimizer with `converged_any` and every time a single model converges from the batch, you can swap out the converged model and continue optimizing. For our problem, this gives us a 15-40% speed-up; we use batch optimisation for different hyperparameter settings, and strongly regularised models converge much quicker. Being able to stop evaluating them for the remaining iterations is an important optimisation.

Fixes #924 

Here's some pseudo-python ripped from our codebase as an example of how this can be used to speed up optimisation:

<details>

```python
num_models = ... # number of models simultaneously being optimized
full_parameters = tf.Variable(
  ...,
  shape=(num_models, num_model_parameters),
)
results = None

active_prefix = num_models
initial_parmaters = full_parameters.read_value()

total_permutation = np.arange(num_models)
while True:
    # Run LBFGS until one of the models converges.
    results = tfp.optimizer.lbfgs_minimize(
        compiled_loss,
        initial_parameters,
        stopping_condition=tfp.converged_any,
        max_iterations=max_iterations,
        initial_state=results,
    )
    # Set the initial parameters to `None` since in future runs, we'll
    # use the previous state instead .
    initial_parmeters = None

    # Update the `full_parameters` with their new positions
    full_parameters[:active_prefix].assign(results.position)

    # If we actually hit `max_iterations` or all models converged, then
    # break out of the loop.
    inactive = converged.numpy() | failed.numpy()
    if (
        results.num_iterations == max_iterations
        or inactive.all()
    ):
        # All models converged.
        break

    # Compute new permutation by swapping converged models with the
    # models at the end of the current `active_prefix`.
    new_permutation = np.arange(num_models)
    new_inactive_indices = list(inactive.nonzero()[0])
    new_prefix = active_prefix - len(new_inactive_indices)
    swap_with = list(range(new_prefix, active_prefix))
    new_permutation[
        new_inactive_indices + swap_with
    ] = new_permutation[swap_with + new_inactive_indices]

    # Apply new permutation to the parameters and optimizer state.
    full_parameters.assign(
        tf.gather(full_parameters.read_value(), new_permutation)
    )
    results = results._replace(
        converged=tf.zeros((new_prefix,), dtype=tf.bool),
        failed=tf.zeros((new_prefix,), dtype=tf.bool),
        position=tf.gather(
            results.position, new_permutation[:new_prefix], axis=0
        ),
        objective_value=tf.gather(
            results.objective_value,
            new_permutation[:new_prefix],
            axis=0,
        ),
        objective_gradient=tf.gather(
            results.objective_gradient,
            new_permutation[:new_prefix],
            axis=0,
        ),
        position_deltas=tf.gather(
            results.position_deltas,
            new_permutation[:new_prefix],
            axis=1,
        ),
        gradient_deltas=tf.gather(
            results.gradient_deltas,
            new_permutation[:new_prefix],
            axis=1,
        ),
    )

    # Keep track of overall permutation.
    total_permutation = total_permutation[new_permutation]
    active_prefix = new_prefix

inverse_total_permutation = np.arange(num_models)
for i_from, i_to in enumerate(total_permutation):
    inverse_total_permutation[i_to] = i_from

full_parameters.assign(
  tf.gather(parameters.read_value(), inverse_total_permutation),
)
```

</details>

### DIFF
--- a/tensorflow_probability/python/optimizer/lbfgs.py
+++ b/tensorflow_probability/python/optimizer/lbfgs.py
@@ -79,6 +79,7 @@ LBfgsOptimizerResults = collections.namedtuple(
 
 def minimize(value_and_gradients_function,
              initial_position,
+             previous_optimizer_results=None,
              num_correction_pairs=10,
              tolerance=1e-8,
              x_tolerance=0,
@@ -143,6 +144,13 @@ def minimize(value_and_gradients_function,
     initial_position: Real `Tensor` of shape `[..., n]`. The starting point, or
       points when using batching dimensions, of the search procedure. At these
       points the function value and the gradient norm should be finite.
+      Exactly one of `initial_position` and `previous_optimizer_results` can be
+      non-None.
+    previous_optimizer_results: An `LBfgsOptimizerResults` namedtuple to
+      intialize the optimizer state from, instead of an `initial_position`.
+      This can be passed in from a previous return value to resume optimization
+      with a different `stopping_condition`. Exactly one of `initial_position`
+      and `previous_optimizer_results` can be non-None.
     num_correction_pairs: Positive integer. Specifies the maximum number of
       (position_delta, gradient_delta) correction pairs to keep as implicit
       approximation of the Hessian matrix.
@@ -209,9 +217,19 @@ def minimize(value_and_gradients_function,
     stopping_condition = bfgs_utils.converged_all
 
   with tf.name_scope(name or 'minimize'):
-    initial_position = tf.convert_to_tensor(
-        initial_position, name='initial_position')
-    dtype = initial_position.dtype.base_dtype
+    if (initial_position is None) == (previous_optimizer_results is None):
+      raise ValueError(
+          'Exactly one of `initial_position` or '
+          '`previous_optimizer_results` may be specified.')
+
+    if initial_position is not None:
+      initial_position = tf.convert_to_tensor(
+          initial_position, name='initial_position')
+      dtype = initial_position.dtype.base_dtype
+
+    if previous_optimizer_results is not None:
+      dtype = previous_optimizer_results.position.dtype.base_dtype
+
     tolerance = tf.convert_to_tensor(
         tolerance, dtype=dtype, name='grad_tolerance')
     f_relative_tolerance = tf.convert_to_tensor(
@@ -253,10 +271,15 @@ def minimize(value_and_gradients_function,
               next_state.objective_gradient - current_state.objective_gradient))
       return [state_after_inv_hessian_update]
 
-    initial_state = _get_initial_state(value_and_gradients_function,
-                                       initial_position,
-                                       num_correction_pairs,
-                                       tolerance)
+    if previous_optimizer_results is None:
+      assert initial_position is not None
+      initial_state = _get_initial_state(value_and_gradients_function,
+                                         initial_position,
+                                         num_correction_pairs,
+                                         tolerance)
+    else:
+      initial_state = previous_optimizer_results
+
     return tf.while_loop(
         cond=_cond,
         body=_body,
--- a/tensorflow_probability/python/optimizer/lbfgs_test.py
+++ b/tensorflow_probability/python/optimizer/lbfgs_test.py
@@ -270,6 +270,72 @@ class LBfgsTest(test_util.TestCase):
       self.assertArrayNear(actual, expected, 1e-5)
     self.assertEqual(batch_results.num_objective_evaluations, 28)
 
+  def test_himmelblau_batch_any_resume_then_all(self):
+    @_make_val_and_grad_fn
+    def himmelblau(coord):
+      x, y = coord[..., 0], coord[..., 1]
+      return (x * x + y - 11) ** 2 + (x + y * y - 7) ** 2
+
+    dtype = 'float64'
+    starts = tf.constant([[1, 1],
+                          [-2, 2],
+                          [-1, -1],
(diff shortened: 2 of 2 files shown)


cristicbz:
@csuter Is there some way for me to get this PR looked at---I understand it may not be top priority, but I think it's a pretty small change that makes the lbfgs optimizer significantly more useful. Thanks!

tensorflow_probability/python/optimizer/lbfgs.py
```diff
@@ -79,6 +79,7 @@
 
 def minimize(value_and_gradients_function,
              initial_position,
+             initial_state=None,
```
csuter, line 82: let's call this `previous_optimizer_results`, which mirrors similar functionality and naming style in the MCMC package (see `previous_kernel_results` argument to `sample_chain`).

cristicbz: Done, thanks!

tensorflow_probability/python/optimizer/lbfgs.py
```diff
@@ -209,9 +216,23 @@ def quadratic_loss_and_gradient(x):
...
   with tf.name_scope(name or 'minimize'):
-    initial_position = tf.convert_to_tensor(
-        initial_position, name='initial_position')
-    dtype = initial_position.dtype.base_dtype
+    if initial_position is not None:
+      if initial_state is not None:
+        raise ValueError(
+            'If `initial_position` is not `None` then `initial_state` must be '
+            '`None`.'
+        )
```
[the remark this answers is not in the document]
cristicbz: Done.

tensorflow_probability/python/optimizer/lbfgs.py
```diff
@@ -209,9 +216,23 @@ def quadratic_loss_and_gradient(x):
     stopping_condition = bfgs_utils.converged_all
 
   with tf.name_scope(name or 'minimize'):
-    initial_position = tf.convert_to_tensor(
-        initial_position, name='initial_position')
-    dtype = initial_position.dtype.base_dtype
+    if initial_position is not None:
```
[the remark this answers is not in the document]
cristicbz: I think you mean `==` rather than `!=` above, but yeah you're right it's cleaner, done.

cristicbz:
Thanks for the review @csuter , I amended as per your comments, please take a look!

tensorflow_probability/python/optimizer/lbfgs.py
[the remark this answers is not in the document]
csuter: indeed -- thanks! :)

tensorflow_probability/python/optimizer/lbfgs.py
```diff
@@ -143,6 +144,13 @@ def quadratic_loss_and_gradient(x):
     initial_position: Real `Tensor` of shape `[..., n]`. The starting point, or
       points when using batching dimensions, of the search procedure. At these
       points the function value and the gradient norm should be finite.
+      Exactly one of `initial_position` and `previous_optimizer_results` can be
+      non-None.
+    previous_optimizer_results: An `LBfgsOptimizerResults` namedtuple to
+      intialize the opitimizer state from, instead of an `initial_position`.
```
csuter, line 150: typo: s/opitimizer/optimizer/

cristicbz: oops, thanks!

cristicbz:
@csuter Thanks for the approval! I fixed that last typo and tests passed, so I think this is good to go?

csuter:
Thanks for the contribution @cristicbz, and thanks for your patience in the review process. Sorry we took so long! The change is checked in to our internal version control and should show up here momentarily.

Issue #944, opened 2020-05-21
TypeError when deepcopying models that reference (and have sampled from) LogNormal


elvijs:
Our training procedure involves deepcopying a model. This fails when we have a parameter with `tfp.distributions.LogNormal` prior that is used early in training, but is subsequently disabled. The failure is `TypeError: can't pickle HashableWeakRef objects`. 

Note that `tfp.distributions.Normal` does not fail in the same way.

## To reproduce

I've attempted to trim away all irrelevant bits from our full training pipeline. It looks like the root cause relates to sampling from `tfp.distributions.LogNormal` (although anecdotally, we've seen similar issues with some bijectors).

**Minimal reproduction**
```python
import copy
import tensorflow as tf
import tensorflow_probability as tfp


def boom() -> None:
    m = tf.Module()
    m.prior = tfp.distributions.LogNormal(0.0, 1.0)  # error
    # m.prior = tfp.distributions.Normal(0.0, 1.0)  # no error

    @tf.function  # no error without the decorator
    def _sample() -> tf.Tensor:
        m.prior.sample()
        # it looks like the sampling pollutes the graph - if you comment out the line above,
        # then we don't explode

    _sample()
    # now that we've done some sampling at model initialisation, 
    # we actually want to disable the priors for training
    m.prior = None  # if you comment out this line, then we don't explode

    copy.deepcopy(m)


if __name__ == '__main__':
    boom()
```

**Stack trace**:
```Traceback (most recent call last):
  File "/home/elvijs/.PyCharmCE2019.2/config/scratches/scratch_7.py", line 27, in <module>
    boom()
  File "/home/elvijs/.PyCharmCE2019.2/config/scratches/scratch_7.py", line 23, in boom
    copy.deepcopy(m)
  File "/home/elvijs/.virtualenvs/automl_res/lib/python3.7/copy.py", line 180, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/home/elvijs/.virtualenvs/automl_res/lib/python3.7/copy.py", line 280, in _reconstruct
    state = deepcopy(state, memo)
  File "/home/elvijs/.virtualenvs/automl_res/lib/python3.7/copy.py", line 150, in deepcopy
    y = copier(x, memo)
  File "/home/elvijs/.virtualenvs/automl_res/lib/python3.7/copy.py", line 240, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
[...]

Issue #547, opened 2019-09-05
Deep copy of bijectors


awav:
Hello all,

I found that `deepcopy` doesn't work on bijector classes. Actually, I was surprised that deepcopy *just* works in TF2.0 out of the box for variables and tensors, but this is a different story.

Here is a MWE:

```python
import tensorflow_probability as tfp
import tensorflow as tf
from copy import deepcopy

class A(tf.Module):
    def __init__(self):
        super().__init__()
        self.bijector = tfp.bijectors.Softplus()


a = A()
deepcopy(a)

x = tf.convert_to_tensor(-1.0)
y = a.bijector.forward(x)

try:
    deepcopy(a)
except:
    print(f'bijector._from_x is NOT empty: {a.bijector._from_x}')
```

After the first call, the `deepcopy` works fine, but once you do *forward* or *inverse* the `deepcopy` throws an error "_TypeError: can't pickle HashableWeakRef objects_".

The bijector class is adjusted to TF1.0, where tensors are not actual results of operations, but a computation nodes in the graph. The bijector class has the cache dictionary inside where a key is a tensor or a numpy object and a value is a result of the `forward` or `inverse` operations. Such a simple solution helps against graph over-growing, in other words it is a computation graph optimization technique. The key of the cache have `tensorflow_probability.python.bijectors.bijector.HashableWeakRef` type , which inherits from the `weakref`, which in turn is not compatible with `deepcopy`.

In TF2.0 this cache related code is redundant. **Moreover, it is a potential leak for TF2.0 case.** Whenever we compute *forward* or *inverse*, the bijector saves the input tensor and the result in the cache. Check example below:

```python
# ... continue from previous MWE
[a.bijector.forward(tf.random.normal((1,))) for _ in range(100)]
assert a.bijector._from_x >= 100
```

My opinion that this code should be just removed. There is no benefit from having it in TF2.0. It may stay because of backward compatibility, but in that case, the issue with `deepcopy` should be fixed and cache shouldn't grow linearly with number of calls to *forward* and *inverse*.

Thanks

brianwa84:
The references in there are weak, so there *shouldn't* be a memory leak even in 2.0/eager land.
I agree it'd be nice if deepcopy would work, feel free to send a PR that just skips the _from_x/_from_y fields (create new ones for the copy).

emilyfertig:
Closing since `deepcopy` now works for bijectors, following global caching changes.

PR #947, opened 2020-05-23
Deepcopy of weak structure references

awav:
This PR addresses issues at https://github.com/tensorflow/probability/issues/547 and https://github.com/tensorflow/probability/issues/944 by defining [`__deepcopy__`](https://docs.python.org/3/library/copy.html#copy.deepcopy) method of the `WeakStructureRef` class, that holds a structure with `HashableWeakRef`s. However, there are many nuances about `weakrefs`, and it seems impossible to provide complete deepcopy support.

The current version of deepcopy works well with tensors - simply because deepcopy returns the same tensor. When we apply deepcopy on a numpy array we get a new version of it and the reference count for the copy is one. For a deep copy of a weak structure reference, we also wrap the new numpy array with `weakref` and its reference count inside `__deepcopy__` is one. The numpy array lifetime is limited to the scope of the `__deepcopy__` method. Once we exit `__deepcopy__`, the reference count is down to 0, and the `weakref` changes its status to "dead".

```python
import tensorflow as tf
import numpy as np

ndarray = np.zeros([2, 3])
tensor = tf.constant([1., 2., 3.])
ref1 = WeakStructRef(tensor, subkey='b')
ref2 = WeakStructRef(ndarray, subkey='a')

ref1_copy = deepcopy(ref1)  # new weak reference, but the same tensor
ref2_copy = deepcopy(ref2)  # new weak reference and dead status

# Trick to make a weak reference alive
memo = {}  # retains a copy of the ndarray, therefore the reference count is 1
ref2_copy = deepcopy(ref2, memo)  # alive weak reference
```

The alternative method would be: never copy a cache and leave it empty.

Offtopic: I would appreciate if somebody could explain how to setup vscode or pycharm to run tests from `tensorflow_probability`. Thanks.


csuter:
Thanks for working on this fix @awav! I think a trivial __deepcopy__ implementation that doesn't try to persist the cache contents is probably the best approach here. Copying in-process is likely rare, and in the primary problematic use case of serialization/deserialization I think re-animating weakrefs in the cache will be unhelpful (indeed, as you note, they will just be cleaned up when the copying is over).

For testing, we use bazel; I'm not sure about IDE integration, but the scripts in the top-level testing/ dir may be useful guides. Note the run_tests.sh script is complicated by the fact that it 1. is written to run on Travis and 2. is written to shard tests (taking size into account) to improve overall latency on Travis. Anyway, the commands you'd need to run are suggested there -- note the bazel flags, some of which are crucial.

awav:
Hello @csuter. I made a custom `__deepcopy__` method for bijectors. Other solutions are very fragile. The bijector's cache contains function closures. I added a test case, but Travis fails in a different place and the error is not obvious. Could you take a look?

nathanhack:
@csuter, @awav  is there any status on this feature?

awav:
Hi, @nathanhack. The PR is more or less done, it is up to the TFP team to merge it or not, or give feedback :)

awav:
@csuter, What's the status of the review for this PR? Can someone take a look and tell what needs to be done?

I keep getting "hashable" errors at deep copying TFP objects. They are  especially hard to handle because of differences between versions 0.10 and 0.11

st--:
I'd be really keen for this to get merged. As it stands, we need to maintain very implementation-specific workarounds in our downstream code in GPflow that we have to keep updating (https://github.com/GPflow/GPflow/pull/1553, https://github.com/GPflow/GPflow/pull/1574) to ensure continued compatibility with tensorflow-probability...

csuter:
Apologies for the lapse @awav -- I went on parental leave at the end of June and this totally slipped off my radar. Thanks for keeping up with it, and tracking the recent (substantial) changes to bijector fundamentals. This basically LGTM; I'm going to ask @emilyfertig to have a pass as well.

csuter (commented):
Sorry thought I sent these comments earlier!

emilyfertig:
Hi @awav -- Thanks for the fix to this, and sorry we weren't on top of it over the summer! 

We've made a major change to bijector caching since the 0.11 release, where bijectors now access a shared global cache, instead of maintaining instance-level caches. With these changes, in the nightly build, I can't seem to reproduce any deepcopy/pickle errors (including the examples in #547 and #944). 

If the deepcopy issue is indeed resolved at head (and will go into the 0.12 release, probably within a couple months, following TF's release), is there anything you'd still want merged? Or would it be sufficient if we add a unit test to confirm that deepcopy works on bijectors?

(Also, related to the example above, I'm curious as to when Numpy arrays are wrapped in `WeakStructRef`s -- Bijector methods call `tf.convert_to_tensor` on their inputs, so in TFP library code, I think `WeakStructRef` will only wrap `Tensor`s, unless you're using TFP's Numpy backend).

emilyfertig:
Closing since `deepcopy` now works for bijectors, following global caching changes.

PR #954, opened 2020-05-29
Add some helper scripts for running tests and use them in travis tests.

csuter:
Specifically:
  - add a script wrapping bazel test invocation, setting commonly used flags
  - add a sourceable alias definition script, which aliases `tfp_test` to point to the tfp_test.sh script
  - add a script to install dependencies, along with checks for common pitfalls and helpful error messages.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -62,21 +62,69 @@ for recommended test setup.
 Unit tests ensure new features (a) work correctly and (b) guard against future
 breaking changes (thus lower maintenance costs).
 
-To run existing unit-tests on CPU, use the command:
+### Setup
 
+#### `bazel`
+
+We use [`bazel`](https://bazel.build/) to manage building, packaging, and
+testing TFP. You'll need to install `bazel` before running our tests (we have
+recently added some experimental support for running some tests with pytest, but
+for a variety of reasons this will probably never work 100%). See instructions
+[here](https://docs.bazel.build/versions/3.2.0/install-os-x.html) on installing
+`bazel`.
+
+#### `virtualenv`
+
+We strongly recommend running unit tests in an active
+[virtualenv](https://virtualenv.pypa.io/en/latest/). Doing so requires some
+extra bazel flags, so we created a wrapper script, which we suggest using. An
+example invocation (presumed to run from the root of the TFP repo:
+
+#### Helper scripts
 
 ```shell
-bazel test --copt=-O3 --copt=-march=native //tensorflow_probability/...
+# Run all TFP tests.
+./testing/tfp_test.sh //tensorflow_probability/...
 ```
 
-from the root of the `tensorflow_probability` repository. To run tests on GPU,
-you just need to ensure the GPU-enabled version of TensorFlow is installed.
-However, you will also need to include the flag `--jobs=1`, since by default
-Bazel will run many tests in parallel, and each one will try to claim all the
-GPU memory:
+See comments at the top of the script for more info.
+
+For convenience, also consider sourcing the following script to alias `tfp_test`
+to the above script:
 
 ```shell
-bazel test --jobs=1 --copt=-O3 --copt=-march=native //tensorflow_probability/...
+source ./testing/define_testing_alias.sh
+# Run all TFP tests.
+tfp_test //tensorflow_probability/...
+```
+
+#### Dependencies
+
+To run the unit tests, you'll need several packages installed (again, we
+strongly recommend you work in a virtualenv). We include a script to do this for
+you, which also does some sanity checks on the environtment:
+
+```shell
+./testing/install_test_dependencies.sh
+```
+
+See the
+[header comments in that script](https://github.com/tensorflow/probability/blob/master/testing/install_test_dependencies.sh)
+for more details.
+
+### Additional considerations
+
+As of early 2020, tensorflow and tf-nightly include GPU support by default,
+which means if you have a GPU installed and run tests with the default
+`tf-nightly` pip package, tests will try to run using the GPU. To avoid this,
+the dependency install script installs tf-nightly-cpu by default. If you *want*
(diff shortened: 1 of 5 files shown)


testing/run_tests.sh
```diff
@@ -20,6 +20,9 @@ set -x  # print commands as they are executed
 set -e  # fail and exit on any command erroring
 set -u  # fail and exit on any undefined variable reference
 
+# Get the absolute path to the directory containing this script.
+DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
```
SiegeLordEx, line 24: Can we just require running the script from the root directory?

testing/define_testing_alias.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
SiegeLordEx, line 1: I don't see the purpose of this. It doesn't save the user any lines or any characters.

```bash
./testing/define_testing_alias.sh
tfp_test
```

vs

```
./testing/tfp_test.sh
```

It also feels too presumptive of people's bash habits.

testing/tfp_test.sh
```diff
@@ -0,0 +1,50 @@
...
+# recommend) requires some special flags. Additionally, some tests are defined
+# but known not to be working and are therefore tagged as such; using an
+# additional flag ensures such tests are skipped. This script wraps the `bazel`
+# command and ensures these flags are set. Arguments to this test are passed
+# along to `bazel test` after the aforementioned flags
+#
+# ## Example 1: Run all tests
+#
+# ```bash
+#  $ tfp_test.sh //tensorflow_probability/...
```
SiegeLordEx, line 13: ./testing/test.sh

testing/define_testing_alias.sh
csuter: Well, it's not required to use. But if you're cd'ed several dirs deep into the code then you need a bunch of ../'s to find the testing dir and the script. i'm assuming one would define the alias once and then use it a bunch of times in a session, so in the long run it'd save many typed chars. If this is strongly opposed I can remove it but I'd certainly like to have it around myself.

CONTRIBUTING.md
```diff
@@ -62,21 +62,69 @@ for recommended test setup.
...
+for more details.
+
+### Additional considerations
+
+As of early 2020, tensorflow and tf-nightly include GPU support by default,
+which means if you have a GPU installed locally and run tests with the default
+`tf-nightly` pip package, tests will try to run using the GPU. To avoid this,
+you can pass the --cpu_only flag to `testing/install_test_dependencies.sh`.
+
+If you *want* to use the GPU to run tests, you will need to include the flag
```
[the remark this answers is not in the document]
csuter: That's a fair point, but actually we only disable tests tagged with 'gpu' (or the other long nvidia one). The distinction here is what the installed TF package will *try* to do, which doesn't care about test tags at all. I.e., with tf-nightly and a GPU present, all tests will try to run on the GPU by default.

I thought about making tf-nightly-cpu the default dependency and changing the dep install script to require a --enable_gpu flag in order to install the other one. WDYT?

CONTRIBUTING.md
```diff
@@ -62,21 +62,69 @@ for recommended test setup.
...
+`bazel`.
+
+#### `virtualenv`
+
+We strongly recommend running unit tests in an active
+[virtualenv](https://virtualenv.pypa.io/en/latest/). Doing so requires some
+extra bazel flags, so we created a wrapper script, which we suggest using. An
+example invocation (presumed to run from the root of the TFP repo:
+
+#### Helper scripts
```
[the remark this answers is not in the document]
csuter: the section on CI discusses linting. you're saying we should add text here on how to run it locally? that'd be good. I don't know how, though, off the top of my head, so I wouldn't mind deferring that to a separate change.

testing/run_tests.sh
csuter: well, this is just for travis so it probably doesn't matter much, but i'd rather be explicit and safe than make the script more fragiel w.r.t. location.

testing/define_testing_alias.sh
csuter: ie, with this i can do

```
source ./testing/define_testing_alias.sh
cd tensorflow_probability/python/mcmc
tfp_test :hmc_test
# ...
tfp_test :dual_averaging_step_size_test
# etc
```

vs

```
cd tensorflow_probability/python/mcmc
../../../testing/tfp_test.sh :hmc_test
# ...
../../../testing/tfp_test.sh :dual_averaging_step_size_test
# etc
```

testing/tfp_test.sh
csuter: Done

CONTRIBUTING.md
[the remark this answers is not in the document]
SiegeLordEx: Looks like it needs the same treatment of splitting some Travis things out first. Leaving it to a different change sounds fine to me.

testing/define_testing_alias.sh
SiegeLordEx: Just don't cd into the subdirectory and liberally use bash history. Like I said, different people have different habits. I guess it's fine to leave this here, the verbiage in the readme is non-committal enough to not make it seem that this is mandatory.

csuter: Thanks, Pavel. +@ColCarroll to get another perspective on whether it's weird to have this around. I don't feel super strongly, but it would be convenient for my usual workflow/habits.

CONTRIBUTING.md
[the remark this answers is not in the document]
csuter: FYI I switched the default to use tf-nightly-cpu, changed the flag to --enable_gpu instead of --cpu_only, and added additional sanity checks around each possible setting. I also updated the verbiage in CONTRIBUTING.md

PR #960, opened 2020-06-02
small optimization for `batch_evaluate_objective=True` in Nelder Mead optimizer

jeffpollock9:
Hi,

I have an expensive objective function which I am trying to optimize with Nelder Mead, it supports batching and stepping through the code I noticed we could take advantage of that a little more with this change, I think.

In short, previously when `objective_at_initial_vertex is None` then the code would call `objective_function` then `_evaluate_objective_multiple`. I've changed it to make a single call to `_evaluate_objective_multiple` with a bigger batch which I think is almost always preferred.

Hope it is useful, let me know if there is anything else I can do with this.

Jeff


jeffpollock9:
Hi @saxena-ashish-g, sorry to bother you but is there any chance you can have a quick look at this and let me know if there is anything else I can do to get this merged or if it should just be closed? Tagging you since as far as I can tell you wrote most of the code in this area. Thanks.

axch:
Sorry for the long delay in responding.  These changes look good to me; I'll see about getting them merged.

PR #966, opened 2020-06-10
TransformedTransitionKernel call with missing `target_log_prob_fn` bugfix

rupei:
Currently, calling `TransformedTransitionKernel` on an `inner_kernel` without a `target_log_prob_fn` will result in an error. For example, calling `TransformedTransitionKernel` on a `SimpleStepSizeAdaptation` object as the inner kernel.

Akin to how `SimpleStepSizeAdaptation` edits `step_size`, `TransformedTransitionKernel` now iterates through `inner_kernel`s until it finds a valid one, then propagates that change. Tests are included, docstring edited to reflect the change, and a `ValueError` is thrown if no `inner_kernel` satisfies the condition.

### DIFF
--- a/tensorflow_probability/python/mcmc/kernel.py
+++ b/tensorflow_probability/python/mcmc/kernel.py
@@ -90,16 +90,16 @@ class TransitionKernel(object):
 
     Args:
       override_parameter_kwargs: Python String/value `dictionary` of
-      initialization arguments to override with new values.
+        initialization arguments to override with new values.
 
     Returns:
-      new_kernel: A `TransitionKernel` object of same type as `self`,
-      initialized with override_parameter_kwargs
+      new_kernel: `TransitionKernel` object of same type as `self`,
+        initialized with the union of self.parameters and
+        override_parameter_kwargs, with any shared keys overridden by the
+        value of override_parameter_kwargs, i.e.,
+        `dict(self.parameters, **override_parameters_kwargs)`.
     """
     parameters = dict(self.parameters, **override_parameter_kwargs)
     new_kernel = type(self)(**parameters)
-    # pylint: disable=protected-access
-    new_kernel._parameters = parameters
-    # pylint: enable=protected-access
-    new_kernel = type(self)(**parameters)
+    new_kernel._parameters = parameters # pylint: disable=protected-access
     return new_kernel
--- a/tensorflow_probability/python/mcmc/kernel.py
+++ b/tensorflow_probability/python/mcmc/kernel.py
@@ -84,3 +84,22 @@ class TransitionKernel(object):
         `Tensor`s representing internal calculations made within this function.
     """
     return []
+
+  def copy(self, **override_parameter_kwargs):
+    """Non-destructively creates a deep copy of the kernel
+
+    Args:
+      override_parameter_kwargs: Python String/value `dictionary` of
+      initialization arguments to override with new values.
+
+    Returns:
+      new_kernel: A `TransitionKernel` object of same type as `self`,
+      initialized with override_parameter_kwargs
+    """
+    parameters = dict(self.parameters, **override_parameter_kwargs)
+    new_kernel = type(self)(**parameters)
+    # pylint: disable=protected-access
+    new_kernel._parameters = parameters
+    # pylint: enable=protected-access
+    new_kernel = type(self)(**parameters)
+    return new_kernel
--- a/tensorflow_probability/python/mcmc/kernel.py
+++ b/tensorflow_probability/python/mcmc/kernel.py
@@ -101,5 +101,4 @@ class TransitionKernel(object):
     """
     parameters = dict(self.parameters, **override_parameter_kwargs)
     new_kernel = type(self)(**parameters)
-    new_kernel._parameters = parameters # pylint: disable=protected-access
     return new_kernel
--- a/tensorflow_probability/python/mcmc/transformed_kernel.py
+++ b/tensorflow_probability/python/mcmc/transformed_kernel.py
@@ -105,8 +105,7 @@ def make_transformed_log_prob(
     return tlp + sum(ldj_fn(state_parts, event_ndims))
   return transformed_log_prob_fn
 
-
-def target_log_prob_getter(kernel):
+def _make_kernel_stack(kernel):
   kernel_stack = [kernel]
   while 'target_log_prob_fn' not in kernel.parameters:
     if 'inner_kernel' not in kernel.parameters:
@@ -114,14 +113,18 @@ def target_log_prob_getter(kernel):
                        '`target_log_prob_fn`."')
     kernel = kernel.inner_kernel
     kernel_stack.append(kernel)
-  return kernel.parameters['target_log_prob_fn'], kernel_stack
+  return kernel_stack
 
+def get_target_log_prob(kernel):
+  kernel_stack = _make_kernel_stack(kernel)
+  last_kernel = kernel_stack[len(kernel_stack) - 1]
+  return last_kernel.parameters['target_log_prob_fn']
 
-def target_log_prob_setter(kernel_stack, new_target_log_prob):
-  # edit the appropriate kernel with the new log_prob_fn
-  last_kernel = kernel_stack.pop()
-  last_kernel_kwargs = last_kernel.parameters.copy()
-  last_kernel_kwargs.update(target_log_prob_fn=new_target_log_prob)
+def update_target_log_prob(kernel, new_target_log_prob):
+  kernel_stack = _make_kernel_stack(kernel)
+  # get the last kernel and update to the new_target_log_prob
+  last_kernel = kernel_stack.pop().copy(target_log_prob_fn=new_target_log_prob)
+  last_kernel_kwargs = last_kernel.parameters
 
   # propogate upwards
   while kernel_stack:
@@ -130,8 +133,7 @@ def target_log_prob_setter(kernel_stack, new_target_log_prob):
       curr_kernel._inner_kernel = type(last_kernel)(**last_kernel_kwargs)
     last_kernel = curr_kernel
     last_kernel_kwargs = last_kernel.parameters
-
-  return last_kernel, last_kernel_kwargs
+  return last_kernel
 
 class TransformedTransitionKernel(kernel_base.TransitionKernel):
   """TransformedTransitionKernel applies a bijector to the MCMC's state space.
@@ -154,26 +156,6 @@ class TransformedTransitionKernel(kernel_base.TransitionKernel):
   `tfp.bijectors.Affine`, `tfp.bijectors.RealNVP`, etc. with transition kernels
   `tfp.mcmc.HamiltonianMonteCarlo`, `tfp.mcmc.RandomWalkMetropolis`,
   etc.
-
-  If the provided `inner_kernel` doesn't have a `target_log_prob_fn`,
-  `TransformedTransitionKernel` will iteratively look through deeper layers of
-  kernels to find one that does. It will then apply the transformation to that
-  `target_log_prob_fn` and propogate those changes if necessary.
-
-  For example, in the following example, since the `SimpleStepSizeAdaptation`
-  object has no `target_log_prob_fn`, that of the `HamiltonianMonteCarlo` would
-  be used.
-
-  ```python
-  tfp.mcmc.TransformedTransitionKernel(
-    inner_kernel=tfp.mcmc.SimpleStepSizeAdaptation(
-      inner_kernel=tfp.mcmc.HamiltonianMonteCarlo(
-        ... # doesn't matter
-      ),
-      num_adaptation_steps=9)
-    bijector=tfb.Identity()))
-  ```
-
 
   #### Mathematical Details
 
@@ -264,6 +246,32 @@ class TransformedTransitionKernel(kernel_base.TransitionKernel):
       axis=0)
   ```
 
+  #### Detailed API Contract
+
+  `TransformedTransitionKernel` operates on a `target_log_prob_fn` field, which
+  is expected to be an attribute of some `inner_kernel` (which may or may not
+  be the most immediate). If the provided `inner_kernel` doesn't have a
+  `target_log_prob_fn`, `TransformedTransitionKernel` will iteratively look
+  through deeper layers of kernels to find one that does, stopping at the first
+  one it finds (not the deepest). It will then apply the transformation to
+  that `target_log_prob_fn` and propogate those changes upwards, such that all
+  kernels that wrap around the updated one reflect that change. If it turns out
+  that no `inner_kernel` satisfies that condition, a `ValueError` will be
+  raised.
+
+  In the following example, since the `SimpleStepSizeAdaptation` object has no
+  `target_log_prob_fn`, that of the `HamiltonianMonteCarlo` would be used.
+
+  ```python
+  tfp.mcmc.TransformedTransitionKernel(
+    inner_kernel=tfp.mcmc.SimpleStepSizeAdaptation(
+      inner_kernel=tfp.mcmc.HamiltonianMonteCarlo(
+        ... # doesn't matter
+      ),
+      num_adaptation_steps=9)
+    bijector=tfb.Identity()))
+  ```
+
   #### References
 
   [1]: Matthew Parno and Youssef Marzouk. Transport map accelerated Markov chain
@@ -297,7 +305,7 @@ class TransformedTransitionKernel(kernel_base.TransitionKernel):
         inner_kernel=inner_kernel,
         bijector=bijector,
         name=name or 'transformed_kernel')
-    target_log_prob_fn, kernel_stack = target_log_prob_getter(self)
+    target_log_prob_fn = get_target_log_prob(self)
     new_target_log_prob = make_transformed_log_prob(
         target_log_prob_fn,
         bijector,
@@ -305,7 +313,8 @@ class TransformedTransitionKernel(kernel_base.TransitionKernel):
         # TODO(b/72831017): Disable caching until gradient linkage
         # generally works.
         enable_bijector_caching=False)
-    target_log_prob_setter(kernel_stack, new_target_log_prob)
+    self._inner_kernel = update_target_log_prob(inner_kernel,
+                                                new_target_log_prob)
     # Prebuild `_forward_transform` which is used by `one_step`.
     self._transform_unconstrained_to_target_support = make_transform_fn(
         bijector, direction='forward')
--- a/tensorflow_probability/python/mcmc/transformed_kernel.py
+++ b/tensorflow_probability/python/mcmc/transformed_kernel.py
@@ -33,11 +33,14 @@ __all__ = [
     'TransformedTransitionKernel',
 ]
 
-TransformedTransitionKernelResults = collections.namedtuple(
-    'TransformedTransitionKernelResults', [
-        'transformed_state',
-        'inner_results',
-    ])
+class TransformedTransitionKernelResults(
+    mcmc_util.PrettyNamedTupleMixin,
+    collections.namedtuple('TransformedTransitionKernelResults',
+                           ['transformed_state',
+                            'inner_results',
+                            ])):
+  """Internal state and diagnostics for Transformed kernel."""
+  __slots__ = ()
 
 
 def make_log_det_jacobian_fn(bijector, direction):
@@ -116,24 +119,56 @@ def _make_kernel_stack(kernel):
   return kernel_stack
 
 def get_target_log_prob(kernel):
+  """Fetches the to-be-updated `target_log_prob_fn`.
+
+  Args:
+    kernel: outermost `TransitionKernel` object. The iterative search
+      will go through all nested kernels, including this one, until it
+      finds a `target_log_prob_fn`.
+
+  Returns:
+    target_log_prob_fn: The first `target_log_prob_fn` attribute found.
+      This will be both the `target_log_prob_fn` of the shallowest kernel
+      that contains one, and the one that will be updated by the
+      transformation.
+
+  Raises:
+    ValueError: if none of the `inner_kernel`s contains a "target_log_prob".
+  """
   kernel_stack = _make_kernel_stack(kernel)
-  last_kernel = kernel_stack[len(kernel_stack) - 1]
-  return last_kernel.parameters['target_log_prob_fn']
+  target_log_prob_fn = kernel_stack[-1].parameters['target_log_prob_fn']
+  return target_log_prob_fn
 
 def update_target_log_prob(kernel, new_target_log_prob):
-  kernel_stack = _make_kernel_stack(kernel)
-  # get the last kernel and update to the new_target_log_prob
-  last_kernel = kernel_stack.pop().copy(target_log_prob_fn=new_target_log_prob)
-  last_kernel_kwargs = last_kernel.parameters
+  """Updates a (potentially nested) `TransitionKernel`'s `target_log_prob_fn`.
 
-  # propogate upwards
+  Args:
+    kernel: outermost `TransitionKernel` object. The iterative search
+      will go through all nested kernels, including this one, until it
+      finds a `target_log_prob_fn` to update.
+    new_target_log_prob: Python callable that represents the intended
+      new `target_log_prob_fn`.
+
+  Returns:
+    prev_kernel: updated `TransitionKernel` object. If nested, the
+    `target_log_prob_fn` was updated in the shallowest kernel with such a
+    field. All kernels that wrap around it were appropriately reinstantiated
+    to reflect the change.
+
+  Raises:
+    ValueError: if none of the `inner_kernel`s contains a "target_log_prob".
+  """
+  kernel_stack = _make_kernel_stack(kernel)
+  # update to the new_target_log_prob
+  prev_kernel = kernel_stack.pop().copy(target_log_prob_fn=new_target_log_prob)
+
+  # propagate upwards
   while kernel_stack:
     curr_kernel = kernel_stack.pop()
     with deprecation.silence():
-      curr_kernel._inner_kernel = type(last_kernel)(**last_kernel_kwargs)
-    last_kernel = curr_kernel
-    last_kernel_kwargs = last_kernel.parameters
-  return last_kernel
+      curr_kernel._inner_kernel = type(prev_kernel)(**prev_kernel.parameters)
+    prev_kernel = curr_kernel
+  return prev_kernel
 
 class TransformedTransitionKernel(kernel_base.TransitionKernel):
   """TransformedTransitionKernel applies a bijector to the MCMC's state space.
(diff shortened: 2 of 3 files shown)


tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -33,15 +33,11 @@
     'TransformedTransitionKernel',
 ]
 
-
```
axch, line 36: Please rebase on top of the latest master; this diff looks like it's undoing a change made in parallel.

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -132,6 +155,26 @@ class TransformedTransitionKernel(kernel_base.TransitionKernel):
   `tfp.mcmc.HamiltonianMonteCarlo`, `tfp.mcmc.RandomWalkMetropolis`,
   etc.
 
+  If the provided `inner_kernel` doesn't have a `target_log_prob_fn`,
```
axch, line 158: I think this information should be less prominent in the docstring.  It would be better as its own section called something like "Detailed API Contract" or something, probably near the end, between the examples and the references.  It should also be more complete: say at the beginning that `TransformedTransitionKernel` operates on a `target_log_prob_fn` field, say explicitly that it looks for `inner_kernel` fields, etc. The goal for technical documentation like this is to make it so that someone who carefully reads the docstring but doesn't read the code can implement an end kernel or a wrapper kernel that will be transformed correctly.

tensorflow_probability/python/mcmc/transformed_kernel_test.py
```diff
@@ -308,6 +328,89 @@ def test_is_calibrated(self):
             FakeInnerKernel(lambda x: -x**2 / 2, False),
             tfb.Identity()).is_calibrated)
 
+  def test_transform_then_step(self):
+    new_kernel = transform_then_step(tfb.Identity())
```
axch, line 332: These two tests look redundant with the next two.  Delete them?  Or are they testing something nontrivial?

tensorflow_probability/python/mcmc/transformed_kernel_test.py
```diff
@@ -308,6 +328,89 @@ def test_is_calibrated(self):
...
+        bijector=tfb.Exp()
+    )
+    kernel_results = trans_kernel.inner_kernel.bootstrap_results(tf.zeros(2))
+    for _ in range(2):
+      _, kernel_results = trans_kernel.inner_kernel.one_step(tf.zeros(2),
+                                                             kernel_results)
+    adapted_step_size = self.evaluate(
+        kernel_results.inner_results.accepted_results.step_size)
+    # the step size for `FakeInnerKernel` is 10
+    self.assertFalse(adapted_step_size == 10)
```
axch, line 392: It would be better to check that it moved a significant amount, such as self.assertLess(adapted_step_size, 8).  Take more transitions if necessary.

Also, this test is randomized, so it should follow our test reproducibility conventions:
- Create an initial seed stream using `strm = test_util.test_seed_stream()`
- Pass explicit seeds to all random functions as `foo(..., seed=strm())`
- Check that you got it right by inserting a debug print somewhere and seeing that the adapted step size you get is the same across test runs
- Check that your test bound is reasonable by making sure the test passes consistently if the seed is varied.  You check this by running the test with `bazel test :foo --test_arg=--vary_seed --runs_per_test=10` (or 100, if the test is fast enough for that to be reasonable).

tensorflow_probability/python/mcmc/transformed_kernel_test.py
```diff
@@ -308,6 +328,89 @@ def test_is_calibrated(self):
...
+    pkr_one, pkr_two = self.evaluate([
+        double_step_kernel.bootstrap_results(2.),
+        double_step_kernel.bootstrap_results(9.),
+    ])
+    self.assertNear(np.log(2.),
+                    pkr_one.inner_results.inner_results.transformed_state,
+                    err=1e-6)
+    self.assertNear(np.log(9.),
+                    pkr_two.inner_results.inner_results.transformed_state,
+                    err=1e-6)
```
axch, line 414: Can you add another test to make sure `TransformedTransitionKernel` correctly nests with itself?  Use two different bijectors to make sure you are testing the order in which they are composed.

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -110,6 +106,33 @@ def transformed_log_prob_fn(*state_parts):
...
+    kernel = kernel.inner_kernel
+    kernel_stack.append(kernel)
+  return kernel.parameters['target_log_prob_fn'], kernel_stack
+
+
+def target_log_prob_setter(kernel_stack, new_target_log_prob):
+  # edit the appropriate kernel with the new log_prob_fn
+  last_kernel = kernel_stack.pop()
+  last_kernel_kwargs = last_kernel.parameters.copy()
+  last_kernel_kwargs.update(target_log_prob_fn=new_target_log_prob)
```
axch, line 124: Let's follow @csuter's offline suggestion and encapsulate this updating logic as a `copy(self, **override_parameters_kwargs)` method on the `TransitionKernel` base class, like in `Distribution` (don't worry about the `batch_slice` stuff in `Distribution`; that's not relevant here).

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -110,6 +106,33 @@ def transformed_log_prob_fn(*state_parts):
...
+  while 'target_log_prob_fn' not in kernel.parameters:
+    if 'inner_kernel' not in kernel.parameters:
+      raise ValueError('"None of the nested `inner_kernel`s contains a '
+                       '`target_log_prob_fn`."')
+    kernel = kernel.inner_kernel
+    kernel_stack.append(kernel)
+  return kernel.parameters['target_log_prob_fn'], kernel_stack
+
+
+def target_log_prob_setter(kernel_stack, new_target_log_prob):
```
axch, line 120: This is an example of API simplicity being in tension with performance (and to some extent with implementation simplicity).  This function will be much simpler to understand and use if it accepts the single top-level kernel and the new log_prob function, and computes the kernel stack on its own.  As it is, though, the user has to know where to get the kernel stack from in order to use this, and their code has to have an extra communication path (from the getter to the setter).  This API also hurts the use-case of replacing the target log_prob unconditionally, without reading the target that was there before (not that such a use-case is very common).

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -110,6 +106,33 @@ def transformed_log_prob_fn(*state_parts):
...
+
+  # propogate upwards
+  while kernel_stack:
+    curr_kernel = kernel_stack.pop()
+    with deprecation.silence():
+      curr_kernel._inner_kernel = type(last_kernel)(**last_kernel_kwargs)
+    last_kernel = curr_kernel
+    last_kernel_kwargs = last_kernel.parameters
+
+  return last_kernel, last_kernel_kwargs
```
axch, line 134: This function will also be much simpler to use and understand if it relies on returning the final reconstructed kernel rather than on mutating the `_inner_kernel` field of the top-level kernel, as it currently appears to do.  So, I suggest changing the API of this helper so the call site in `TransformedTransitionKernel` looks like
```
self._inner_kernel = update_target_log_prob(self._inner_kernel, new_target_log_prob)
```

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -110,6 +106,33 @@ def transformed_log_prob_fn(*state_parts):
   return transformed_log_prob_fn
 
 
+def target_log_prob_getter(kernel):
```
axch, line 109: Generally it's a good idea to name functions so that the call site reads as either a noun phrase describing the result computed on those arguments, or as a verb phrase describing the action performed on those arguments.  So, `get_target_log_prob(my_kernel)` parses much more cleanly than `target_log_prob_getter(my_kernel)`.  This is because the latter requires the reader to mentally form the image of a `target_log_prob_getter`, infer what that does, and infer what the result will be in the setting of `my_kernel`; whereas the former has one fewer mental step.

Given the fairly strong conventional meaning of `get`, `set`, and `update`, I think you should rename this to `get_target_log_prob`, and the next one to `update_target_log_prob`.

By the way, the conventional difference between `set` and `update` is that `set` implies mutating the object to change a value, whereas `update` implies constructing a new object that is identical except for the changed value.  I also suggest below that you change the setter to do the latter.

tensorflow_probability/python/mcmc/transformed_kernel.py
rupei: Done

rupei: Fixed.

tensorflow_probability/python/mcmc/transformed_kernel.py
rupei: Moved and added more relevant information.

tensorflow_probability/python/mcmc/transformed_kernel_test.py
rupei: Good point, I deleted them.

tensorflow_probability/python/mcmc/transformed_kernel_test.py
rupei: Done with similar flavor to the rest of the file (but I did test it out with multiple seeds using the test seed stream)

tensorflow_probability/python/mcmc/transformed_kernel_test.py
rupei: Added.

tensorflow_probability/python/mcmc/transformed_kernel.py
rupei: Done in similar style to `Distribution`

tensorflow_probability/python/mcmc/transformed_kernel.py
rupei: Good pointer; I hadn't even thought about it. Edited for API simplicity.

tensorflow_probability/python/mcmc/transformed_kernel.py
rupei: Fixed.

tensorflow_probability/python/mcmc/transformed_kernel.py
rupei: Done.

rupei:
b86ea65 simply undoes a rebase that made edits to the new Weibull Distribution (to make the CLA check happy). Has no impact on the actual contents of the 3 files I modified.

tensorflow_probability/python/mcmc/kernel.py
```diff
@@ -84,3 +84,25 @@ def bootstrap_results(self, init_state):  # pylint: disable=unused-argument
...
+        override_parameters_kwargs, with all shared keys being
+        overriden by the value in override_parameter_kwargs i.e.,
+        `dict(self.parameters, **override_parameters_kwargs)`.
+    """
+    parameters = dict(self.parameters, **override_parameter_kwargs)
+    new_kernel = type(self)(**parameters)
+    # pylint: disable=protected-access
+    new_kernel._parameters = parameters
+    # pylint: enable=protected-access
+    new_kernel = type(self)(**parameters)
```
axch, line 107: This line looks redundant?  Merge failure?

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -109,6 +105,35 @@ def transformed_log_prob_fn(*state_parts):
...
+    if 'inner_kernel' not in kernel.parameters:
+      raise ValueError('"None of the nested `inner_kernel`s contains a '
+                       '`target_log_prob_fn`."')
+    kernel = kernel.inner_kernel
+    kernel_stack.append(kernel)
+  return kernel_stack
+
+def get_target_log_prob(kernel):
+  kernel_stack = _make_kernel_stack(kernel)
+  last_kernel = kernel_stack[len(kernel_stack) - 1]
```
axch, line 120: This is written more concisely as kernel_stack[-1].

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -109,6 +105,35 @@ def transformed_log_prob_fn(*state_parts):
...
+  kernel_stack = [kernel]
+  while 'target_log_prob_fn' not in kernel.parameters:
+    if 'inner_kernel' not in kernel.parameters:
+      raise ValueError('"None of the nested `inner_kernel`s contains a '
+                       '`target_log_prob_fn`."')
+    kernel = kernel.inner_kernel
+    kernel_stack.append(kernel)
+  return kernel_stack
+
+def get_target_log_prob(kernel):
```
axch, line 118: This and `update` are (and, I think, should be) public functions.  Add docstrings explaining what they do, arguments, etc.

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -109,6 +105,35 @@ def transformed_log_prob_fn(*state_parts):
...
+  kernel_stack = _make_kernel_stack(kernel)
+  last_kernel = kernel_stack[len(kernel_stack) - 1]
+  return last_kernel.parameters['target_log_prob_fn']
+
+def update_target_log_prob(kernel, new_target_log_prob):
+  kernel_stack = _make_kernel_stack(kernel)
+  # get the last kernel and update to the new_target_log_prob
+  last_kernel = kernel_stack.pop().copy(target_log_prob_fn=new_target_log_prob)
+  last_kernel_kwargs = last_kernel.parameters
```
axch, line 127: Style: Do you need this here?  Can't you just access this field once in the place where you use `last_kernel_kwargs`, instead of maintaining another loop invariant?

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -221,6 +246,33 @@ def make_likelihood(true_variances):
       axis=0)
   ```
 
+  #### Detailed API Contract
+
+  `TransformedTransitionKernel` operates on a `target_log_prob_fn` field, which
+  is expected to be an attribute of some `inner_kernel` (which may or may not
+  be the most immediate). If the provided `inner_kernel` doesn't have a
+  `target_log_prob_fn`, `TransformedTransitionKernel` will iteratively look
+  through deeper layers of kernels to find one that does, stopping at the first
```
axch, line 255: When you say "deeper layers of kernels", be explicit that it searches in the `inner_kernel` field, if present, at each level.  Also find a way to make that sentence not awkward :)

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -221,6 +246,33 @@ def make_likelihood(true_variances):
...
+  one it finds (not the deepest). It will then apply the transformation to
+  that `target_log_prob_fn` and propogate those changes upwards, such that all
+  kernels that wrap around the updated one reflect that change. If it turns out
+  that no `inner_kernel` satisfies that condition, a `ValueError` will be
+  raised.
+
+  To illustrate, in the following example, since the `SimpleStepSizeAdaptation`
+  object has no `target_log_prob_fn`, that of the `HamiltonianMonteCarlo`
+  would be used.
```
axch, line 264: Would be updated.

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -254,18 +306,16 @@ def __init__(self, inner_kernel, bijector, name=None):
         inner_kernel=inner_kernel,
         bijector=bijector,
         name=name or 'transformed_kernel')
-    inner_kernel_kwargs = inner_kernel.parameters.copy()
-    target_log_prob_fn = inner_kernel_kwargs['target_log_prob_fn']
+    target_log_prob_fn = get_target_log_prob(self)
```
axch, line 309: Same effect but probably clearer style to `get_target_log_prob(inner_kernel)` here, rather than `self`.

tensorflow_probability/python/mcmc/transformed_kernel_test.py
```diff
@@ -308,6 +326,89 @@ def test_is_calibrated(self):
...
+    )
+    kernel_results = trans_kernel.inner_kernel.bootstrap_results(tf.zeros(2))
+    for _ in range(2):
+      _, kernel_results = trans_kernel.inner_kernel.one_step(tf.zeros(2),
+                                                             kernel_results)
+    adapted_step_size = self.evaluate(
+        kernel_results.inner_results.accepted_results.step_size)
+    self.assertLess(adapted_step_size, 7)
+
+  def test_deeply_nested(self):
```
axch, line 375: Um, shouldn't this test put the transformed transition kernel on the outside, to check that it penetrates multiple levels of inner kernels?

tensorflow_probability/python/mcmc/transformed_kernel_test.py
```diff
@@ -308,6 +326,89 @@ def test_is_calibrated(self):
...
+        inner_kernel=trans_kernel,
+        bijector=tfb.Identity())
+    step_kernel = tfp.mcmc.SimpleStepSizeAdaptation(
+        inner_kernel=double_trans_kernel,
+        num_adaptation_steps=9)
+    pkr_one, pkr_two = self.evaluate([
+        step_kernel.bootstrap_results(2.),
+        step_kernel.bootstrap_results(9.),
+    ])
+    self.assertNear(2., pkr_one.inner_results.transformed_state, err=1e-6)
```
axch, line 410: Did this actually work?  It seems like the transformed state should end being mapped through the composition of the transforming bijectors, rather than the last one?  Do we have a bug?

rupei: Just checked to make sure and it does indeed pass. Here's the issue:

Without this PR such a composition wouldn't be possible as `TransformedTransitionKernel` doesn't have a `target_log_prob_fn` itself (it would've never been possible to nest two calls to `TransformedTransitionKernel` without error). Hence, currently, the `bootstrap_results` method just applies the "inverse" function of the single outermost bijector. Presumably, `one_step` has similar behavior as the "forward" function is similarly defined.

My question is if we want to support this kind of compositionality with multiple `TransformedTransitionKernel`s in this PR. If yes, I believe the easiest implementation would be to update the bijector in a similar manner to the `target_log_prob_fn`. I wouldn't know how the `log_det_jacobian` method naturally updates though. The alternative is to preface somewhere that this is the intended behaviour, and we're not meant to stack `TransformedTransitionkernel`s like this (it will override previous transformations).

tensorflow_probability/python/mcmc/kernel.py
```diff
@@ -84,3 +84,22 @@ def bootstrap_results(self, init_state):  # pylint: disable=unused-argument
...
+      override_parameter_kwargs: Python String/value `dictionary` of
+      initialization arguments to override with new values.
+
+    Returns:
+      new_kernel: A `TransitionKernel` object of same type as `self`,
+      initialized with override_parameter_kwargs
+    """
+    parameters = dict(self.parameters, **override_parameter_kwargs)
+    new_kernel = type(self)(**parameters)
+    # pylint: disable=protected-access
```
csuter, line 101: you can put this pylint comment at the end of the line it applies to and avoid the need for the re-enable down below

csuter: (this may not apply anymore after you fix the merge issue alexey notes below, but it's good to know anyway)

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -33,15 +33,11 @@
...
-    mcmc_util.PrettyNamedTupleMixin,
-    collections.namedtuple('TransformedTransitionKernelResults',
-                           ['transformed_state',
-                            'inner_results',
-                            ])):
-  """Internal state and diagnostics for Transformed kernel."""
-  __slots__ = ()
+TransformedTransitionKernelResults = collections.namedtuple(
+    'TransformedTransitionKernelResults', [
+        'transformed_state',
```
csuter, line 38: looks like some recent changes to this file have been lost here

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -109,6 +105,35 @@ def transformed_log_prob_fn(*state_parts):
     return tlp + sum(ldj_fn(state_parts, event_ndims))
   return transformed_log_prob_fn
 
+def _make_kernel_stack(kernel):
+  kernel_stack = [kernel]
+  while 'target_log_prob_fn' not in kernel.parameters:
+    if 'inner_kernel' not in kernel.parameters:
+      raise ValueError('"None of the nested `inner_kernel`s contains a '
```
csuter, line 112: this error message does not seem right, based on the conditions being checked.

tensorflow_probability/python/mcmc/transformed_kernel.py
```diff
@@ -109,6 +105,35 @@ def transformed_log_prob_fn(*state_parts):
...
+  # get the last kernel and update to the new_target_log_prob
+  last_kernel = kernel_stack.pop().copy(target_log_prob_fn=new_target_log_prob)
+  last_kernel_kwargs = last_kernel.parameters
+
+  # propogate upwards
+  while kernel_stack:
+    curr_kernel = kernel_stack.pop()
+    with deprecation.silence():
+      curr_kernel._inner_kernel = type(last_kernel)(**last_kernel_kwargs)
+    last_kernel = curr_kernel
```
csuter, line 134: nit: call this prev_kernel to disambiguate "last" ~= "final"

tensorflow_probability/python/mcmc/transformed_kernel.py
csuter: why don't we copy this docstring over to a doc to collaboratively edit; i am not sure github's comment interface is going to be conducive to this task

tensorflow_probability/python/mcmc/transformed_kernel.py
csuter, line 118: i'd suggest calling this `find_nested_target_log_prob_recursively` which is more descriptive. i am not sure it should be part of the public API. is there a reason it needs to be? i'd prefer not to introduce new API tokens unless strictly necessary. The criterion for adding a new API item should be "does a typical user of this object need to use this functionality?". i don't think this is the case here (please let me know if I'm overlooking something).

tensorflow_probability/python/mcmc/transformed_kernel_test.py
```diff
@@ -41,12 +41,35 @@ def _maybe_seed(seed):
   return seed
 
 
+def transform_then_step(bijector):
```
csuter, line 44: let's name this `make_transofrm_then_adapt_kernel`, and the other one below, similarly

tensorflow_probability/python/mcmc/transformed_kernel_test.py
```diff
@@ -71,7 +96,7 @@ def setUp(self):
     super(TransformedTransitionKernelTest, self).setUp()
     self.dtype = np.float32
 
-  def test_support_works_correctly_with_HMC(self):
+  def test_support_works_correctly_with_hmc(self):
```
csuter, line 99: are these changes because pylint complained?

tensorflow_probability/python/mcmc/transformed_kernel.py
rupei: The idea is that if there are no more `inner_kernel`s we've hit the inner-most kernel and have already checked to see if it has a `target_log_prob_fn`. I did make the implicit assumption that all kernels will have an `inner_kernel` field; please correct me if I'm wrong

tensorflow_probability/python/mcmc/transformed_kernel.py
rupei: Good idea! I'll make a doc for this

tensorflow_probability/python/mcmc/transformed_kernel_test.py
rupei: yeah, all the similar caps lock changes are because of that

tensorflow_probability/python/mcmc/transformed_kernel.py
rupei: Kind of like with `SimpleStepSizeAdaptation` wouldn't it be useful for the user to be able to get the deepest `target_log_prob_fn` or edit it willingly without needing to manually go through the `inner_kernel`s?

tensorflow_probability/python/mcmc/kernel.py
rupei: Definitely good to know!

tensorflow_probability/python/mcmc/kernel.py
rupei: Oopsies on my part; fixed

tensorflow_probability/python/mcmc/transformed_kernel.py
rupei: Yup, fixed

tensorflow_probability/python/mcmc/transformed_kernel.py
rupei: Good point; fixed!

tensorflow_probability/python/mcmc/transformed_kernel_test.py
rupei: Yeah, good point.

tensorflow_probability/python/mcmc/transformed_kernel.py
rupei: I personally like the get/update pair in naming but completely open to change

rupei: Final thoughts: leave the get/update function private

tensorflow_probability/python/mcmc/transformed_kernel_test.py
rupei: final thoughts: keep lowercase

tensorflow_probability/python/mcmc/transformed_kernel_test.py
rupei: Final thoughts: intended behavior, write a test to make sure it is analogous to a bijector chain

tensorflow_probability/python/mcmc/kernel.py
```diff
@@ -84,3 +84,22 @@ def bootstrap_results(self, init_state):  # pylint: disable=unused-argument
...
+    Returns:
+      new_kernel: `TransitionKernel` object of same type as `self`,
+        initialized with the union of self.parameters and
+        override_parameter_kwargs, with any shared keys overridden by the
+        value of override_parameter_kwargs, i.e.,
+        `dict(self.parameters, **override_parameters_kwargs)`.
+    """
+    parameters = dict(self.parameters, **override_parameter_kwargs)
+    new_kernel = type(self)(**parameters)
+    new_kernel._parameters = parameters # pylint: disable=protected-access
```
csuter, line 104: is this needed? doesn't the __init__ call that creates new_kernel set them?

tensorflow_probability/python/mcmc/transformed_kernel.py
csuter: i made some (substantial) edits to the doc -- PTAL!

tensorflow_probability/python/mcmc/kernel.py
rupei: May be redundant now that I look at it; the copy method in `Distribution` has a similar pattern.

rupei:
One should probably also change the __init__ docstring for `TransformedTransitionKernel`. Currently, one of the args is:

inner_kernel: `TransitionKernel`-like object which has a
        `target_log_prob_fn` argument.

which is not necessarily true anymore.

PR #971, opened 2020-06-15
Add LambertW distributions to `__init__.py`

dirmeier:
Hello,

this PR adds the module names of the Lambert W families to `__init__.py`, since they where missing before.

Cheers,
Simon


dirmeier:
Not sure why this error is showing up as I only added the two distributions to init to make them available..

csuter:
Hi, thanks for the PR. These were left out of the init files because of some issues around shape support. The original contributor was about to leave Google and didn't want to lose their work, so we checked it in in an incomplete state. We'll need to address those shape issues before making these units public. I've just reconnected with the original contributor and we'll hopefully be digging back in shortly.

dirmeier:
Awesome! Let me know if I can help otherwise

axch:
In the meantime, I am closing this PR pending resolution of mentioned shape issues.

Issue #978, opened 2020-06-23
trace_fn default inconsistently used


dellaert:
It seems that the "new" default for the `trace_fn` argument of `tfp.mcmc.sample_chain` is inconsistently used in tests and documentation. I discovered this because the example code for RandomWalkMetropolis gave a warning when running it. The code reads:

```
  samples, _ = tfp.mcmc.sample_chain(
      num_results=num_results,
      current_state=init_state,
      kernel=tfp.mcmc.RandomWalkMetropolis(
          target_log_prob_fn=target.log_prob,
          seed=54),
      num_burnin_steps=200,
      num_steps_between_results=1,  # Thinning.
      parallel_iterations=1)
```

but should be

```
  samples = tfp.mcmc.sample_chain(
      num_results=num_results,
      current_state=init_state,
      kernel=tfp.mcmc.RandomWalkMetropolis(
          target_log_prob_fn=target.log_prob,
          seed=54),
      num_burnin_steps=200,
      num_steps_between_results=1,  # Thinning.
      trace_fn=None,
      parallel_iterations=1)
```

per the warning in sample_chain:

```
      warnings.warn("Tracing all kernel results by default is deprecated. Set "
                    "the `trace_fn` argument to None (the future default "
                    "value) or an explicit callback that traces the values "
                    "you are interested in.")
```

PR #979, opened 2020-06-23
Added `trace_fn=None` where intended

dellaert:
This should fix all instances in python code of inconsistent use of `trace_fn`, as outlined in issue #978. I did not touch notebooks. 

I also noticed the examples in soft_clip do not all work. I fixed one spelling mistake but don't have the expertise to track down the remaining issues.

Note I'm still reading code and have not locally built the project, so I'm relying on CI to verify everything compiles - I think it should. I'll happily close the PR if that is not acceptable, but it could be a while before I can devote the time to re-open.


dellaert:
@googlebot I signed it!

dellaert:
Hold off before merging as I figure out CLA - might have acted too soon :-/

dellaert:
OK, sorted that out. Please review at will ;-)

dellaert:
@jvdillon I tried reading more about tfp and even created a PR :-)

PR #981, opened 2020-06-24
Adding Streaming Variance and Covariance to tfp.experimental.stats

rupei:
Adding functionality in tfp.experimental.stats to support streaming variance and covariance. Variance supports scalar and vector cases and is computed using the pairwise version of Welford's Online Algorithm; covariance supports arbitrary rank samples and is computed with a pairwise update formula by Pebay (referenced in Docstring). Moreover, one can specify the event shape with the `event_ndims` parameter for running covariance, which won't calculate the full covariance over all dimensions, but only specified cross-products.


tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -469,6 +476,81 @@ def covariance(x,
 
     return cov
 
+CovarianceState = collections.namedtuple('CovarianceState',
+                                         'num_samples, mean, comoment, dtype')
+def init_running_covariance(shape, dtype=tf.float32):
+  """Initialize the Running Covariance State.
+
+  Args:
+    shape: Python `Tuple` representing the shape of incoming samples.
```
axch, line 485: I believe this code as written also accepts a fully-defined `tf.TensorShape`; check this and augment the docstring to mention that as an option.

Same for `init_running_variance`.

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -469,6 +476,81 @@ def covariance(x,
...
+  """
+  return CovarianceState(
+      num_samples=0,
+      mean=tf.zeros(shape),
+      comoment=tf.zeros(shape + shape),
+      dtype=dtype,
+    )
+
+def update_running_covariance(curr_state, new_sample):
+  """Update the streaming statistics of the `VarianceState` with a new sample
```
axch, line 500: You mean, `CovarianceState`?

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -469,6 +476,81 @@ def covariance(x,
...
+  using a Covariance Update Formula proposed by Philippe Pebay (2008).
+
+  Args:
+    curr_state: `CovarianceState` that depicts the current state of streaming
+      statistics
+    new_sample: Incoming sample with compatible shape and dtype as previously
```
axch, line 506: Slightly better wording here: "Incoming sample with shape and dtype compatible with those used to form the `CovarianceState` by `init_running_covariance`.

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -469,6 +476,81 @@ def covariance(x,
...
+  [1]: Philippe Pebay. Formulas for Robust, One-Pass Parallel Computation of
+  Covariances and Arbitrary-Order Statistical Moments. _Technical Report
+  SAND2008-6212_, 2008.
+  https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2008/086212.pdf
+  """
+  new_sample = tf.convert_to_tensor(new_sample, dtype=curr_state.dtype)
+  new_n = curr_state.num_samples + 1
+  delta_mean = new_sample - curr_state.mean
+  new_mean = curr_state.mean + delta_mean / new_n
+  all_pairwise_deltas = tf.tensordot(delta_mean, delta_mean, axes=0)
```
axch, line 522: Let's see whether we can get a batched version of this working.

The einsum equation for batch scalar covariance (= variance) by this algorithm is, I think,
tf.einsum("...i,...i->...i", delta_mean, delta_mean)
(I wonder whether tf.einsum("...,...->...", delta_mean, delta_mean) will also just do that.)

For batch vector covariance, I think we could use
tf.einsum("...i,...j->...ij", delta_mean, delta_mean)

For batch matrix covariance, I think we should be able to use
tf.einsum("...ij,...kl->...ijkl", delta_mean, delta_mean)

For batch over rank 3, we should be able to use
tf.einsum("...ijk,...lmn->...ijklmn", delta_mean, delta_mean)

etc.

I think everything else is transparently batch-friendly already.

WDYT?

rupei: Should definitely be possible; in fact, the streaming variance pairwise works now (it had a subtle bug). Working on this now in addition to quantiles.

tensorflow_probability/python/stats/sample_stats.py
rupei: Ah yes, thanks for the catch; the Docstrings were updated quite late at night and this slipped.

tensorflow_probability/python/stats/sample_stats.py
rupei: Yup! TensorShapes are good

brianwa84:
Can we group together the init, update, and finalize functions somehow? Even putting the symbols into submodules might suffice to let us logically group them? Or namedtuple or a proper object.

I'd suggest we expose the new symbols under experimental for now so we can toy with what feels natural. Impl could still live here unless imports prove a challenge.

`tfp.experimental.stats.running_covariance.{init,update,finalize}`?

tensorflow_probability/python/stats/sample_stats_test.py
```diff
@@ -603,6 +603,198 @@ def test_bad_axis_dynamic(self):
...
+      running_var = tfp.stats.update_running_variance(running_var, sample)
+    final_cov, final_var = self.evaluate([
+        tfp.stats.finalize_running_covariance(running_cov).covariance,
+        tfp.stats.finalize_running_variance(running_var).variance])
+    self.assertEqual(final_cov.shape, (5, 2))
+    self.assertAllClose(final_cov, final_var, rtol=1e-5)
+
+  def test_covariance_with_ndims(self):
+    rng = np.random.RandomState(seed=test_util.test_seed())
+    x = rng.rand(100, 3, 5, 2)
```
brianwa84, line 774: Could you add an example that accumulates a final covariance shaped 3,5,2,5,2? Ie a batch of 3x 5x2 covariances? Is that supported?

brianwa84:
Code looks awesome btw!

tensorflow_probability/python/stats/sample_stats_test.py
rupei: Yeah, good idea! It is supported so a test is probably warranted

tensorflow_probability/python/experimental/stats/BUILD
```diff
@@ -0,0 +1,62 @@
...
+    srcs = ["sample_stats_test.py"],
+    shard_count = 5,
+    deps = [
+        ":sample_stats",
+        # numpy dep,
+        # tensorflow dep,
+        "//tensorflow_probability",
+        "//tensorflow_probability/python/internal:test_util",
+    ],
+)
```
rupei, line 62: oops, will add a new line on the next commit

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -645,6 +807,104 @@ def variance(x, sample_axis=0, keepdims=False, name=None):
...
+      num_samples=tf.zeros((), dtype=dtype),
+      mean=tf.zeros(shape, dtype=dtype),
+      m2=tf.zeros(shape, dtype=dtype),
+      dtype=dtype,
+  )
+
+
+def update_running_variance(curr_state, new_sample, axis=None):
+  """Update the streaming statistics of the `VarianceState` with a new sample
+  using Welford's Online Algorithm.
```
ColCarroll, line 836: Can you cite the 1962 paper?

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -645,6 +807,104 @@ def variance(x, sample_axis=0, keepdims=False, name=None):
         x, y=None, sample_axis=sample_axis, event_axis=None, keepdims=keepdims)
 
 
+VarianceState = collections.namedtuple('VarianceState',
+                                       'num_samples, mean, m2, dtype')
```
ColCarroll, line 811: Is there a better name than `m2`? I have used "raw_variance_estimate"?

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -645,6 +807,104 @@ def variance(x, sample_axis=0, keepdims=False, name=None):
...
+        axis=axis)
+  else:
+    b_n = tf.ones((), dtype=curr_state.dtype)
+    b_mean = new_sample
+    b_m2 = tf.zeros(prefer_static.shape(curr_state.m2), dtype=curr_state.dtype)
+
+  total_n = curr_state.num_samples + b_n
+  delta = b_mean - curr_state.mean
+  new_mean = curr_state.mean + delta * (b_n / total_n)
+  new_m2 = (curr_state.m2 + b_m2
```
ColCarroll, line 872: You might actually note in the docstring that this works for batches or individual updates. I guess instead of the typical:

```python
old_delta = (x[t] - mean[t])
mean[t+1] = mean[t] + old_delta / t
new_delta = (x[t] - mean[t + 1])
raw_var[t + 1] = raw_var[t] + old_delta * new_delta
```

you are using

```python
delta = (x[t] - mean[t])
mean[t+1] = mean[t] + delta / t
raw_var[t + 1] = raw_var[t] + delta * delta * t / t + 1
```
and `b_m2` is a correction for batches?

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -645,6 +807,104 @@ def variance(x, sample_axis=0, keepdims=False, name=None):
...
+  return VarianceState(
+      num_samples=tf.zeros((), dtype=dtype),
+      mean=tf.zeros(shape, dtype=dtype),
+      m2=tf.zeros(shape, dtype=dtype),
+      dtype=dtype,
+  )
+
+
+def update_running_variance(curr_state, new_sample, axis=None):
+  """Update the streaming statistics of the `VarianceState` with a new sample
```
ColCarroll, line 835: Keeping wording consistent:
```suggestion
  """Update the running statistics of the `VarianceState` with a new sample
```

tensorflow_probability/python/experimental/stats/sample_stats_test.py
```diff
@@ -0,0 +1,154 @@
...
+import tensorflow.compat.v2 as tf
+import tensorflow_probability as tfp
+from tensorflow_probability.python.internal import test_util
+
+@test_util.test_all_tf_execution_regimes
+class RunningStatsTest(test_util.TestCase):
+  """Example use of class-structured RunningCovariance. Individual methods
+  tested more rigorously in tfp.stats"""
+
+  def test_running_covariance(self):
```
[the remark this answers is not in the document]
rupei: Much better than writing two tests just with a change of parameters; thanks!

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -470,6 +477,161 @@ def covariance(x,
...
+  Covariances and Arbitrary-Order Statistical Moments. _Technical Report
+  SAND2008-6212_, 2008.
+  https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2008/086212.pdf
+  """
+  if tf.is_tensor(new_sample):
+    new_sample = tf.cast(new_sample, dtype=curr_state.dtype)
+  else:
+    new_sample = tf.convert_to_tensor(new_sample, dtype=curr_state.dtype)
+
+  def _generate_pairwise_deltas(sample):
```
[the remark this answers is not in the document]
rupei: Yeah that's exactly what's going on. You're right though, it is a convoluted function that should be documented

tensorflow_probability/python/stats/sample_stats.py
rupei: Yup, pretty much. Though instead of calculating the raw variance, I just calculate the sum of squared distances from the mean (i.e. no  t / t + 1 term which is handled in finalize); hence, m2 is meant to represent that

tensorflow_probability/python/stats/sample_stats.py
rupei: Perhaps `sum_squared_residuals` given that `m2` isn't exactly an estimate of the variance yet? I'll refactor to that for now but open to naming suggestions (:

tensorflow_probability/python/stats/sample_stats.py
rupei: Done

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -0,0 +1,49 @@
...
+  def update(self, new_sample, axis=None):
+    self.covariance_state = update_running_covariance(
+        self.covariance_state, new_sample, axis)
+
+  def finalize(self, ddof=0):
+    return finalize_running_covariance(self.covariance_state, ddof)
+
+  @property
+  def num_samples(self):
+    return self.num_samples
```
brianwa84, line 45: surely this doesn't work?

also: this object will not be able to pass through the arguments/return values of a tf.while_loop body. I'd suggest instead just grouping the functions as a namedtuple a la:

stats/__init__.py
```
streaming_covariance = collections.namedtuple('streaming_stat', 'init,update,finalize')(
  init=init_running_covariance,
  update=update_running_covariance,
  finalize=finalize_running_covariance)
```

Then usage would be something like
```
streaming_stat = tfp.experimental.stats.streaming_covariance
_, s = tf.while_loop(
    lambda i, _: i < 10,
    lambda i, s: i+1, streaming_stat.update(s, some_val, [axis here?])
    # or make `tfp.experimental.stats.streaming_covariance(axis)` a function that returns a namedtuple with update closing over axis?
    (0, streaming_stat.init(..))
streaming_stat.finalize(s)  # again, should ddof go here or be captured by a closure?

Comment (brianwa84) [reply tensorflow_probability/python/experimental/stats/sample_stats.py:45]: (by the first line, i mean surely `self.num_samples` doesn't work; the rest of the comment is higher-level; sorry if confusing)

Comment (rupei) [reply tensorflow_probability/python/experimental/stats/sample_stats.py:45]: haha it definitely doesn't, was meant to be `self.covariance_state.num_samples`. And ohhh yeah the namedtuple approach makes sense, I'll switch it over. Thanks for the feedback Brian!

File: tensorflow_probability/python/stats/__init__.py [hunk 14]
```diff
@@ -40,7 +40,13 @@
 from tensorflow_probability.python.stats.sample_stats import cholesky_covariance
 from tensorflow_probability.python.stats.sample_stats import correlation
 from tensorflow_probability.python.stats.sample_stats import covariance
+from tensorflow_probability.python.stats.sample_stats import finalize_running_covariance
```
axch, line 43: Let's also move these functions into the experimental directory, and either make them private or inline them into methods.  One goal of making a class or tuple to hold the functions was to reduce the number of symbols in the stats namespace.

rupei: Sounds good! So to be clear, everything should live in experimental?

tensorflow_probability/python/stats/sample_stats.py
rupei: Sounds good!

brianwa84:
I think I might like @axch's idea of making the update and finalize methods of the state, though I'd like to hear others' opinions.
Also would like to understand if there is a common `update` interface between covariance, variance, mean, rhat, etc.

To do so, you may need to use a `@staticmethod` or `@classmethod` for init, since namedtuple provides `__init__` e.g.:

```
class RunningCovariance(collections.namedtuple('StreamingCovarianceState', [...])):
  __slots__ = ()
  @classmethod
  def init(cls, ...):
    return cls(*init_running_covariance(..))
  def update(self, ...):
    ...
  def finalize(self, ...):
    ...
```

+1 to adding a couple of while_loop based tests.

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -470,6 +477,186 @@ def covariance(x,
...
+  Returns:
+    state: `CovarianceState` with updated calculations
+
+  References:
+  [1]: Philippe Pebay. Formulas for Robust, One-Pass Parallel Computation of
+  Covariances and Arbitrary-Order Statistical Moments. _Technical Report
+  SAND2008-6212_, 2008.
+  https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2008/086212.pdf
+  """
+  if tf.is_tensor(new_sample):
```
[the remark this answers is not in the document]
rupei: Seems like it does; nice catch!

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -470,6 +477,186 @@ def covariance(x,
...
+      b_comoment += _generate_pairwise_deltas(
+          tf.gather(b_delta_mean, i, axis=axis))
+      return (i + 1, b_comoment)
+    # if a dynamic shape is given, we have a special case of `shape_invariant`
+    if new_sample.shape == tf.TensorShape(None):
+      comoment_shape_invariant = new_sample.shape
+    else:
+      comoment_shape_invariant = tf.TensorShape(
+          [None for _ in range(prefer_static.rank(curr_state.comoment))])
+    _, b_comoment = tf.while_loop(
```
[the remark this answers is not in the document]
rupei: 100% the better way to do it

tensorflow_probability/python/stats/__init__.py
axch: Yes, I think that's the right place to start with this one.

rupei:
Some comments in this new commit regarding the design decision of number of end-points (separate variance and covariance) or even the use of a separate mechanism for variance when it's handled by running covariance with `event_ndims=0`. The comments depict the similarities between the two algorithms, and I think (unless there's a performance trade-off in the implementation I overlooked) that having covariance is sufficient. Thoughts?

tensorflow_probability/python/experimental/stats/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+
+from tensorflow_probability.python.experimental.stats.sample_stats import RunningCovariance
+from tensorflow_probability.python.experimental.stats.sample_stats import RunningCovarianceState
+
+# pylint: enable=unused-import,wildcard-import,line-too-long,g-importing-member
+
+from tensorflow.python.util.all_util import remove_undocumented  # pylint: disable=g-direct-tensorflow-import
+
+_allowed_symbols = [
+    'RunningCovariance',
```
axch, line 31: Let's also add a `RunningVariance` object that's an alias for `RunningCovariance` but defaults the event_ndims to 0 instead of `full`.  You can probably use inheritance to make this have very little boilerplate code.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -0,0 +1,355 @@
...
+        For example, use `ddof=0` for population covariance and `ddof=1` for
+        sample covariance. Defaults to the population covariance.
+
+    Returns:
+      covariance: an estimate of the covariance
+    """
+    return _finalize_running_covariance(state, ddof=ddof)
+
+def _init_running_covariance(shape, event_ndims='full', dtype=tf.float32):
+  if event_ndims == 'full':
```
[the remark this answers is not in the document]
rupei: Noted, will go with relying exclusively on `None`

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -0,0 +1,355 @@
...
+      mean=new_mean,
+      comoment=new_comoment,
+  )
+
+
+def _finalize_running_covariance(curr_state, ddof=0):
+  return curr_state.comoment / (curr_state.num_samples - ddof)
+
+
+VarianceState = collections.namedtuple(
```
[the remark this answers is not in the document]
rupei: Yup, that's the plan

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -0,0 +1,355 @@
...
+def _init_running_covariance(shape, event_ndims='full', dtype=tf.float32):
+  if event_ndims == 'full':
+    event_ndims = len(shape)
+  if event_ndims > len(shape):
+    raise ValueError('Cannot calculate cross-products in {} dimensions for '
+                     'samples of rank {}'.format(event_ndims, len(shape)))
+  if event_ndims > 13:
+    raise ValueError('`event_ndims` over 13 not supported')
+  extra_ndims_shape = () if event_ndims == 0 else shape[-event_ndims:]
+  return RunningCovarianceState(
```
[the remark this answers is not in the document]
rupei: Yup, makes sense!

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -0,0 +1,355 @@
...
+      num_samples=tf.zeros((), dtype=dtype),
+      mean=tf.zeros(shape, dtype=dtype),
+      comoment=tf.zeros(shape + extra_ndims_shape, dtype=dtype),
+    ), event_ndims
+
+
+def _generate_pairwise_deltas(delta_mean, event_ndims):
```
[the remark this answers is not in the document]
rupei: A consideration I didn't make, thanks for letting me know! I'm assuming it would still be useful to lay out the current use-case in the Docstring, but please correct me if otherwise.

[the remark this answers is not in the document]
rupei: I've convinced myself that describing a specific use-case may be more confusing that just leaving it general; hence, I now lean towards opting for the latter.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -28,64 +28,147 @@
...
+  new_mean = state.mean + chunk_n * delta_mean / new_n
+  all_pairwise_deltas = _generate_batch_outer_product(
+      delta_mean, event_ndims)
+  adj_factor = state.num_samples * chunk_n / (state.num_samples + chunk_n)
+  new_comoment = (state.comoment + chunk_comoment
+                  + adj_factor * all_pairwise_deltas)
+
+  return type(state)(new_n, new_mean, new_comoment)
+
+def _generate_batch_outer_product(target, event_ndims):
```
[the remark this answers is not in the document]
rupei: Sounds good! I'll try to actively pay more attention to the noun vs verb naming rationale from now on

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -28,64 +28,147 @@
...
+        comoment=tf.zeros(self.shape + extra_ndims_shape, dtype=self.dtype),
+    )
     return RunningCovarianceState(**init_state._asdict())
 
   def update(self, state, new_sample, axis=None):
-    """Update the running statistics of the `RunningCovarianceState` with a
-    new sample using the covariance update formula proposed by Philippe
-    Pebay (2008).
+    """Update the `RunningCovarianceState` with a new sample using the update
+    formula proposed by Philippe Pebay (2008).
```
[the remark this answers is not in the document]
rupei: Clever

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -28,64 +28,147 @@
...
       dtype: Dtype of incoming samples and the resulting statistics.
-        By default, the dtype is `tf.float32`
+        By default, the dtype is `tf.float32`.
+
+    Raises:
+      ValueError: if `event_ndims` is greater than the rank of the intended
+        incoming samples, or if it's greater than 13. In the former case,
+        the operation is extraneous. In the latter, `tf.einsum` indexing
+        will fail in subsequent updates since `event_ndims * 2` letters are
+        required for indexing.
```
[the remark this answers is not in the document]
rupei: Ah, this is a limitation of this implementation.

Regarding (i): technically, tf.einsum can work on 13+ dimensions if making the Tensor/the computation doesn't break Tensorflow (i.e. with `tf.ones((1, 1..., 1))` in 14 dimensions). So you're right that we definitely shouldn't be implying this.
(ii): Another good point, I'm actually not sure if this breaks something in Tensorflow (I've noticed there's a lot of `Resource Exhausted` errors when dealing with larger rank Tensors).

I'm with you for the idea of just leaving the 13 out of the docstring, but keeping the error message the same.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -167,15 +166,31 @@ def initialize(self):
...
+    computation at once, which is more efficient than one by one.
+
+    To further illustrate the difference between batching and chunking, consider
+    the following example:
+
+    ```python
+    import tensorflow as tf
+    import tensorflow_probability as tfp
+    # treat as 3 independent random variables of shape (5, 2), where each of
+    # those random variables represent 5 shape (2,) samples from the same chain
```
axch, line 185: Wait, wouldn't this be 3 samples from each of 5 independent vector random variables of shape (2,)?

tensorflow_probability/python/experimental/stats/BUILD
```diff
@@ -0,0 +1,62 @@
...
+    srcs = ["sample_stats.py"],
+    srcs_version = "PY2AND3",
+    deps = [
+        # numpy dep,
+        # tensorflow dep,
+        "//tensorflow_probability/python/stats"
+    ],
+)
+
+py_test(
```
axch, line 50:
```suggestion
# py2and3
py_test(
```

A little bit of bureaucracy to do with our internal Github integration.  End result should look like, for example, https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/distributions/BUILD#L3235-L3236

rupei:
Squashed the ~22 commits in 22cf1e3

axch:
For future reference, it's actually more helpful not to squash the full history, because then when iterating on code review, the reviewer can review just the commits added since their last review, instead of having to read the whole PR again.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -0,0 +1,246 @@
...
+    chunk_mean = new_sample
+    chunk_comoment = tf.zeros(ps.shape(state.comoment), dtype=dtype)
+
+  new_n = state.num_samples + chunk_n
+  delta_mean = chunk_mean - state.mean
+  new_mean = state.mean + chunk_n * delta_mean / new_n
+  all_pairwise_deltas = _batch_outer_product(
+      delta_mean, event_ndims)
+  adj_factor = state.num_samples * chunk_n / (state.num_samples + chunk_n)
+  new_comoment = (state.comoment + chunk_comoment
```
SiegeLordEx, line 63: The original paper divides the comoment by n each time, meaning that it doesn't require the 'finalize' method (at least, modulo the ddof arg), and let's you easily do smoothing and moving averages by manipulating the `num_samples` state independent of the updates. Was there a reason why the `num_samples` was multiplied into the formula here?

Either way, as is `comoment` is misnamed, since it's `comoment_sum` or something like that.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -0,0 +1,246 @@
...
+    event_ndims: Both the number of dimensions that specify the event
+      shape and the desired number of dimensions for cross product
+      terms.
+
+  Returns:
+    outer_product: A `Tensor` with shape B + E + E for all pairwise
+      products of `target` in the event dimensions.
+  """
+  assign_indices = ''.join(list(
+      map(chr, range(ord('a'), ord('a')+event_ndims*2))))
```
SiegeLordEx, line 93: There should be spaces around binary operators (+ and * here)

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -0,0 +1,246 @@
...
+  first_indices = assign_indices[:event_ndims]
+  second_indices = assign_indices[event_ndims:]
+  einsum_formula = "...{},...{}->...{}".format(
+      first_indices, second_indices, assign_indices)
+  return tf.einsum(einsum_formula, target, target)
+
+
+class RunningCovariance(object):
```
SiegeLordEx, line 101: It's weird to me that we ended up in a world where we have this class, and then a Reducer class which will presumably be just a thin wrapper around it. But, I guess that's why this is in experimental first.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -0,0 +1,246 @@
...
+    if self.event_ndims == 0:
+      extra_ndims_shape = ()
+    else:
+      extra_ndims_shape = self.shape[-self.event_ndims:]
+    init_state = RunningCovarianceState(
+        num_samples=tf.zeros((), dtype=self.dtype),
+        mean=tf.zeros(self.shape, dtype=self.dtype),
+        comoment=tf.zeros(self.shape + extra_ndims_shape, dtype=self.dtype),
+    )
+    return RunningCovarianceState(**init_state._asdict())
```
SiegeLordEx, line 166: Why the weird copy here?

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -0,0 +1,246 @@
...
+      new_sample: Incoming sample with shape and dtype compatible with those
+        used to form the `RunningCovarianceState`.
+      axis: If chunking is desired, this is an integer that specifies the axis
+        with chunked samples. For individual samples, set this to `None`. By
+        default, samples are not chunked (`axis` is None).
+
+    Returns:
+      state: `RunningCovarianceState` with updated calculations.
+
+    References:
```
SiegeLordEx, line 207: The format here is wrong (see the STYLE_GUIDE.md), it should be:

```
#### References

[1]: Tony Finch. Incremental calculation of weighted mean and variance.
     _Technical Report_, 2009. Blah blah

File: tensorflow_probability/python/experimental/stats/sample_stats_test.py [hunk 32]
```diff
@@ -0,0 +1,437 @@
...
+      state = running_var.update(state, sample)
+    final_mean, final_var = self.evaluate([
+        state.mean,
+        running_var.finalize(state)])
+    self.assertEqual(final_mean, 0.)
+    self.assertEqual(final_var, 0.)
+
+  @parameterized.parameters(0, 1)
+  def test_running_variance(self, ddof):
+    rng = np.random.RandomState(seed=test_util.test_seed())
```
SiegeLordEx, line 49: This won't work in general. RandomState requires 32 bit integers as seeds, while test_seed returns something much larger. If you insist upon using test_seed like this, you should use test_util.test_np_rng() instead.

tensorflow_probability/python/experimental/stats/sample_stats_test.py
```diff
@@ -0,0 +1,437 @@
...
+    self.assertEqual(running_cov.dtype, tf.float32)
+
+  def test_tf_while(self):
+    rng = np.random.RandomState(seed=test_util.test_seed())
+    x = rng.rand(100, 10)
+    tensor_x = tf.convert_to_tensor(x, dtype=tf.float32)
+    running_cov = tfp.experimental.stats.RunningCovariance((10,))
+    running_var = tfp.experimental.stats.RunningVariance((10,))
+    _, cov_state = tf.while_loop(
+        lambda i, _: tf.less(i, 100),
```
SiegeLordEx, line 367: i < 100 will work

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: The point about simplifying without a `finalize` sounds important. Is the "original paper" the technical report I cited (https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2008/086212.pdf) or something else? I'm not entirely clear about the ease of smoothing/moving averages. `num_samples` is in the formula because I believe that was the final proposal from the technical paper. Do you think it would be better to change the API to omit `finalize`? I wanted to support multiple choices of degrees of freedom without redoing computation (i.e. streaming computation on a bunch of samples, then ask for the sample and population covariance).

Also, I thought comoment was defined as that sum (i.e. \sum_{i=1}^{n} (x_i - \bar{x})(y_i - \bar{y}). Please let me know if I'm mistaken haha then the naming should definitely change.

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: Good catch!

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: Has also felt somewhat weird to me. Maybe in future iterations we make it solely as a reducer?

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: Oops, this is unnecessary

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: Not exactly sure which part is formatted wrong (could you point me towards it)? Is it the fact that leaving _Technical Report_ is sufficient?

SiegeLordEx: References should be a markdown heading, and it should use hanging alignment.

```
#### References

[1]: Philippe Pebay. Formulas for Robust, One-Pass Parallel Computation of
     Covariances and Arbitrary-Order Statistical Moments. _Technical Report
     SAND2008-6212_, 2008.
     https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2008/086212.pdf
```

rupei: Oops, I was looking at all the wrong places haha. Will fix

tensorflow_probability/python/experimental/stats/sample_stats.py
SiegeLordEx: Looks like I was confused by the notation in that paper, there they always operate on sums rather than averages. Moments, however, are averages (or expectations) by (common?) definition, so the 'M' (and 'C') in that paper are not (co)moments. I don't know if there's a nice name for that, it's commonly described as `sum of squared deviations` in case of variance.

I still would prefer to have this implemented by maintaining a running covariance, rather (num_samples * covariance), but I don't have a reference for this being well behaved, so aside from the name of `comoment`, I retract my objections with the code.

rupei: Sounds good!

tensorflow_probability/python/experimental/stats/__init__.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2018 The TensorFlow Probability Authors.
```
SiegeLordEx, line 1: 2020

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -0,0 +1,265 @@
+# Copyright 2018 The TensorFlow Probability Authors.
```
SiegeLordEx, line 1: 2020

tensorflow_probability/python/experimental/stats/sample_stats_test.py
```diff
@@ -0,0 +1,438 @@
+# Copyright 2018 The TensorFlow Probability Authors.
```
SiegeLordEx, line 1: 2020

tensorflow_probability/python/experimental/stats/sample_stats_test.py
```diff
@@ -0,0 +1,438 @@
...
+    tensor_x = tf.convert_to_tensor(x, dtype=tf.float32)
+    running_cov = tfp.experimental.stats.RunningCovariance((10,))
+    def _loop_body(i, state):
+      if not tf.executing_eagerly():
+        sample = tf1.placeholder_with_default(tensor_x[i], shape=None)
+      else:
+        sample = tensor_x[i]
+      return (i+1, running_cov.update(state, sample))
+    _, state = tf.while_loop(
+        lambda i, _: tf.less(i, 100),
```
SiegeLordEx, line 392: This one too.

tensorflow_probability/python/experimental/stats/sample_stats_test.py
```diff
@@ -0,0 +1,438 @@
...
+    tensor_x = tf.convert_to_tensor(x, dtype=tf.float32)
+    running_var = tfp.experimental.stats.RunningVariance((10,))
+    def _loop_body(i, state):
+      if not tf.executing_eagerly():
+        sample = tf1.placeholder_with_default(tensor_x[i], shape=None)
+      else:
+        sample = tensor_x[i]
+      return (i+1, running_var.update(state, sample))
+    _, state = tf.while_loop(
+        lambda i, _: tf.less(i, 100),
```
SiegeLordEx, line 420: And this one.

tensorflow_probability/python/experimental/stats/sample_stats_test.py
```diff
@@ -0,0 +1,438 @@
...
+    rng = test_util.test_np_rng()
+    x = rng.rand(100, 10)
+    tensor_x = tf.convert_to_tensor(x, dtype=tf.float32)
+    running_var = tfp.experimental.stats.RunningVariance((10,))
+    def _loop_body(i, state):
+      if not tf.executing_eagerly():
+        sample = tf1.placeholder_with_default(tensor_x[i], shape=None)
+      else:
+        sample = tensor_x[i]
+      return (i+1, running_var.update(state, sample))
```
SiegeLordEx, line 418: Binary ops should have spaces around them.

PR #1001, opened 2020-07-06
Add `step_kernel` driver to tfp.experimental.mcmc

rupei:
This new driver behaves somewhat similarly to `sample_chain`, but is much more minimal to support the incoming `Reducer` class. Notably, its sole job is to run a `TransitionKernel`'s one step method a specified number of times. Technically, users can now write their own streaming MCMC functionality, but this will be made much easier in the near future.


tensorflow_probability/python/experimental/mcmc/BUILD
```diff
@@ -325,3 +326,38 @@ py_test(
...
+    srcs_version = "PY2AND3",
+    deps = [
+        # numpy dep,
+        # tensorflow dep,
+        "//tensorflow_probability/python/mcmc/internal:util",
+    ],
+)
+
+py_library(
+    name = "reducer",
```
rupei, line 342: For the `Reducer` base class class (immediate next step)

tensorflow_probability/python/experimental/mcmc/streaming_reductions.py
```diff
@@ -0,0 +1,104 @@
...
+  `current_state`.
+
+  Args:
+    num_steps: Integer number of Markov chain steps.
+    current_state: `Tensor` or Python `list` of `Tensor`s representing the
+      current state(s) of the Markov chain(s).
+    previous_kernel_results: A `Tensor` or a nested collection of `Tensor`s
+      representing internal calculations made within the previous call to this
+      function (or as returned by the kernel's `bootstrap_results` if None).
+      Default value: `None` (i.e., return of kernel.bootstrap_results)
```
axch, line 54:
```suggestion
    previous_kernel_results: A `Tensor` or a nested collection of `Tensor`s.
      Warm-start for the auxiliary state needed by the given `kernel`.
      If not supplied, `step_kernel` will cold-start with `kernel.bootstrap_results`.
```

tensorflow_probability/python/experimental/mcmc/streaming_reductions.py
```diff
@@ -0,0 +1,104 @@
...
+    kernel: An instance of `tfp.mcmc.TransitionKernel` which implements one step
+      of the Markov chain.
+    return_final_kernel_results: If `True`, then the final kernel results are
+      returned alongside the chain state after `num_steps` steps are taken.
```
axch, line 58:
```suggestion
    return_final_kernel_results: If `True`, then the final kernel results are
      returned alongside the chain state after `num_steps` steps are taken.
      This can be useful to inspect the final auxiliary state, or for a later
      warm restart.
```

tensorflow_probability/python/experimental/mcmc/streaming_reductions.py
```diff
@@ -0,0 +1,104 @@
...
+    seed: Optional, a seed for reproducible sampling.
+    name: Python `str` name prefixed to Ops created by this function.
+      Default value: `None` (i.e., 'mcmc_sample_chain').
+
+  Returns:
+    next_state, final_kernel_results: If `return_final_kernel_results` is
+      `True`, the kernel results after `num_steps` steps will be returned
+      alongside the resultant Markov chain state.
+    next_state: Conversely, if `return_final_kernel_results` is `False`,
+      only the resultant state will be returned.
```
axch, line 70: The convention for these is to list each member of the tuple on its own line, describing just those.  In this case, that would be

```
Returns:
  final_state: ...
  final_kernel_results: ...  Only returned if `return_final_kernel_results` was `True`.
```

tensorflow_probability/python/experimental/mcmc/streaming_reductions.py
```diff
@@ -0,0 +1,104 @@
...
+      return [passalong_seed] + list(
+          kernel.one_step(*state_and_results, **one_step_kwargs))
+
+    _, next_state, final_kernel_results = mcmc_util.smart_for_loop(
+        loop_num_iter=num_steps,
+        body_fn=_seeded_one_step,
+        initial_loop_vars=list((seed, current_state, previous_kernel_results)),
+        parallel_iterations=parallel_iterations)
+
+    # return semantics are simple enough to not warrant the need of named tuples
```
axch, line 99:
```suggestion
    # return semantics are simple enough to not warrant the use of named tuples
```

tensorflow_probability/python/experimental/mcmc/streaming_reductions_test.py
```diff
@@ -0,0 +1,127 @@
...
+      tfp.mcmc.sample_chain(
+          num_results=2,
+          current_state=0,
+          kernel=kernel,
+          trace_fn=lambda current_state, kernel_results: kernel_results,
+          seed=test_util.test_seed())
+    self.assertTrue(
+        any('supplied `TransitionKernel` is not calibrated.' in str(
+            warning.message) for warning in triggered))
+
```
axch, line 124: Add a smoke test for reproducibility, to check that seeds are handled correctly.  For this test,

- Make a new fake kernel that ignores the current state and moves to some random (but seeded) state.
- Run at least two transitions of this kernel with `step_kernel`, at least twice with the same initial seed.
- Check that both result sequences are the same, but within one sequence the samples are very different.

Report back if anything goes mysteriously wrong -- there are subtleties of platform + seed handling that we may need to discuss offline.

rupei: A couple of interesting facts I found out for myself (i.e. tfp seeding after `sampler.sanitize_seed` doesn't work with numpy as the seeds can become negative). Hence, I simply used some arbitrary `tfp.random` mechanism.

More importantly, an elegant way of ensuring that "within one sequence the samples are very different" didn't roll off of the top of my head, probably because `step_kernel` is meant to only return the last result by design (if you traced everything this would go against its spirit and it doesn't allow that implicitly by not giving any freedom in the loop variable shape), and with a truly random state it doesn't really support warm restart. Hence, I had to do a weird for loop implementation that does redo work.

tensorflow_probability/python/experimental/mcmc/streaming_reductions_test.py
```diff
@@ -122,6 +125,32 @@ def test_calibration_warning(self):
         any('supplied `TransitionKernel` is not calibrated.' in str(
             warning.message) for warning in triggered))
 
+  def test_seed_reproducibility(self):
+    first_fake_kernel = RandomTransitionKernel()
+    last_state = self.evaluate(step_kernel(
+        num_steps=1,
+        current_state=0,
+        kernel=first_fake_kernel,
+        seed=test_util.test_seed(),
```
axch, line 134: There should be a single `seed=test_util.test_seed()` at the top, and the test should reuse that seed in each of these three places.  The reason is that when testing robustness to seed changes with `--test_arg=--vary_seed`, each call to `test_util.test_seed()` will actually return a different seed.

rupei: When fixing the `seed` structure, I actually noticed that running that procedure in eager mode doesn't lead to deterministic results. Even when seeded, `first_fake_kernel` and `second_fake_kernel` give different results after `num_steps` steps. After some digging, I discovered the cause was because the seed was changing after calls to `samplers.sanitize_seed`. After subsequently looking through the tests for `samplers`, I also realized this is expected behaviour for eager mode.

Though the fix is rather simple, those realizations led me to also add a seed reproducibility test to `sample_chain` (since none of a similar flavor have been written yet), which also failed at first. I added all the changes I made to this PR (for reference), though the `sample_chain` edits should maybe be moved to another PR if it's actually a bug.

I'm wondering if what I found is actually expected behaviour (i.e. the user has to somehow find a way to seed if they're running eager mode, and it's not the responsibility of the driver), or if it's actually an oversight.

tensorflow_probability/python/experimental/mcmc/streaming_reductions.py
```diff
@@ -72,7 +72,10 @@ def step_kernel(
       `return_final_kernel_results` is `True`.
   """
   is_seeded = seed is not None
-  seed = samplers.sanitize_seed(seed, salt='mcmc.sample_chain')
+  if tf.executing_eagerly():
+    tf.random.set_seed(seed)
```
axch, line 76: Definitely don't do this.

I'm afraid we didn't actually explain what our prng seeding semantics actually are, which is why you got confused.  There are actually two seed types:
- A Python integer, which triggers "Classic TF" seed semantics, and
- A Tensor of 2 int32, which triggers "stateless" seed semantics.

The stateless ones are simpler: Any random function from TFP, and from tf.random.stateless_*, when given a seed of that type, is just a deterministic function of the seed.

The Classic TF ones are complicated, and use mysteriously changing hidden state to try to approximate the "mutating PNG" style that one gets from, say, numpy.  Unfortunately, the resulting semantics are very complicated in graph mode, don't correctly support parallel iterations in while loops, are different in eager mode, and (IIRC) don't work with XLA at all (because XLA doesn't support the required hidden changing state).

TFP supports the Classic TF semantics for backward compatibility, but all of our code (at least that pertains to MCMC) is internally coded in stateless style, so we convert incoming seeds to the type that triggers stateless behavior on input to every public function.

That's what `sanitize_seed` does: if the incoming seed is of stateless type (i.e., a Tensor of two int32), it just passes it along (possibly with salting, though that's irrelevant).  But if the incoming seed is of Classic TF type, i.e. the user asked for Classic TF sampling semantics, then `sanitize_seed` invokes `tf.random.uniform` (which has Classic TF seeding semantics) to generate a stateless seed.  Thus, `sanitize_seed(7)` is not a deterministic function of 7, per the complicated Classic TF rules; but if you only call `sanitize_seed(7)` once in a given test, the returned object is a seed that will trigger the simple, deterministic prng semantics we all know and love.

The last piece of this is that `test_util.test_seed()` returns a Classic TF seed (i.e., a Python integer), because it's older than the migration to stateless seeds.  Maybe it's getting to be time to fix that (or at least add another test utility).

Bottom line: a modern reproducibility test should look like this:
```
def test_repro_foo_stateless(self):
  seed = samplers.sanitize_seed(test_util.test_seed())
  ans1_ = something_random(seed=seed)
  ans2_ = something_random(seed=seed)
  ans1, ans2 = self.evaluate([ans1, ans2])
  self.assertEqual(ans1, ans2)
```

There is also a pattern for reproducibility tests for TF Classic seeding semantics, but I don't think it's important for this PR (or much in general, as long as we maintain the discipline of sanitizing seeds).

```suggestion
```

tensorflow_probability/python/experimental/mcmc/streaming_reductions_test.py
```diff
@@ -127,28 +127,29 @@ def test_calibration_warning(self):
 
   def test_seed_reproducibility(self):
     first_fake_kernel = RandomTransitionKernel()
+    second_fake_kernel = RandomTransitionKernel()
+    seed = test_util.test_seed()
```
axch, line 131:
```suggestion
    seed = samplers.sanitize_seed(test_util.test_seed())
```

tensorflow_probability/python/mcmc/sample.py
```diff
@@ -308,8 +308,9 @@ def trace_everything(states, previous_kernel_results):
        http://statweb.stanford.edu/~owen/reports/bestthinning.pdf
   """
   is_seeded = seed is not None
+  if tf.executing_eagerly():
```
axch, line 311: Let's pull the `sample_chain` test into a separate PR.

A reasonable heuristic for sizing PRs is that a PR is the unit of roll-backs (at least internally).  If it was necessary to roll back the test of `sample_chain` for some reason, would it also be logical to roll back `step_kernel`?  To me the answer seems to be "no", so I recommend splitting them.

Also, of course, my comments about seeds apply here too.

tensorflow_probability/python/experimental/mcmc/streaming_reductions.py
rupei: Thanks so much for the in-depth explanation! Everything makes a lot more sense now haha

tensorflow_probability/python/experimental/mcmc/streaming_reductions_test.py
```diff
@@ -0,0 +1,161 @@
...
+          seed=test_util.test_seed())
+    self.assertTrue(
+        any('supplied `TransitionKernel` is not calibrated.' in str(
+            warning.message) for warning in triggered))
+
+  def test_seed_reproducibility(self):
+    first_fake_kernel = RandomTransitionKernel()
+    second_fake_kernel = RandomTransitionKernel()
+    seed = test_util.test_seed()
+    if tf.executing_eagerly():
```
rupei, line 133: This if statement is necessary so the test also runs correctly in graph mode

tensorflow_probability/python/mcmc/sample.py
rupei: Sounds good!

tensorflow_probability/python/experimental/mcmc/streaming_reductions_test.py
axch: I expect that sanitizing unconditionally should work in both.  Do you know what goes wrong?

Note that one of the weird features of TF Classic PRNG semantics is the effect under multiple calls to `self.evaluate`.  The safe thing to do is to only ever compare results from the same `self.evaluate` call.  In this test, that would look like
```
last_state_t = step_kernel...  # No evaluate here
for num_steps in range
  first_final_state_t = step_kernel...  # Not here either
  second_final_state_t = step_kernel...  # Nor here
  last_state, first_final_state, second_final_state = self.evaluate([last_state_t, first_final_state_t, second_final_state_t])
  self.assertEqual ...
```
I'm actually not sure what the `for` loop will do here.  Ought to work as I suggested, but we could also fall back to building up a big list of Tensors, evaluating them once outside the `for`, and then writing another `for` for the assertions.

rupei: that did the trick!

tensorflow_probability/python/experimental/mcmc/streaming_reductions.py
```diff
@@ -0,0 +1,107 @@
+# Copyright 2018 The TensorFlow Probability Authors.
```
axch, line 1: Copyright year should be 2020.

tensorflow_probability/python/experimental/mcmc/streaming_reductions_test.py
```diff
@@ -0,0 +1,167 @@
+# Copyright 2018 The TensorFlow Probability Authors.
```
axch, line 1: Copyright year should be 2020 here too.

tensorflow_probability/python/experimental/mcmc/streaming_reductions_test.py
```diff
@@ -0,0 +1,167 @@
...
+  def test_simple_operation(self):
+    fake_kernel = TestTransitionKernel()
+    final_state, kernel_results = step_kernel(
+        num_steps=2,
+        current_state=0,
+        kernel=fake_kernel,
+        return_final_kernel_results=True
+    )
+    final_state = self.evaluate(final_state)
+    self.assertEqual(final_state, 2)
```
axch, line 93: This test doesn't actually pass, at least internally.  The error is
```
  File "/build/work/3186db25de8c50f31eeb96306c7279a3f85c/google3/runfiles/google3/third_party/py/absl/third_party/unittest3_backport/case.py", line 37, in testPartExecutor
    yield
  File "/build/work/3186db25de8c50f31eeb96306c7279a3f85c/google3/runfiles/google3/third_party/py/absl/third_party/unittest3_backport/case.py", line 162, in run
    testMethod()
  File "/build/work/3186db25de8c50f31eeb96306c7279a3f85c/google3/runfiles/google3/third_party/py/absl/testing/parameterized.py", line 282, in bound_param_test
    return test_method(self, **testcase_params)
  File "/build/work/3186db25de8c50f31eeb96306c7279a3f85c/google3/runfiles/google3/third_party/py/tensorflow_probability/python/internal/test_combinations.py", line 307, in decorated
    execute_test_method()
  File "/build/work/3186db25de8c50f31eeb96306c7279a3f85c/google3/runfiles/google3/third_party/py/tensorflow_probability/python/internal/test_combinations.py", line 295, in execute_test_method
    test_method(**kwargs_to_pass)
  File "/build/work/3186db25de8c50f31eeb96306c7279a3f85c/google3/runfiles/google3/third_party/py/tensorflow_probability/python/experimental/mcmc/streaming_reductions_test.py", line 88, in test_simple_operation
    self.assertEqual(kernel_results.counter_1, 2)
  File "<embedded stdlib>/unittest/case.py", line 513, in assertEqual
    assertion_func(first, second, msg=msg)
  File "<embedded stdlib>/unittest/case.py", line 506, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: <tf.Tensor: shape=(), dtype=int32, numpy=2> != 2
```

Presumably because a Tensor is not actually equal to 2.  Do you also see that on your end?  Should be possible to fix it by using assertAllEqual and/or tf.constant(2, dtype=tf.int32).

tensorflow_probability/python/experimental/mcmc/BUILD
```diff
@@ -41,6 +41,7 @@ py_library(
         ":particle_filter_augmentation",
         ":sample_sequential_monte_carlo",
         ":sequential_monte_carlo_kernel",
+        ":streaming_reductions",
```
axch, line 44: Let's actually put this code in tensorflow_probability/python/experimental/mcmc/sample.py, to imply that we are suggesting it will move to ensorflow_probability/python/mcmc/sample.py when it graduates.  The actual reducers and WithReductions (name TBD) should still go into streaming_reductions as you have here.

Up to you whether you think it will be simpler to make the rearrangement now or in a follow-up PR; also open to you feeling like it's not a good idea at all.

tensorflow_probability/python/experimental/mcmc/streaming_reductions_test.py
rupei: Talked to Alexey about this offline. There's some discrepancy between BUILDs that makes me unable to reproduce these errors. We concluded that it wouldn't be worthwhile for me to try to get around this (i.e. by building TensorFlow from source). Alexey will handle these errors in the pull in (thanks Alexey!)

tensorflow_probability/python/experimental/mcmc/BUILD
rupei: Sounds like a good idea. I didn't know if we actually wanted everything driver-like to go into sample.py or a different file. I can make the rearrangement now.

A note on the following reducers: I currently have them set up in different files (i.e. one for the base class, one for covariance reducer, one for WithReductions) to match how tfp.mcmc seems to organize it all right now. I think it's cleaner and better than mushing everything together, but let me know if it's better to group them together in experimental.

axch: Separate files are fine.

PR #1003, opened 2020-07-10
Adding a seed reproducibility test to `sample_chain`

rupei:
After implementing `step_kernel` in `tfp.experimental.mcmc`, I realized `sample_chain` has no seed reproducibility test. Adding one as a sanity check for a comprehensive test suite and to hopefully better illustrate seeding semantics with some precedent.

### DIFF
--- a/tensorflow_probability/python/mcmc/sample_test.py
+++ b/tensorflow_probability/python/mcmc/sample_test.py
@@ -37,6 +37,7 @@ TestTransitionKernelResults = collections.namedtuple(
 
 @test_util.test_all_tf_execution_regimes
 class TestTransitionKernel(tfp.mcmc.TransitionKernel):
+  """Fake deterministic `TransitionKernel` for testing purposes"""
 
   def __init__(self, is_calibrated=True, accepts_seed=True):
     self._is_calibrated = is_calibrated
@@ -58,6 +59,11 @@ class TestTransitionKernel(tfp.mcmc.TransitionKernel):
 
 
 class RandomTransitionKernel(tfp.mcmc.TransitionKernel):
+   """Fake `TransitionKernel` that randomly assigns the next state.
+
+  Regardless of the current state, the `one_step` method will always
+  randomly sample from a Reyleigh Distribution.
+  """
 
   def __init__(self, is_calibrated=True, accepts_seed=True):
     self._is_calibrated = is_calibrated
--- a/tensorflow_probability/python/mcmc/sample_test.py
+++ b/tensorflow_probability/python/mcmc/sample_test.py
@@ -35,7 +35,6 @@ TestTransitionKernelResults = collections.namedtuple(
     'TestTransitionKernelResults', 'counter_1, counter_2')
 
 
-@test_util.test_all_tf_execution_regimes
 class TestTransitionKernel(tfp.mcmc.TransitionKernel):
   """Fake deterministic `TransitionKernel` for testing purposes"""
 
@@ -59,7 +58,7 @@ class TestTransitionKernel(tfp.mcmc.TransitionKernel):
 
 
 class RandomTransitionKernel(tfp.mcmc.TransitionKernel):
-   """Fake `TransitionKernel` that randomly assigns the next state.
+  """Fake `TransitionKernel` that randomly assigns the next state.
 
   Regardless of the current state, the `one_step` method will always
   randomly sample from a Reyleigh Distribution.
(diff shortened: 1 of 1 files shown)


tensorflow_probability/python/mcmc/sample_test.py
```diff
@@ -309,6 +334,29 @@ def log_prob(x):
...
+    for num_results in range(2, 5):
+      first_final_state = tfp.mcmc.sample_chain(
+          num_results=num_results,
+          current_state=0.,
+          kernel=first_fake_kernel,
+          seed=seed,
+      )
+      second_final_state = tfp.mcmc.sample_chain(
+          num_results=num_results,
+          current_state=1., # difference should be irrelevant
```
axch, line 350:
```suggestion
          current_state=1.,  # difference should be irrelevant
```

tensorflow_probability/python/mcmc/sample_test.py
```diff
@@ -56,6 +58,29 @@ def is_calibrated(self):
...
+  """
+
+  def __init__(self, is_calibrated=True, accepts_seed=True):
+    self._is_calibrated = is_calibrated
+    self._accepts_seed = accepts_seed
+
+  def one_step(self, current_state, previous_kernel_results, seed=None):
+    if seed is not None and not self._accepts_seed:
+      raise TypeError('seed arg not accepted')
+    random_next_state = tfp.random.rayleigh((), seed=seed)
```
SiegeLordEx, line 75: I know this is just a test, but I'd recommend doing `tfp.random.rayleigh(current_state.shape, seed=seed)`, so that this kernel supports multiple chains.

tensorflow_probability/python/mcmc/sample_test.py
```diff
@@ -56,6 +58,29 @@ def is_calibrated(self):
...
+class RandomTransitionKernel(tfp.mcmc.TransitionKernel):
+  """Fake `TransitionKernel` that randomly assigns the next state.
+
+  Regardless of the current state, the `one_step` method will always
+  randomly sample from a Reyleigh Distribution.
+  """
+
+  def __init__(self, is_calibrated=True, accepts_seed=True):
+    self._is_calibrated = is_calibrated
+    self._accepts_seed = accepts_seed
```
SiegeLordEx, line 70: You don't actually use these args?

tensorflow_probability/python/mcmc/sample_test.py
```diff
@@ -309,6 +334,29 @@ def log_prob(x):
     # Checking that shape inference doesn't fail.
     sample(2)
 
+  def testSeedReproducibility(self):
+    first_fake_kernel = RandomTransitionKernel()
+    second_fake_kernel = RandomTransitionKernel()
+    seed = samplers.sanitize_seed(test_util.test_seed())
+    for num_results in range(2, 5):
```
SiegeLordEx, line 341: If you really feel this loop is necessary, then I'd recommend using parameterized testing for this. The reason being is that if this test were to fail as it is now, you'd have no idea what `num_results` was when it failed (you could log it, certainly, but the idiomatic way to solve this issue is to use parameterized tests).

rupei: good point! The loop is probably actually not that necessary.

tensorflow_probability/python/mcmc/sample_test.py
rupei: Very true. The reason why I kept them was because the Transition Kernel needs to have an `is_calibrated` property anyways, and I wanted to set it up so that if this object is used for future tests, it's already all good to go. Alternatively, I could get rid of a formal `__init__`, remove the seed check in `one_step`, and have an `is_calibrated` property that does nothing (i.e. returns None). I think I prefer how it looks currently since it just seems more complete even if I might not be explicitly using those arguments. Thoughts?

tensorflow_probability/python/mcmc/sample_test.py
rupei: Good point!

tensorflow_probability/python/mcmc/sample_test.py
SiegeLordEx: In general, it's probably best to only write code for something you have a clear plan for. Future proofing rarely works. That said, this is a minor thing, you can keep this as is if you prefer.

rupei: Gotcha. Thanks for bringing it up though, I'll keep that in mind in the future!

PR #1012, opened 2020-07-17
Add Reducer base class and Covariance Reducer to tfp.experimental.mcmc

rupei:
Defines the `Reducer` base class -- a slightly different contract than a `TransitionKernel` for reductions (and soon to be wrapped in a Transition Kernel itself) and a pre-packaged `CovarianceReducer` specifically for streaming covariance using the algorithms in tfp.experimental.stats.


tensorflow_probability/python/experimental/mcmc/reducer.py
```diff
@@ -0,0 +1,105 @@
...
+
+  @abc.abstractmethod
+  def initialize(self, initial_chain_state, initial_inner_kernel_results):
+    """Initializes a reducer state corresponding to the stream of no samples.
+
+    This is an abstract method and must be overridden by subclasses.
+
+    Args:
+      initial_chain_state: `Tensor` or Python `list` of `Tensor`s representing
+        the current state(s) of the Markov chain(s).
```
axch, line 53:
```suggestion
      initial_chain_state: `Tensor` or Python `list` of `Tensor`s representing
        the current state(s) of the Markov chain(s).  This is to be used for
        any needed shape, dtype, etc information, but should not be considered
        part of the stream being reduced.
```

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,185 @@
...
+class CovarianceReducer(reducer_base.Reducer):
+  """`Reducer` that computes a running covariance.
+
+  As with all reducers, CovarianceReducer does not hold state information;
+  rather, it stores supplied metadata. Intermediate calculations are held in
+  a state object, which is returned via `initialize` and `one_step` method
+  calls.
+  """
+
+  def __init__(self, shape, event_ndims=None, dtype=tf.float32, name=None):
```
axch, line 44: Can we infer the shape(s) and dtype(s) from the first state that comes in to `initialize`, instead of requiring them here?

This would be a way the Reducer packaging would make this computation a little easier to use.

Also, can we support computing variances over structured latent spaces (like the Cox process model that has several separate Tensors in its latents)?  As a first simplification, let's assume that we're not trying to compute any cross-products across Tensors.  Then we should be able to get away with packing this Reducer with as many `RunningCovariance` instances as there are Tensors in the latent state (which becomes available to `initialize`), and letting them all operate independently.

Does this make sense?  Want to talk about it?

tensorflow_probability/python/experimental/mcmc/reducer.py
```diff
@@ -0,0 +1,105 @@
...
+        structure as `current_reducer_state`.
+    """
+    raise NotImplementedError()
+
+  def finalize(self, final_state):
+    """Finalizes target statistic calculation from the `final_state`.
+
+    This is not an abstract method, but the base implementation is an
+    identity function of the `final_state`. Hence, it should be overriden
+    for more complex streaming calculations.
```
axch, line 96:
```suggestion
    This is an identity function of the `final_state` by default. Subclasses can override
    it for streaming calculations whose running state is not the same as the desired
    result.
```

tensorflow_probability/python/experimental/mcmc/reducer.py
```diff
@@ -0,0 +1,105 @@
...
+  @abc.abstractmethod
+  def one_step(self, sample, current_reducer_state, previous_kernel_results):
+    """Takes one step of the `Reducer`.
+
+    This is an abstract method and must be overridden by subclasses.
+
+    Args:
+      sample: Incoming sample with shape and dtype compatible with those
+        specified when initializing the `Reducer` and with the
+        `current_reducer_state`.
```
axch, line 75:
```suggestion
      sample: Incoming sample with shape and dtype compatible with the
        `initial_chain_state` with which the `current_reducer_state` was
        produced by `initialize`.
```

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,185 @@
...
+      variables, and computes their variances in batch.  The shape of the result
+      is `[5, 7]`.
+    - `event_ndims=1` treats the samples as a `[5]` batch of vector random
+      variables of shape `[7]`, and computes their covariances in batch.  The
+      shape of the result is `[5, 7, 7]`.
+    - `event_ndims=2` treats the samples as a single random variable of
+      shape `[5, 7]` and computes its covariance.  The shape of the result
+      is `[5, 7, 5, 7]`.
+
+    Args:
```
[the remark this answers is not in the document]
rupei: Yup, it will be added in a future PR

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,185 @@
...
+    with tf.name_scope(
+        mcmc_util.make_name(self.name, 'covariance_reducer', 'one_step')):
+      return self.strm.update(current_reducer_state, sample, axis=axis)
+
+  def finalize(self, final_state, ddof=0):
+    """Finalizes covariance calculation from the `final_state`.
+
+    final_state: `RunningCovarianceState` that represents the final state of
+        running statistics.
+    ddof: Requested dynamic degrees of freedom. For example, use `ddof=0`
```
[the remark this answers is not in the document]
rupei: An alternative is to have the user pass in `finalize` kwargs in `sample_fold`. For example having

```
def sample_fold(..., **kwargs):
  ...
  reduction_result = nest.map_structure_up_to(
      reducers,
      lambda r, s: r.finalize(s, **kwargs),
      reducers,
      final_kernel_results.streaming_calculations,
      check_types=False)
  ...
``` 

That `sample_fold` use-case allows users to create multi-purposed reducers. For example, one covariance reducer is enough to support all ddof finalizes, instead of needing to make multiple. Should we opt for this alternative instead?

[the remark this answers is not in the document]
axch: As written, this seems to call for every reducer being able to accept any kwarg, so that a hypothetical ESSReducer's finalize would also have to accept (and ignore) ddof, to cover the situation where a user wants both covariance and ESS, and wants to pass a ddof argument to the covariance.

I'm also not sure what benefit you are looking for.  Reusing the same CovarianceReducer object across multiple calls to sample_fold, with different ddof arguments?  Why?  CovarianceReducer is stateless; making new ones is cheap computationally, and I'm not sure the potential usability savings is worth the mental complexity of plumbing arguments through sample_fold to finalize.  Open to being persuaded otherwise, however.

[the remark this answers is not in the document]
rupei: Good points. I wanted to bring this possibility up since it popped into my head, but I think I'm now convinced that what I just described isn't really a good alternative.

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -61,46 +67,92 @@ def __init__(self, shape, event_ndims=None, dtype=tf.float32, name=None):
...
+        self._parameters['event_ndims'] = tf.nest.map_structure(
+            lambda chain_state: ps.rank(chain_state),
+            initial_chain_state
+        )
+      elif not nest.is_sequence(self.event_ndims):
+        self._parameters['event_ndims'] = tf.nest.map_structure(
+            lambda _: self.event_ndims,
+            initial_chain_state,
+        )
+      self.strm = nest.map_structure_up_to(
```
axch, line 143: Probably better to name self.strm in the plural, since the rest of the code will treat it as a collection of streaming computations, rather than just one.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -109,6 +109,10 @@ class RunningCovariance(object):
   which includes intermediate calculations, are held in a
   `RunningCovarianceState` as returned via `initialize` and `update` method
   calls.
+
+  `RunningCovariance` is meant to serve more general streaming covariance needs.
+  For a specialized version that fits MCMC sampling, see `CovarianceReducer` in
+  `tfp.experimental.mcmc`.
```
axch, line 115:
```suggestion
  `RunningCovariance` is meant to serve general streaming covariance needs.
  For a specialized version that fits streaming over MCMC samples, see
  `CovarianceReducer` in `tfp.experimental.mcmc`.
```

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -246,6 +250,10 @@ class RunningVariance(RunningCovariance):
   which includes intermediate calculations, are held in a
   `RunningCovarianceState` as returned via `initialize` and `update` method
   calls.
+
+  `RunningVariance` is meant to serve more general streaming variance needs.
```
axch, line 254: Same wording fix here.

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -61,46 +67,92 @@ def __init__(self, shape, event_ndims=None, dtype=tf.float32, name=None):
...
+
     Args:
-      initial_chain_state: `Tensor` or Python `list` of `Tensor`s representing
-        the current state(s) of the Markov chain(s). For streaming covariance,
-        this argument has no influence on computation. Hence, by default, it
-        is `None`.
+      initial_chain_state: A (possibly nested) structure of `Tensor`s or Python
+        `list`s of `Tensor`s representing the current state(s) of the Markov
+        chain(s). For streaming covariance, this argument has no influence on
+        computation. Hence, by default, it is `None`. However, this argument is
```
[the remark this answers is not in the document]
rupei: It's probably best to just steer away from stating this anymore. While it's still true that the `initial_chain_state` isn't really influencing streaming calculation (it's just used as an object to infer metadata in this case), I see how that can be misinterpreted.

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,260 @@
...
+from __future__ import print_function
+
+# Dependency imports
+import tensorflow.compat.v2 as tf
+
+from tensorflow_probability.python.experimental.mcmc import reducer as reducer_base
+from tensorflow_probability.python.experimental.stats import sample_stats
+from tensorflow_probability.python.internal import prefer_static as ps
+from tensorflow_probability.python.mcmc.internal import util as mcmc_util
+from tensorflow.python.util import nest
```
SiegeLordEx, line 28: This needs a: `# pylint: disable=g-direct-tensorflow-import` at the end of the line.

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,260 @@
...
+      is `[5, 7]`.
+    - `event_ndims=1` treats the samples as a `[5]` batch of vector random
+      variables of shape `[7]`, and computes their covariances in batch.  The
+      shape of the result is `[5, 7, 7]`.
+    - `event_ndims=2` treats the samples as a single random variable of
+      shape `[5, 7]` and computes its covariance.  The shape of the result
+      is `[5, 7, 5, 7]`.
+
+    For nested latent samples, both `event_ndims` and `ddof` must be either
+    a single value or an identical structure. For example, if the chain state
```
SiegeLordEx, line 71: Identical to what?

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,260 @@
...
+    is the tuple of dictionaries ({pairA, pairB}, {pairC}), one could use
+    scalar values of `event_ndims` (integer or `None`) and `ddof` (integer) that
+    apply to all states in the latent. If not, `event_ndims` and `ddof` must
+    have structures that ressemble ({keyA: argA, keyB: argB}, {keyC: argC}).
```
SiegeLordEx, line 75: "resemble"

SiegeLordEx: Or rephase this, as it's not clear what "resemble" would mean here. `tf.nest` has very limited duck-typing of structures.

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,260 @@
...
+      self._parameters['shape'] = tf.nest.map_structure(
+          lambda chain_state: chain_state.shape,
+          initial_chain_state)
+      self._parameters['dtype'] = tf.nest.map_structure(
+          lambda chain_state: chain_state.dtype,
+          initial_chain_state)
+      if self.event_ndims is None:
+        self._parameters['event_ndims'] = tf.nest.map_structure(
+            ps.rank, initial_chain_state)
+      elif not nest.is_sequence(self.event_ndims):
```
SiegeLordEx, line 132: is_nested?

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,260 @@
...
+        self._parameters['event_ndims'] = tf.nest.map_structure(
+            lambda _: self.event_ndims,
+            initial_chain_state,
+        )
+      self.strms = nest.map_structure_up_to(
```
SiegeLordEx, line 137: strms seems like an impenetrable name.

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,260 @@
...
+    Returns:
+      state: `RunningCovarianceState` representing a stream of no inputs and
+        with identical structure to the `initial_chain_state`.
+    """
+    with tf.name_scope(
+        mcmc_util.make_name(self.name, 'covariance_reducer', 'initialize')):
+      initial_chain_state = tf.nest.map_structure(
+          tf.convert_to_tensor,
+          initial_chain_state)
+      self._parameters['shape'] = tf.nest.map_structure(
```
SiegeLordEx, line 123: This doesn't feel right, why are we mutating the class state in a method? In TFP we are *very* careful not to use mutable state, only violating this directive for caching purposes (and internal mutabiity of Variables).

I think the right solution here is to re-compute all these properties in initialize/one_step/finalize, creating new `RunningCovariance` objects each time.

Either way, storing these side-effects in _parameters is incorrect as well, as they're not parameters (which are exclusively understood to be the arguments to the class initializer in TFP).

SiegeLordEx: See _prepare_args pattern in a few places in TFP for what I mean.

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,260 @@
...
+          check_types=False,
+      )
+      return tf.nest.map_structure(
+          lambda strm: strm.initialize(),
+          self.strms)
+
+
+  def one_step(
+      self,
+      sample,
```
SiegeLordEx, line 151: chain_state? It's a bit presumptive to call these samples. It's what you called this in initialize as well, so at least we should be consistent across the board.

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,260 @@
...
+      current_reducer_state,
+      previous_kernel_results=None,
+      axis=None):
+    """Update the `current_reducer_state` with a new sample.
+
+    Chunking semantics are similar to those of batching and are specified by the
+    `axis` parameter. If chunking is enabled (axis is not `None`), all elements
+    along the specified `axis` will be treated as separate samples. If a
+    single scalar value is provided for a non-scalar sample structure, that
+    value will be used for all samples in the latent. If not, an identical
```
SiegeLordEx, line 161: latent? Maybe "for all elements in the structure"

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,260 @@
...
+        chain(s). It is used to infer the shape and dtype of future samples.
+      initial_kernel_results: A (possibly nested) `tuple`, `namedtuple` or
+        `list` of `Tensor`s representing internal calculations made in a related
+        `TransitionKernel`. For streaming covariance, this argument also has no
+        influence on computation; hence, it is also `None` by default.
+        Nonetheless, this argument is still accepted as it will be supplied by
+        the Streaming MCMC framework.
+
+    Returns:
+      state: `RunningCovarianceState` representing a stream of no inputs and
```
SiegeLordEx, line 115: I think you should create a brand new named tuple that contains the RCS inside of it, rather than expose the details of the implementation. This'll slightly simplify future extensions and just feels nicer from an encapsulation standpoint.

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,260 @@
...
+    """
+    with tf.name_scope(
+        mcmc_util.make_name(self.name, 'covariance_reducer', 'one_step')):
+      if not nest.is_sequence(axis):
+        axis = tf.nest.map_structure(
+            lambda _: axis,
+            sample,
+        )
+      return nest.map_structure_up_to(
+          sample,
```
SiegeLordEx, line 194: dtype?

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,260 @@
...
+        (`axis` is None).
+
+    Returns:
+      new_state: `RunningCovarianceState` with updated running statistics and
+        identical structure to the `current_reducer_state`.
+    """
+    with tf.name_scope(
+        mcmc_util.make_name(self.name, 'covariance_reducer', 'one_step')):
+      if not nest.is_sequence(axis):
+        axis = tf.nest.map_structure(
```
SiegeLordEx, line 189: Have you considered using `tfp.python.internal.nest_util.broadcast_structure` here (and elsewhere)?

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,260 @@
...
+
+    Args:
+      sample: A (possibly nested) structure of incoming sample(s) with shape
+        and dtype compatible with those used to initialize the
+        `current_reducer_state`.
+      current_reducer_state: A (possibly nested) structure of
+        `RunningCovarianceState`s representing the current state of the running
+        covariance.
+      previous_kernel_results: A (possibly nested) `tuple`, `namedtuple` or
+        `list` of `Tensor`s representing internal calculations made in a related
```
SiegeLordEx, line 172: You already have a nice clause for this. I.e. just say "A (possibly nested) structure of ..."

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -0,0 +1,260 @@
...
+
+    Args:
+      initial_chain_state: A (possibly nested) structure of `Tensor`s or Python
+        `list`s of `Tensor`s representing the current state(s) of the Markov
+        chain(s). It is used to infer the shape and dtype of future samples.
+      initial_kernel_results: A (possibly nested) `tuple`, `namedtuple` or
+        `list` of `Tensor`s representing internal calculations made in a related
+        `TransitionKernel`. For streaming covariance, this argument also has no
+        influence on computation; hence, it is also `None` by default.
+        Nonetheless, this argument is still accepted as it will be supplied by
```
SiegeLordEx, line 111: Sure, but that's somewhat backwards. The reason its here is because it's in the super-class and we're following that interface.

tensorflow_probability/python/experimental/mcmc/covariance_reducer_test.py
```diff
@@ -0,0 +1,149 @@
...
+    final_cov = self.evaluate(cov_reducer.finalize(state))
+    self.assertAllClose(final_cov, tf.zeros((2, 3, 2, 3)), rtol=1e-6)
+
+  def test_nested_chain_state(self):
+    cov_reducer = tfp.experimental.mcmc.CovarianceReducer(event_ndims=0)
+    chain_state = ({'one': tf.ones((2, 3)), 'zero': tf.zeros((2, 3))},
+                   {'two': tf.ones((2, 3)) * 2})
+    state = cov_reducer.initialize(chain_state)
+    _, state = tf.while_loop(
+        lambda i, _: tf.less(i, 10),
```
SiegeLordEx, line 121: `i < 10` (and elsewhere)

tensorflow_probability/python/experimental/mcmc/covariance_reducer_test.py
```diff
@@ -0,0 +1,149 @@
...
+    self.assertEqual(cov_reducer.event_ndims, 1)
+    self.assertEqual(cov_reducer.dtype, tf.float64)
+    self.assertEqual(cov_reducer.ddof, 1)
+
+  def test_tf_while(self):
+    cov_reducer = tfp.experimental.mcmc.CovarianceReducer()
+    state = cov_reducer.initialize(tf.ones((2, 3)))
+    _, state = tf.while_loop(
+        lambda i, _: tf.less(i, 100),
+        lambda i, state: (i+1, cov_reducer.one_step(
```
SiegeLordEx, line 108: Please add spaces around binary ops (here and elsewhere)

tensorflow_probability/python/experimental/mcmc/reducer.py
```diff
@@ -0,0 +1,107 @@
...
+        `list` of `Tensor`s representing internal calculations made in a related
+        `TransitionKernel`. This allows for introspection of deeper layers of
+        `TransitionKernel`s that have bearing to the nature of the initial
+        reducer state.
+
+    Returns:
+      init_state: `tuple`, `namedtuple` or `list` of `Tensor`s representing the
+        stream of no samples.
+    """
+    raise NotImplementedError()
```
SiegeLordEx, line 66: Pretty sure you can delete this. The whole point of using `abstractmethod` is ensuring that this will always be implemented.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -109,6 +109,10 @@ class RunningCovariance(object):
   which includes intermediate calculations, are held in a
   `RunningCovarianceState` as returned via `initialize` and `update` method
   calls.
+
+  `RunningCovariance` is meant to serve general streaming covariance needs.
```
SiegeLordEx, line 113: I don't think this forward-reference is necessary.

tensorflow_probability/python/experimental/mcmc/BUILD
```diff
@@ -352,3 +354,39 @@ py_test(
...
+        "//tensorflow_probability/python/experimental/stats:sample_stats",
+        "//tensorflow_probability/python/mcmc/internal:util",
+    ],
+)
+
+# py2and3
+py_test(
+    name = "covariance_reducer_test",
+    size = "small",
+    shard_count = 5,
```
SiegeLordEx, line 383: I don't think it needs any shards.

tensorflow_probability/python/experimental/mcmc/BUILD
```diff
@@ -352,3 +354,39 @@ py_test(
         "//tensorflow_probability/python/internal:test_util",
     ],
 )
+
+py_library(
```
SiegeLordEx, line 358: This isn't anything you need to fix, but I am annoyed that this directory isn't multi-substrate (to support the JAX backend).

tensorflow_probability/python/experimental/mcmc/reducer.py
```diff
@@ -0,0 +1,107 @@
...
+        reducer state (i.e. updating based on a value in the kernel results of
+        some `TransitionKernel`).
+
+    Returns:
+      new_state: The new reducer state after updates. This has the same type and
+        structure as `current_reducer_state`.
+    """
+    raise NotImplementedError()
+
+  def finalize(self, final_state):
```
SiegeLordEx, line 93: final_reducer_state

tensorflow_probability/python/experimental/mcmc/reducer.py
```diff
@@ -0,0 +1,107 @@
...
+        any needed (shape, dtype, etc.) information, but should not be
+        considered part of the stream being reduced.
+      initial_inner_kernel_results: A (possibly nested) `tuple`, `namedtuple` or
+        `list` of `Tensor`s representing internal calculations made in a related
+        `TransitionKernel`. This allows for introspection of deeper layers of
+        `TransitionKernel`s that have bearing to the nature of the initial
+        reducer state.
+
+    Returns:
+      init_state: `tuple`, `namedtuple` or `list` of `Tensor`s representing the
```
SiegeLordEx, line 63: init_reducer_state

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
axch: Yeah, I was wondering what you would think of the mutation.

For my part, I had let it slide because the expected use pattern leads to the mutation occurring outside the sampler's `while_loop`, but you're right, `prepare_args` is probably better.

tensorflow_probability/python/experimental/mcmc/reducer.py
rupei: good catch! The `TransitionKernel` base class also has this redundancy then.

tensorflow_probability/python/experimental/mcmc/covariance_reducer_test.py
rupei: good catches. Sorry for the repeat style mistake, the code was written before the last comment you made on it and I forgot to fix it outside the scope of that PR.

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
rupei: I actually think it's more confusing to call it `chain_state`. I think under the overall framework, it's natural to call these samples (i.e the idea is in `WithReductions` some new sample is taken from the inner kernel then all these `one_step` methods are invoked). Leaving it as `chain_state` might make it less clear that this is supposed to be a new sample? I think in `initialize` it makes sense to refer to it as the initial chain state as it's really not a sample (we don't count it as one). Interested in both what you think about that thought though; might be a good opportunity for learning about better naming strategies. Overall, it's a trivial point that I don't mind changing.

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: I'm leaning towards keeping it simply because it can only do more good than harm (right?)

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
rupei: Didn't even know this existed. Helps a lot!

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -34,6 +37,10 @@
 ]
 
 
+CovarianceReducerState = collections.namedtuple(
+    'CovarianceReducerState', 'init_structure, cov_state')
```
SiegeLordEx, line 41: It's not a huge deal, but could you extract `init_structure` from `cov_state`, somehow? It's a little bit wasteful to carry `init_structure` which isn't changing between steps.

rupei: I tried playing around with it for the good part of the morning, but extracting from `cov_state` runs into the issue of some arbitrarily nested structure. In say the case of a tuple of dictionaries of reducer states, without keeping some reference to the structure of that initial chain state, we then don't know exactly how to extract a Tensor that would help. I'm sure there's a workaround, but that would probably lead to much more complex code.

SiegeLordEx: Would it not just be `map_structure(lambda cs: cs.mean, cov_state)`? I'm confused when a case would happen where the structure of `cov_state` wouldn't match `init_structure`.

rupei: Wouldn't that actually go as deep as possible until it hits `cov_state`'s Tensors? It would be calling something like Tensor.mean (which doesn't work). We still need some `map_structure_up_to` to specify the shallow tree. However, in `finalize` we can't do that because all we have is the final reducer state.

SiegeLordEx: Ah yes, of course! You're right. I guess let's keep the `init_structure` for now, and we can think of ways of getting rid of it later.

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
rupei: Good points! I like `new_chain_state`.

Issue #1033, opened 2020-07-30
`smart_for_loop` is using a Python for loop for large statically known `loop_num_iter`


rupei:
`smart_for_loop` is a utility function in tensorflow_probability/python/mcmc/internal/util.py that first checks if the number of loop iterations is statically known. If so, it executes that many iterations using a Python for loop; if not, it opts for a `tf.while_loop`. The issue arises when `smart_for_loop` is called on some large statically known value in graph mode. With many iterations, compilation becomes very expensive and takes ridiculously long; opting for a `tf.while_loop` would be much more optimal.

This issue has flown under the radar since `sample_chain` is just about the only place where this utility is used. In the status quo, `sample_chain` uses a Tensor schedule that dictates the number of steps to be taken before returning a sample. This means the number of iterations passed in `smart_for_loop` is never static. Hence, the Python loop is never executed.

The fix would be to only use the Python for loop if the number of iterations is statically known to be 0 or 1 (thus benefiting from loop unrolling). I'll go ahead and make this fix, but am still submitting this issue for reference.

PR #1035, opened 2020-07-31
Fixing ridiculous loop unrolling in `smart_for_loop`

rupei:
A direct fix of the issue described in https://github.com/tensorflow/probability/issues/1033. I've also added a test for `tfp.experimental.mcmc.step_kernel` to ensure that future building blocks in the streaming MCMC framework will work as intended

### DIFF
--- a/tensorflow_probability/python/experimental/mcmc/sample_test.py
+++ b/tensorflow_probability/python/experimental/mcmc/sample_test.py
@@ -170,17 +170,16 @@ class StepKernelTest(test_util.TestCase):
       self.assertNotEqual(first_final_state, last_state)
       last_state_t = first_final_state_t
 
-  @test_util.test_graph_and_eager_modes
-  def test_smart_for_loop_uses_tf_function(self):
+  @test_util.test_graph_mode_only
+  def test_smart_for_loop_uses_tf_while(self):
     dtype = np.float32
     true_mean = dtype([0, 0])
     true_cov = dtype([[1, 0.5],
                       [0.5, 1]])
     true_cov_chol = np.linalg.cholesky(true_cov)
-    num_results = 1000
+    num_results = 100
     counter = collections.Counter()
 
-    @tf.function
     def target_log_prob(x, y):
       counter['target_calls'] += 1
       z = tf.stack([x, y], axis=-1) - true_mean
@@ -197,7 +196,7 @@ class StepKernelTest(test_util.TestCase):
         return_final_kernel_results=False,
         seed=test_util.test_seed())
 
-    self.assertAllEqual(dict(target_calls=1), counter)
+    self.assertAllEqual(dict(target_calls=4), counter)
 
 
 if __name__ == '__main__':
--- a/tensorflow_probability/python/experimental/mcmc/sample_test.py
+++ b/tensorflow_probability/python/experimental/mcmc/sample_test.py
@@ -20,8 +20,9 @@ from __future__ import print_function
 
 import collections
 import warnings
+
 # Dependency imports
-from absl.testing import parameterized
+import numpy as np
 
 import tensorflow.compat.v2 as tf
 import tensorflow_probability as tfp
@@ -72,7 +73,7 @@ class RandomTransitionKernel(tfp.mcmc.TransitionKernel):
   def one_step(self, current_state, previous_kernel_results, seed=None):
     if seed is not None and not self._accepts_seed:
       raise TypeError('seed arg not accepted')
-    random_next_state = tfp.random.rayleigh((1,), seed=seed)
+    random_next_state = tfp.random.rayleigh((), seed=seed)
     return random_next_state, previous_kernel_results
 
   @property
@@ -80,9 +81,9 @@ class RandomTransitionKernel(tfp.mcmc.TransitionKernel):
     return self._is_calibrated
 
 
-@test_util.test_all_tf_execution_regimes
 class StepKernelTest(test_util.TestCase):
 
+  @test_util.test_all_tf_execution_regimes
   def test_simple_operation(self):
     fake_kernel = TestTransitionKernel()
     final_state, kernel_results = step_kernel(
@@ -96,6 +97,7 @@ class StepKernelTest(test_util.TestCase):
     self.assertEqual(kernel_results.counter_1, 2)
     self.assertEqual(kernel_results.counter_2, 4)
 
+  @test_util.test_all_tf_execution_regimes
   def test_defined_pkr(self):
     fake_kernel = TestTransitionKernel()
     init_pkr = TestTransitionKernelResults(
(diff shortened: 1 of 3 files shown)


tensorflow_probability/python/mcmc/internal/util.py
```diff
@@ -332,8 +332,10 @@ def smart_for_loop(loop_num_iter, body_fn, initial_loop_vars,
   """
   with tf.name_scope(name or 'smart_for_loop'):
     loop_num_iter_ = tf.get_static_value(loop_num_iter)
-    if (loop_num_iter_ is None or tf.executing_eagerly() or
```
axch, line 335: Please add a comment explaining the loop_num_iter_ > 1 check.  The existence of Issue #1033 helps make the comment concise, but it should be there.

axch: A comment around line 350 that by that point `loop_num_iter_` is no more than 1 would also help make this code clearer.

tensorflow_probability/python/mcmc/internal/util_test.py
```diff
@@ -196,8 +196,28 @@ def fn(x, y):
 @test_util.test_all_tf_execution_regimes
 class SmartForLoopTest(test_util.TestCase):
 
-  def test_python_for_loop(self):
+  @parameterized.parameters(0, 1)
```
axch, line 199: Why split this test into two, rather than testing on 0, 1, 10?  The expected number of calls then becomes `max(1, loop_iters)`, does it not (except for JAX and Eager mode, as usual)?

tensorflow_probability/python/experimental/mcmc/sample_test.py
```diff
@@ -166,6 +170,35 @@ def test_seed_reproducibility(self):
...
+  def test_smart_for_loop_uses_tf_function(self):
+    dtype = np.float32
+    true_mean = dtype([0, 0])
+    true_cov = dtype([[1, 0.5],
+                      [0.5, 1]])
+    true_cov_chol = np.linalg.cholesky(true_cov)
+    num_results = 1000
+    counter = collections.Counter()
+
+    @tf.function
```
axch, line 183: With tf.function here, what is this test actually testing?  As written, it would presumably pass regardless of the change made in this PR, except for possible problems with timeouts due to large graphs induced by HMC.one_step.  Does it make sense to instead test just graph mode, without the tf.function annotation, and expect 4 rather than 1000 target calls?

tensorflow_probability/python/mcmc/internal/util.py
```diff
@@ -332,8 +332,10 @@ def smart_for_loop(loop_num_iter, body_fn, initial_loop_vars,
...
-        control_flow_util.GraphOrParentsInXlaContext(
+    if (loop_num_iter_ is None
+        or tf.executing_eagerly()
+        or loop_num_iter_ > 1
```
SiegeLordEx, line 337: I think this 1 is probably too conservative. I've seen numbers <100 to be good compromises, e.g. https://github.com/google/jax/commit/0a45fc3b9bf8ae53b371d0e1175ff37127a6bc34. Maybe make it configurable? Have you tested with anything other than 1?

axch: Configurability is reasonable.  That said, I am pretty strongly in favor of being very conservative here by default, because the function being unrolled is unknown.  If I understand your example correctly, the loop they are unrolling is just a multiply-add, where unrolling to 64 iterations makes sense.  But if the loop body is an HMC transition on the LogGaussianCoxProcess model, which is where this came up for us, then unrolling even 2 of them has a decent chance of being counterproductive (but I am not aware of experiments to that effect).

SiegeLordEx: Yeah, that makes perfect sense. Was that still true even after tf.function decorating? I'm be fine with configurability + default of 1 loop.

tensorflow_probability/python/mcmc/internal/util.py
rupei: Ah right. My bad -- forgot about that comment

tensorflow_probability/python/mcmc/internal/util_test.py
rupei: That is the expected behavior unless the `loop_iters=0`, which is probably the test case I wanted to isolate.

tensorflow_probability/python/experimental/mcmc/sample_test.py
rupei: Yup, this makes much more sense!

tensorflow_probability/python/mcmc/internal/util.py
axch: I don't know.  We didn't conduct a formal study, except to note that going from 4 copies to 1 (by tf.function decorating the density) took 50% = 1.5s off the graph construction time.  That's not perfectly comparable, because the issue here is about whether to replicate (HMC kernel + 4 * density) num_steps times or once, but in any case, I think we are in agreement.

tensorflow_probability/python/mcmc/internal/util_test.py
```diff
@@ -196,25 +196,48 @@ def fn(x, y):
...
+      def body(x):
+        counter['body_calls'] += 1
+        return [x + 1]
+
+      result = util.smart_for_loop(
+          loop_num_iter=n, body_fn=body, initial_loop_vars=[tf.constant(1)])
+      if tf.executing_eagerly() and not JAX_MODE:  # JAX always traces loops
+        self.assertEqual(iters, counter['body_calls'])
+      else:
+        self.assertEqual(1, counter['body_calls'])
```
axch, line 237: Would the following have the same test coverage and let you include the 0 case in `@parameterzed` instead of having to copy it separately?

```suggestion
      if JAX_MODE:  # JAX always traces loop bodies exactly once
        self.assertEqual(1, counter['body_calls'])
      elif tf.executing_eagerly():
        self.assertEqual(iters, counter['body_calls'])
      else:
        self.assertEqual(max(1, iters), counter['body_calls'])
```

tensorflow_probability/python/mcmc/internal/util.py
axch: @rupei Clarifying: consensus here is to add an optional argument to `smart_for_loop` named something like `unroll_threshold=1`, and use that instead of the hardcoded 1 limit on unrolls.

tensorflow_probability/python/mcmc/internal/util_test.py
rupei: Yeah something slightly different works since 0 iterations means the function is never called even in graph mode (i.e. always `counter['body_calls'] = 0`). I wanted to keep them separate as the 0 case is special, but eliminating code duplication is also important.

Issue #1031, opened 2020-07-29
Awkward thinning/burn-in semantics for `sample_chain`


rupei:
Currently, `sample_chain` implements thinning and burn-in as follows:
- the first sample will step the inner kernel `num_burnin_steps` + 1 times
- any other samples will step `num_steps_between_results` + 1 times

This runs into the issue of never accomplishing thinning with repeated calls to the driver with `num_steps=1`. To illustrate, consider the toy example:

```
current_state = 0
for _ in range(5):
  new_sample, _ = sample_chain(
      num_steps=1,
      current_state=current_state,
      kernel=<some_kernel>,
      num_burnin_steps=0,
      num_steps_between_results=100)
  current_state = do_something(new_sample)
```

At every iteration, the first sample drawn from the `kernel` is returned from `sample_chain`, though one should expect some thinning. The workaround would be doing everything manually from `kernel.one_step` calls, but that kind of defeats the purpose of having a smart driver in the first place. 

It seems more natural to have semantics as follows:
- the first sample will step the inner kernel `num_burnin_steps` + `num_steps_between_results` + 1 times
- any other samples will step `num_steps_between_results` + 1 times

PR #1036, opened 2020-07-31
Add `sample_fold` driver to `tfp.experimental.mcmc`

rupei:
`sample_fold` is a high-level driver that automatically applies reducers to stream over MCMC samples. It draws samples from the provided `kernel`, updates reducer states with those samples, and finalizes reducer computation. This piece makes it easy to conduct streaming MCMC without worrying about manually creating the Transition Kernel onion.

Thinning/burn-in semantics follow that proposed in https://github.com/tensorflow/probability/issues/1031

### DIFF
--- a/tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
+++ b/tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
@@ -52,6 +52,34 @@ class CovarianceReducer(reducer_base.Reducer):
   `CovarianceReducer` is meant to fit into the larger Streaming MCMC
   framework. `RunningCovariance` in `tfp.experimental.stats` is better suited
   for more generic streaming covariance needs.
+
+  There are two ways to stream over MCMC samples. The first is to manually
+  wrap `CovarianceReducer` in a `tfp.experimental.mcmc.WithReductions`
+  `TransitionKernel`. Doing so enables a wide variety of possible compositions.
+  For example, to perform covariance calculation on samples that survive thinning
+  and burn-in, `WithReductions` should wrap around an appropriate
+  `SampleDiscardingKernel`.
+  
+  If the sole objective is to compute some reduction statistic, it may be more
+  convenient to use a pre-defined driver like `tfp.experimental.mcmc.sample_fold`,
+  which accepts a kernel to sample from and a collection of reducers. `sample_fold`
+  will automatically create the Transition Kernel onion and call each reducer's
+  `finalize` method. This makes it easy to calculate streaming covariance over
+  MCMC samples.
+
+  ```
+  python 
+
+  kernel = ...
+  reducer = tfp.experimental.mcmc.CovarianceReducer()
+  covariance_estimate = tfp.experimental.mcmc.sample_fold(
+      num_steps=...,
+      current_state=...,
+      previous_kernel_results=...,
+      kernel=kernel,
+      reducers=reducer,
+  )
+  ```
   """
 
   def __init__(self, event_ndims=None, ddof=0, name=None):
--- a/tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
+++ b/tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
@@ -70,7 +70,7 @@ class CovarianceReducer(reducer_base.Reducer):
 
   kernel = ...
   reducer = tfp.experimental.mcmc.CovarianceReducer()
-  covariance_estimate = tfp.experimental.mcmc.sample_fold(
+  covariance_estimate, _, _ = tfp.experimental.mcmc.sample_fold(
       num_steps=...,
       current_state=...,
       previous_kernel_results=...,
--- a/tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
+++ b/tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
@@ -56,17 +56,17 @@ class CovarianceReducer(reducer_base.Reducer):
   There are two ways to stream over MCMC samples. The first is to manually
   wrap `CovarianceReducer` in a `tfp.experimental.mcmc.WithReductions`
   `TransitionKernel`. Doing so enables a wide variety of possible compositions.
-  For example, to perform covariance calculation on samples that survive thinning
-  and burn-in, `WithReductions` should wrap around an appropriate
+  For example, to perform covariance calculation on samples that survive
+  thinning and burn-in, `WithReductions` should wrap around an appropriate
   `SampleDiscardingKernel`.
-  
+
   If the sole objective is to estimate covariance, it may be more convenient to
-  use a pre-defined driver like `tfp.experimental.mcmc.sample_fold`. `sample_fold`
-  will automatically create the Transition Kernel onion, apply `kernel` samples
-  to update reducer states, and `finalize` calculations.
+  use a pre-defined driver like `tfp.experimental.mcmc.sample_fold`.
+  `sample_fold` will automatically create the Transition Kernel onion, apply
+  `kernel` samples to update reducer states, and `finalize` calculations.
 
   ```
-  python 
+  python
 
   kernel = ...
   reducer = tfp.experimental.mcmc.CovarianceReducer()
--- a/tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
+++ b/tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
@@ -60,12 +60,10 @@ class CovarianceReducer(reducer_base.Reducer):
   and burn-in, `WithReductions` should wrap around an appropriate
   `SampleDiscardingKernel`.
   
-  If the sole objective is to compute some reduction statistic, it may be more
-  convenient to use a pre-defined driver like `tfp.experimental.mcmc.sample_fold`,
-  which accepts a kernel to sample from and a collection of reducers. `sample_fold`
-  will automatically create the Transition Kernel onion and call each reducer's
-  `finalize` method. This makes it easy to calculate streaming covariance over
-  MCMC samples.
+  If the sole objective is to estimate covariance, it may be more convenient to
+  use a pre-defined driver like `tfp.experimental.mcmc.sample_fold`. `sample_fold`
+  will automatically create the Transition Kernel onion, apply `kernel` samples
+  to update reducer states, and `finalize` calculations.
 
   ```
   python 
@@ -79,6 +77,7 @@ class CovarianceReducer(reducer_base.Reducer):
       kernel=kernel,
       reducers=reducer,
   )
+  covariance_estimate  # estimate of covariance (not `CovarianceReducerState`)
   ```
   """
 
--- a/tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
+++ b/tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
@@ -22,7 +22,7 @@ import collections
 
 # Dependency imports
 import tensorflow.compat.v2 as tf
-from tensorflow_probability.python.experimental.mcmc import step_kernel
+from tensorflow_probability.python.experimental import mcmc
 from tensorflow_probability.python.mcmc import kernel as kernel_base
 from tensorflow_probability.python.mcmc.internal import util as mcmc_util
 
@@ -64,7 +64,7 @@ class SampleDiscardingKernel(kernel_base.TransitionKernel):
 
     Args:
       inner_kernel: `TransitionKernel` whose `one_step` will generate
-        MCMC sample(s).
+        MCMC results.
       num_burnin_steps: Integer or scalar `Tensor` representing the number
         of chain steps to take before starting to collect results.
         Defaults to 0 (i.e., no burn-in).
@@ -75,11 +75,6 @@ class SampleDiscardingKernel(kernel_base.TransitionKernel):
       name: Python `str` name prefixed to Ops created by this function.
         Default value: `None` (i.e., "sample_discarding_kernel").
     """
-    if tf.get_static_value(num_burnin_steps):
-      num_burnin_steps = tf.get_static_value(num_burnin_steps)
-    if tf.get_static_value(num_steps_between_results):
-      num_steps_between_results = tf.get_static_value(
-          num_steps_between_results)
     self._parameters = dict(
         inner_kernel=inner_kernel,
         num_burnin_steps=num_burnin_steps,
@@ -91,34 +86,33 @@ class SampleDiscardingKernel(kernel_base.TransitionKernel):
     """Calculates how many samples to skip based on the call number."""
     # If `self.num_burnin_steps` is statically known to be 0,
     # `self.num_steps_between_results` will be returned outright.
-    if self.num_burnin_steps == 0:
+    num_burnin_steps_ = tf.get_static_value(self.num_burnin_steps)
+    if num_burnin_steps_ == 0:
       return self.num_steps_between_results
     else:
       return (tf.where(tf.equal(call_counter, 0), self.num_burnin_steps, 0) +
               self.num_steps_between_results)
 
   def one_step(self, current_state, previous_kernel_results, seed=None):
-    """Collects one non-discarded sample.
+    """Collects one non-discarded chain state.
 
     Args:
       current_state: `Tensor` or Python `list` of `Tensor`s
         representing the current state(s) of the Markov chain(s),
-      previous_kernel_results: `SampleDiscardingKernelResults` named tuple.
-        `SampleDiscardingKernelResults` contains the `call_counter`
-        and a reference to kernel results of nested `TransitionKernel`s.
+      previous_kernel_results: `collections.namedtuple` containing `Tensor`s
+        representing values from previous calls to this function (or from the
+        `bootstrap_results` function).
       seed: Optional seed for reproducible sampling.
 
     Returns:
       new_chain_state: Newest non-discarded MCMC chain state drawn from
         the `inner_kernel`.
-      kernel_results: `SampleDiscardingKernelResults` representing updated
-        kernel results. The `call_counter` will be incremented by 1 and
-        its `inner_results` will reflect the new kernel results of nested
-        `TransitionKernel`s.
+      kernel_results: `collections.namedtuple` of internal calculations used to
+        advance the chain.
     """
     with tf.name_scope(
         mcmc_util.make_name(self.name, 'sample_discarding_kernel', 'one_step')):
-      new_chain_state, inner_kernel_results = step_kernel(
+      new_chain_state, inner_kernel_results = mcmc.step_kernel(
           num_steps=self._num_samples_to_skip(
               previous_kernel_results.call_counter
           ) + 1,
@@ -141,9 +135,8 @@ class SampleDiscardingKernel(kernel_base.TransitionKernel):
         state(s) of the Markov chain(s).
 
     Returns:
-      kernel_results: `SampleDiscardingKernelResults` state with
-        `call_counter=0` and the result of nested `TransitionKernel`s
-        `bootstrap_results`.
+      kernel_results: `collections.namedtuple` of `Tensor`s representing
+        internal calculations made within this function.
     """
     with tf.name_scope(
         mcmc_util.make_name(
--- a/tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
+++ b/tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
@@ -39,7 +39,20 @@ class SampleDiscardingKernelResults(
                             'inner_results'])):
   __slots__ = ()
 
+
 class SampleDiscardingKernel(kernel_base.TransitionKernel):
+  """Appropriately discards samples to conduct thinning and burn-in.
+
+  `SampleDiscardingKernel` is a composable `TransitionKernel` that
+  applies thinning and burn-in to samples returned by its
+  `inner_kernel`. All Transition Kernels wrapping it will only
+  see non-discarded samples.
+
+  The burn-in step conducts both burn-in and one step of thinning.
+  In other words, the first call to `one_step` will skip
+  `num_burnin_steps + num_steps_between_results` samples. All
+  subsequent calls skip only `num_steps_between_results` samples.
+  """
 
   def __init__(
       self,
@@ -47,6 +60,20 @@ class SampleDiscardingKernel(kernel_base.TransitionKernel):
       num_burnin_steps=0,
       num_steps_between_results=0,
       name=None):
+    """Instantiates this object.
+
+    Args:
+      inner_kernel: `TransitionKernel` whose `one_step` will generate
+        MCMC sample(s).
+      num_burnin_steps: Integer number of chain steps to take before starting to
+        collect results. Defaults to 0 (i.e., no burn-in).
+      num_steps_between_results: Integer number of chain steps between
+        collecting a result. Only one out of every
+        `num_steps_between_samples + 1` steps is included in the returned
+        results. Defaults to 0 (i.e., no thinning).
+      name: Python `str` name prefixed to Ops created by this function.
+        Default value: `None` (i.e., "sample_discarding_kernel").
+    """
     if tf.get_static_value(num_burnin_steps):
       num_burnin_steps = tf.get_static_value(num_burnin_steps)
     if tf.get_static_value(num_steps_between_results):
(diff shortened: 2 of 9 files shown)


tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
```diff
@@ -0,0 +1,178 @@
...
+      num_steps_between_results=0,
+      name=None):
+    """Instantiates this object.
+
+    Args:
+      inner_kernel: `TransitionKernel` whose `one_step` will generate
+        MCMC sample(s).
+      num_burnin_steps: Integer number of chain steps to take before starting to
+        collect results. Defaults to 0 (i.e., no burn-in).
+      num_steps_between_results: Integer number of chain steps between
```
axch, line 70: Both of these should be able to be Tensors also, which corresponds to statically unknown numbers of thinning and burn-in steps.  It's possible that the code already does that, in which case the only thing that needs to be updated is documentation and/or tests.

tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
```diff
@@ -0,0 +1,178 @@
...
+        num_burnin_steps=num_burnin_steps,
+        num_steps_between_results=num_steps_between_results,
+        name=name or 'sample_discarding_kernel'
+    )
+
+  def _num_samples_to_skip(self, call_counter):
+    """Calculates how many samples to skip based on the call number.
+
+    If `self.num_burnin_steps` is statically known to be 0,
+    `self.num_steps_between_results` will be returned outright.
```
axch, line 93: Google docstring style: If the docstring is longer than one line, then an Args: and a Returns: section are required.

In this case, since the method is private, you could choose to shorten the docstring to just the summary, and make the rest a code comment.

tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel_test.py
```diff
@@ -0,0 +1,286 @@
...
+        0., discarder.bootstrap_results(0.))
+    second_state, kernel_results = discarder.one_step(
+        first_state, kernel_results)
+    first_state, second_state, kernel_results = self.evaluate([
+        first_state, second_state, kernel_results])
+    self.assertEqual(12, first_state)
+    self.assertEqual(14, second_state)
+    self.assertEqual(2, kernel_results.call_counter)
+    self.assertEqual(14, kernel_results.inner_results.counter_1)
+    self.assertEqual(28, kernel_results.inner_results.counter_2)
```
axch, line 127: Please add a quick comment explaining these magic numbers.  Also in at least the next test.

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -0,0 +1,118 @@
...
+    name=None,
+):
+  """Finalizes `num_steps` Reducer steps with `kernel` samples.
+
+  This driver will appropriately apply `WithReductions` and
+  `SampleDiscardingKernel` over the given `kernel` and `reducers`. It will
+  then update reducer states `num_steps` times (perform `num_steps` calls to
+  the `Reducer`'s `one_step` method). An arbitrary collection of `reducers`
+  can be provided, and the resulting finalized statistic(s) will be returned
+  in an identical structure.
```
axch, line 53:
```suggestion
  """Computes the requested reductions over the `kernel`'s samples.

  To wit, runs the given `kernel` for `num_steps` steps, and consumes
  the stream of samples with the given `Reducer`s' `one_step` method(s).
  This runs in constant memory (unless a given `Reducer` builds a
  large structure).
  
  The driver internally composes the correct onion of `WithReductions`
  and `SampleDiscardingKernel` to implement the requested optionally
  thinned reduction.
  
  An arbitrary collection of `reducers` can be provided, and the resulting
  finalized statistic(s) will be returned in an identical structure.
```

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -0,0 +1,118 @@
...
+
+  Args:
+    num_steps: Integer number of `Reducer` steps.
```
axch, line 56: Or Tensor?

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -0,0 +1,118 @@
...
+      current state(s) of the Markov chain(s).
+    previous_kernel_results: A `Tensor` or a nested collection of `Tensor`s.
+      Warm-start for the auxiliary state needed by the given `kernel`.
+      If not supplied, `sample_fold` will cold-start with
+      `kernel.bootstrap_results`.
+    kernel: An instance of `tfp.mcmc.TransitionKernel` which implements one step
+      of the Markov chain.
+    reducers: A (possibly nested) structure of `Reducer`s to be evaluated
+      on the `kernel`'s samples. If no reducers are given (`reducers=None`), an
+      empty list will be returned in place of streaming calculations.
```
axch, line 67: It would be more stylish to return `None` rather than an empty list, since that more closely mirrors the input structure.

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -0,0 +1,118 @@
...
+    num_burnin_steps: Integer number of chain steps to take before starting to
+      collect results. Defaults to 0 (i.e., no burn-in).
+    num_steps_between_results: Integer number of chain steps between collecting
```
axch, line 70: Presumably both of these can be Tensor also.

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -0,0 +1,118 @@
...
+      included in the returned results. Defaults to 0 (i.e., no thinning).
+    parallel_iterations: The number of iterations allowed to run in parallel. It
+      must be a positive integer. See `tf.while_loop` for more details.
+    seed: Optional seed for reproducible sampling.
+    name: Python `str` name prefixed to Ops created by this function.
+      Default value: `None` (i.e., 'mcmc_sample_fold').
+
+  Returns:
+    reduction_results: A (possibly nested) structure of finalized reducer
+      statistics. The structure identically mimics that of `reducers`
```
axch, line 81:
```suggestion
      statistics. The structure identically mimics that of `reducers`.
```

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -0,0 +1,118 @@
...
+    warm_restart_package: `Tuple` with the final state of the Markov chain(s)
+      and kernel results, including those of applied wrapper kernels (i.e.
```
axch, line 83:
```suggestion
      and kernel results, including those of applied wrapper kernels (i.e.,
```

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -0,0 +1,118 @@
...
+      `WithReductions`).
+  """
+  with tf.name_scope(name or 'mcmc_sample_fold'):
+    num_steps = tf.convert_to_tensor(
+        num_steps, dtype=tf.int32, name='num_steps')
+    current_state = tf.nest.map_structure(
+        lambda x: tf.convert_to_tensor(x, name='current_state'),
+        current_state)
+    if reducers is None:
+      reducers = []
```
axch, line 93: Is this actually needed?  Does WithReductions fail to do the right thing if the reducers argument is `None`?

tensorflow_probability/python/experimental/mcmc/sample_fold_test.py
```diff
@@ -0,0 +1,301 @@
...
+    return TestTransitionKernelResults(
+        counter_1=tf.zeros(self._shape),
+        counter_2=tf.zeros(self._shape))
+
+  @property
+  def is_calibrated(self):
+    return self._is_calibrated
+
+
+class RandomTransitionKernel(tfp.mcmc.TransitionKernel):
```
axch, line 61: Do you think it's time to make a separate module collecting these identical, replicated test `TransitionKernel`s?  It's a tradeoff between explicitness and code de-duplication.

tensorflow_probability/python/experimental/mcmc/sample_fold_test.py
```diff
@@ -0,0 +1,301 @@
...
+    fake_reducer = TestReducer()
+    _, warm_restart_pkg = tfp.experimental.mcmc.sample_fold(
+        num_steps=5,
+        current_state=0.,
+        kernel=fake_kernel,
+        reducers=fake_reducer,)
+    # verify `warm_restart_pkg` works as intended
+    reduction_rslt, warm_restart_pkg = tfp.experimental.mcmc.sample_fold(
+        num_steps=5,
+        current_state=warm_restart_pkg[0],
```
axch, line 153: The fact that you needed to do this [0] suggests to me that a better API would be for `sample_fold` to return three things: the reduction result, the final chain state, and the final kernel results.  That way the pieces of the return value map more cleanly to the arguments that must be supplied for the next call.

tensorflow_probability/python/experimental/mcmc/sample_fold_test.py
rupei: Yeah, I've been thinking the same thought more and more as I've been using these objects more and more. Would grouping them with other test_utils in a separate PR be appropriate?

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -0,0 +1,118 @@
...
+# ============================================================================
+"""Drivers for streaming reductions framework."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+# Dependency imports
+import tensorflow.compat.v2 as tf
+from tensorflow_probability.python.experimental.mcmc import with_reductions
```
rupei, line 23: also a curious question here. No matter what I try it seems I can't import `WithReductions` (i.e. `from tensorflow_probability.python.experimental.mcmc import WithReductions`)

tensorflow_probability/python/experimental/mcmc/sample_fold.py
rupei: It was since `WithReductions` didn't have a None check (oops). I will add that to the next commit though

rupei: WithReductions isn't currently implemented to take in a `None` reducer (it'll break). After all, with no reducer, why even use WithReductions?

One could have WithReductions return None as the streaming calculations field if the reducers are None, but then `sample_fold`'s call to `step_kernel` invokes a `tf.while` via `smart_for_loop` that's not too happy with a None value in the kernel results (can't be converted to a `Tensor`). If we really wanted to return `None`, we could make reducers an empty list and then just do some post-processing to make everything relevant being returned into `None`, but then the kernel results also aren't compatible with a `tf.while_loop` anymore (probably can't go through future drivers for warm restart), which seems problematic.

I've gone ahead and made this fix anyways as in situations like these in the past, I've always overlooked something haha

tensorflow_probability/python/experimental/mcmc/sample_fold_test.py
axch: I'm ambivalent about whether to call them test_utils or test_fixtures, since they are sort of the latter rather than the former.  Separate PR is reasonable.

tensorflow_probability/python/experimental/mcmc/sample_fold.py
axch: Don't.  Import mcmc and refer to the symbol as `mcmc.WithReductions`.

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -114,5 +123,9 @@ def sample_fold(
         reducers,
         final_kernel_results.streaming_calculations,
         check_types=False)
-    warm_restart_package = (end_state, final_kernel_results)
-    return reduction_results, warm_restart_package
+    if reduction_results == []:
+      reduction_results = None
```
axch, line 126: So, this doesn't actually work either, because if the user had passed `reducers=[]` explicitly, they will get `None` reduction results, and should get `[]`.  That's why input canonicalization mixes poorly with maintaining the user's choice of data structures: canonicalization loses information that must then be re-constructed.

In this case, you have argued that the underlying tf.while_loop can't handle a `None` reducers structure.  If that is the case, then something has to deal with it, like this:
```
if reducers is None:
  reducers_was_none = True
  reducers = []
...
if reducers_was_none:
  reducer_parallel_results = None
```

If it doesn't make sense to put this into `WithReductions`, then it unfortunately has to go here.

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -20,9 +20,8 @@
 
 # Dependency imports
 import tensorflow.compat.v2 as tf
-from tensorflow_probability.python.experimental.mcmc import with_reductions
-from tensorflow_probability.python.experimental.mcmc import SampleDiscardingKernel
-from tensorflow_probability.python.experimental.mcmc import step_kernel
+from tensorflow_probability.python.experimental import mcmc
+from tensorflow_probability.python.experimental.mcmc.with_reductions import WithReductionsKernelResults
```
axch, line 24: Why not just add WithReductionsKernelResults to the experimental.mcmc `__init__` file?

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -52,6 +52,33 @@ class CovarianceReducer(reducer_base.Reducer):
...
+  use a pre-defined driver like `tfp.experimental.mcmc.sample_fold`.
+  `sample_fold` will automatically create the Transition Kernel onion, apply
+  `kernel` samples to update reducer states, and `finalize` calculations.
+
+  ```
+  python
+
+  kernel = ...
+  reducer = tfp.experimental.mcmc.CovarianceReducer()
+  covariance_estimate = tfp.experimental.mcmc.sample_fold(
```
SiegeLordEx, line 73: This isn't right, sample_fold returns 3 outputs.

tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
```diff
@@ -0,0 +1,177 @@
...
+      self,
+      inner_kernel,
+      num_burnin_steps=0,
+      num_steps_between_results=0,
+      name=None):
+    """Instantiates this object.
+
+    Args:
+      inner_kernel: `TransitionKernel` whose `one_step` will generate
+        MCMC sample(s).
```
SiegeLordEx, line 67: Instead of samples, maybe call these 'results' to match the other args?

tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
```diff
@@ -0,0 +1,177 @@
...
+        of chain steps to take before starting to collect results.
+        Defaults to 0 (i.e., no burn-in).
+      num_steps_between_results: Integer or scalar `Tensor` representing
+        the number of chain steps between collecting a result. Only one out
+        of every `num_steps_between_samples + 1` steps is included in the
+        returned results. Defaults to 0 (i.e., no thinning).
+      name: Python `str` name prefixed to Ops created by this function.
+        Default value: `None` (i.e., "sample_discarding_kernel").
+    """
+    if tf.get_static_value(num_burnin_steps):
```
SiegeLordEx, line 78: I feel like it'd be clearer to do this inside `_num_samples_to_skip`. `parameters` really should contain the original values, if possible, and it's also more self-explanatory to do it there.

tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
```diff
@@ -0,0 +1,177 @@
...
+        num_burnin_steps=num_burnin_steps,
+        num_steps_between_results=num_steps_between_results,
+        name=name or 'sample_discarding_kernel'
+    )
+
+  def _num_samples_to_skip(self, call_counter):
+    """Calculates how many samples to skip based on the call number."""
+    # If `self.num_burnin_steps` is statically known to be 0,
+    # `self.num_steps_between_results` will be returned outright.
+    if self.num_burnin_steps == 0:
```
SiegeLordEx, line 94: I don't see how this can work if self.num_burnin_steps is not statically known, I think this code is broken. I'd have expected this code to look like:

```python
num_burnin_steps_ = tf.get_static_value(self.num_burnin_steps)
if num_burnin_steps_ == 0:
  return self.num_steps_between_results
else:
  return (tf.where(tf.equal(call_counter, 0), self.num_burnin_steps, 0) +
              self.num_steps_between_results)
...
```

tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
```diff
@@ -0,0 +1,177 @@
...
+              self.num_steps_between_results)
+
+  def one_step(self, current_state, previous_kernel_results, seed=None):
+    """Collects one non-discarded sample.
+
+    Args:
+      current_state: `Tensor` or Python `list` of `Tensor`s
+        representing the current state(s) of the Markov chain(s),
+      previous_kernel_results: `SampleDiscardingKernelResults` named tuple.
+        `SampleDiscardingKernelResults` contains the `call_counter`
```
SiegeLordEx, line 107: It's not necessary to inline the definition of `SampleDiscardingKernelResults`, it only increases maintenance burden.

tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
```diff
@@ -0,0 +1,177 @@
...
+    # If `self.num_burnin_steps` is statically known to be 0,
+    # `self.num_steps_between_results` will be returned outright.
+    if self.num_burnin_steps == 0:
+      return self.num_steps_between_results
+    else:
+      return (tf.where(tf.equal(call_counter, 0), self.num_burnin_steps, 0) +
+              self.num_steps_between_results)
+
+  def one_step(self, current_state, previous_kernel_results, seed=None):
+    """Collects one non-discarded sample.
```
SiegeLordEx, line 101: s/sample/chain_state/

tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
```diff
@@ -0,0 +1,177 @@
...
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import collections
+
+# Dependency imports
+import tensorflow.compat.v2 as tf
+from tensorflow_probability.python.experimental.mcmc import step_kernel
```
SiegeLordEx, line 25: Does doing `from tensorflow_probability.python.experimental import mcmc; ... mcmc.step_kernel` work? If so, that's the style preferred by the style guide.

tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel_test.py
```diff
@@ -0,0 +1,294 @@
...
+    self.assertEqual(14, final_sample)
+    self.assertEqual(2, kernel_results.call_counter)
+    self.assertEqual(14, kernel_results.inner_results.counter_1)
+    self.assertEqual(28, kernel_results.inner_results.counter_2)
+
+  def test_tensor_thinning_and_burnin(self):
+    fake_inner_kernel = TestTransitionKernel()
+    discarder = tfp.experimental.mcmc.SampleDiscardingKernel(
+        fake_inner_kernel,
+        num_burnin_steps=tf.convert_to_tensor(10),
```
SiegeLordEx, line 222: If this was meant to test the non-statically known pathway, it doesn't as these are trivially statically known. I recommend replacing them with tf.Variable.

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -0,0 +1,131 @@
...
+    'sample_fold',
+]
+
+
+def sample_fold(
+    num_steps,
+    current_state,
+    previous_kernel_results=None,
+    kernel=None,
+    reducers=None,
```
SiegeLordEx, line 37: For nests, I'm a fan of naming the arguments in the singular (i.e.. `reducer`). Not only is that likely the common situation, it also matches what we do for nests in the rest of TFP (e.g. JointDistribution.event_shape rather than JointDistribution.event_shape**s**).

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -0,0 +1,131 @@
...
+    reduction_results = nest.map_structure_up_to(
+        reducers,
+        lambda r, s: r.finalize(s),
+        reducers,
+        final_kernel_results.streaming_calculations,
+        check_types=False)
+    if reduction_results == [] and reducers_was_none:
+      reduction_results = None
+      final_kernel_results = mcmc.WithReductionsKernelResults(
+          None, final_kernel_results.inner_results
```
SiegeLordEx, line 129: It think we should return the same type as `previous_kernel_results`. Any reason not to?

tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
```diff
@@ -0,0 +1,177 @@
...
+      num_steps_between_results: Integer or scalar `Tensor` representing
+        the number of chain steps between collecting a result. Only one out
+        of every `num_steps_between_samples + 1` steps is included in the
+        returned results. Defaults to 0 (i.e., no thinning).
+      name: Python `str` name prefixed to Ops created by this function.
+        Default value: `None` (i.e., "sample_discarding_kernel").
+    """
+    if tf.get_static_value(num_burnin_steps):
+      num_burnin_steps = tf.get_static_value(num_burnin_steps)
+    if tf.get_static_value(num_steps_between_results):
```
SiegeLordEx, line 80: This doesn't seem necessary.

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
rupei: Ah good catch! This was a fix we made in the last round of reviews and I forgot to change this

tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
rupei: Yup! After doing the `get_static_value` here and not in the `init` it does look something like that. For learning purposes though, why would it have not been functional if the `get_static_value` was done in the init?

tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
rupei: wasn't exactly sure if I interpreted what this referred to right; please follow up if I indeed didn't!

tensorflow_probability/python/experimental/mcmc/sample_fold.py
rupei: This way, the user can still introspect the kernel results of `WithReductions` and `SampleDiscardingKernel`. The main benefit is that they have a way to re-reference the final reducer states (in `WithReductions`' kernel results). It may be somewhat unintuitive though, but assuming the user has also read the documentation they could also get to the same layer as their given pkr by taking `final_kernel_results.inner_results.inner_results`

tensorflow_probability/python/experimental/mcmc/sample_discarding_kernel.py
SiegeLordEx: There's nothing wrong with doing that per se, what raised a red flag for me is storing the output of that in `parameters`, which was bad for two reaons:

1. `get_static_value` will return None sometimes, so you were losing information.
2. `parameters` elsewhere in TFP stores the original arguments to `__init__` without any modification. In fact, in distributions we explicitly do `parameters = dict(locals())` as the first line of `__init__` to accomplish that automatically. I don't know why that isn't the style in `tfp.mcmc`, but I think we should follow the spirit of that code.

tensorflow_probability/python/experimental/mcmc/sample_fold.py
SiegeLordEx: Hmm, I see the conundrum. Here are my concerns:

1. I kind of dislike the fact that we have `final_kernel_results` and `previous_kernel_results` in the same code block, but with different meanings of `kernel`.
2. Regardless of the name, this will be the first driver which is non-markovian in the sense of thestate type changing; this seems an unfortunate departure from precedent.
3. Allowing people to introspect is nice, but this driver does not let you warmstart the reduction with the reducer state, which is the primary use case I can think of for doing this.

There are a number of things we can try doing to resolve this:

1. Make the driver accept both inner kernel results and outer kernel results. This handles many common use cases, but is ambiguous sometimes.
2. Force the driver to only accept WithReductionKernelResults as input. I think that partially defeats the purpose of this driver, as it was meant to hide that state.
3. Make the driver return the inner kernel results only. This precludes introspection and warmstarting the reducer state, even in principle.
4. Do nothing.

I argue that (3) is the best option, as we can consider this driver to merely be syntax sugar for a common use case, and for people who want to do the introspection you're talking about + warmstarting the reduction state, they should use step_kernel + WithReductions manually (presumably this is the less common use case).

@axch, what do you think?

axch: On reflection, I think I agree with (3).  And document that if warm-restart of the reductions is desired, then step_kernel with an explicit WithReductions kernel is the way to go.

(Or, amusingly, sample_fold with no reducers and an explicit WithReductions kernel will also gain the same effect!).

Reasoning: The fundamental concept is step_kernel, and composability through nesting kernels to step.  Sample_fold is a guess as to what people will consider convenient.

PR #1042, opened 2020-08-05
Make Replica Exchange / simple step size adaption work together

### LINKED ISSUE #1032: SimpleStepSizeAdaption kernel not working when used within ReplicaExchangeMC kernel
It seems currently not possible to use the `SimpleStepSizeAdaption` kernel within a `ReplicaExchangeMC` kernel, because the former's `SimpleStepSizeAdaptionResults` is missing the `target_log_prob` field required by `ReplicaExchangeMC`. See a minimal-ish failing example in [this Colab notebook](https://colab.research.google.com/drive/1nRQqwbpYwZSUin0IIB5XcidyXMCH6A3U?usp=sharing).

Naively, it seems to me like it's an easy thing to fix by just adding the field to the results named tuple and propagating the `target_log_prob` from the inner kernel when returning the results `namedtuple` at the end of `one_step()` and in `bootstrap_results()`. If that's indeed the case, I'd be happy to fix this myself!

simeoncarstens:
Fixes #1032.
~~The changes in this PR are two-fold, but closely related (let me know if this merits two separate PRs):~~

1. ~~for the `SimpleStepSizeAdaption`, the `target_log_prob` field of the inner kernel results gets propagated to the `SimpleStepSizeAdaption` kernel results `namedtuple`~~
2. ~~while I'm not sure whether the `target_log_prob` field in results `namedtuple`s of existing kernels is used anywhere else other than in `ReplicaExchangeMC`, the change in 1.) allows, in principle, to embed the `SimpleStepSizeAdaption` kernel in the `ReplicaExchangeMC` kernel. For that to work, some further changes in the `ReplicaExchangeMC` code were necessary to propagate states to and from `SimpleStepSizeAdaption`'s inner kernel results~~

~~Currently, in `ReplicaExchangeMC`, there are distinctions via `if... else...` for two different kernels (HMC and RWMH) and kernel results. This PR just piles on with the necessary distinctions for the `SimpleStepSizeAdaption` kernel and HMC and RWMH as possible inner kernels. It gets somewhat unwieldy, but (and please correct me if I'm wrong) I feel that to improve that, there needs to be a better-defined interface of `tfp.mcmc.TransitionKernel` for recursive propagation of states and possibly other information.~~

This PR now implements @axch's [suggestion from the comments](https://github.com/tensorflow/probability/pull/1042#pullrequestreview-461931484). `ReplicaExchangeMC` _should_ now work with any nested kernel (as long as its kernel results `namedtuple` uses a `inner_results` / `accepted_results` field. At least it does with `SimpleStepSizeAdaptation`, which was the original issue.

I'm somewhat unsure about the tests. For the changes to `ReplicaExchangeMC`, I added a `SimpleStepSizeAdaptation` kernel parameter to the two existing functional tests of a 1- and 2D mixture of normal distributions, but did not test anything further apart from the `get_field()` / `update_field()` helper functions I moved to `tfp.mcmc.python.mcmc.internal.util` / wrote.
In case this is not sufficient, I'd appreciate some input on what additional tests are required and would be happy to implement them.

### DIFF
--- a/tensorflow_probability/python/mcmc/internal/util.py
+++ b/tensorflow_probability/python/mcmc/internal/util.py
@@ -741,11 +741,7 @@ def get_field(kernel_results, field_name):
   if accepted_results is None:
     raise TypeError('Cannot extract {} from {}'.format(
         field_name, kernel_results))
-  attr = getattr(accepted_results, field_name, None)
-  if attr is None:
-    raise TypeError('Cannot extract {} from {}'.format(
-        field_name, kernel_results))
-  return attr
+  return get_field(accepted_results, field_name)
 
 
 @make_innermost_setter
@@ -757,8 +753,5 @@ def update_field(kernel_results, field_name, value):
   if accepted_results is None:
     raise TypeError('Cannot set {} in {}'.format(
         field_name, kernel_results))
-  if not hasattr(accepted_results, field_name):
-    raise TypeError('Cannot set {} in {}'.format(
-        field_name, accepted_results))
   return kernel_results._replace(
-      accepted_results=accepted_results._replace(**{field_name: value}))
+      accepted_results=update_field(accepted_results, field_name, value))
--- a/tensorflow_probability/python/mcmc/internal/util.py
+++ b/tensorflow_probability/python/mcmc/internal/util.py
@@ -728,3 +728,37 @@ def index_remapping_gather(params,
     return dist_util.move_dimension(result_t,
                                     source_idx=broadcast_indices_ndims - 1,
                                     dest_idx=axis)
+
+
+# TODO(b/111801087): Use a standardized API, when available.
+@make_innermost_getter
+def get_field(kernel_results, field_name):
+  """Get field value from kernel_results or kernel_results.accepted_results."""
+  attr = getattr(kernel_results, field_name, None)
+  if attr is not None:
+    return attr
+  accepted_results = getattr(kernel_results, 'accepted_results', None)
+  if accepted_results is None:
+    raise TypeError('Cannot extract {} from {}'.format(
+        field_name, kernel_results))
+  attr = getattr(accepted_results, field_name, None)
+  if attr is None:
+    raise TypeError('Cannot extract {} from {}'.format(
+        field_name, kernel_results))
+  return attr
+
+
+@make_innermost_setter
+def update_field(kernel_results, field_name, value):
+  """Set field value in kernel_results or kernel_results.accepted_results."""
+  if hasattr(kernel_results, field_name):
+    return kernel_results._replace(**{field_name: value})
+  accepted_results = getattr(kernel_results, 'accepted_results', None)
+  if accepted_results is None:
+    raise TypeError('Cannot set {} in {}'.format(
+        field_name, kernel_results))
+  if not hasattr(accepted_results, field_name):
+    raise TypeError('Cannot set {} in {}'.format(
+        field_name, accepted_results))
+  return kernel_results._replace(
+      accepted_results=accepted_results._replace(**{field_name: value}))
--- a/tensorflow_probability/python/mcmc/internal/util_test.py
+++ b/tensorflow_probability/python/mcmc/internal/util_test.py
@@ -584,24 +584,24 @@ FakeKernelResults = collections.namedtuple('FakeKernelResults', 'some_field')
 
 
 FakeKernelAcceptedResults = collections.namedtuple(
-  'FakeKernelAcceptedResults',
-  'accepted_results')
+    'FakeKernelAcceptedResults',
+    'accepted_results')
 
 
 FakeAcceptedResults = collections.namedtuple(
-  'FakeAcceptedResults', 'some_field')
+    'FakeAcceptedResults', 'some_field')
 
 
 FakeIncompleteKernelResults = collections.namedtuple(
-  'FakeIncompleteKernelResults', 'some_field')
+    'FakeIncompleteKernelResults', 'some_field')
 
 
 class GetFieldTest(test_util.TestCase):
   @parameterized.parameters(
-    [FakeKernelResults(some_field='yak')],
-    [FakeKernelAcceptedResults(
-      accepted_results=FakeAcceptedResults(some_field='yak'))]
-    )
+      [FakeKernelResults(some_field='yak')],
+      [FakeKernelAcceptedResults(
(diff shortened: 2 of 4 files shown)


simeoncarstens:
Oh, I see. I indeed misunderstood your comment in #1032&mdash;I thought that applied only to the required changes in `SimpleStepSizeAdaption`.
But now I understand; thanks for the detailed explanation! I think I was thinking about something similar when writing the comment on propagation of information in the PR description.
Your road map really makes sense and would definitely be much, much better than what I proposed. I'll give that a try then and post an update in case I hit a road block.

axch:
Great!  Looking forward to seeing what you come up with!

simeoncarstens:
I made the changes accordingly, which turned out to be quite easy following your list. Thanks again!
The recursive `get_field()` / `update_field()` methods actually actually remove the the kernel(_results) checks and thus make the REMC code a bit nicer to read.
The case of `grads_target_log_prob` not existing in the kernel results onion is simply handled with a `try... except...` statement, which catches the error being thrown when a field cannot be extracted.
While this looks to be only the minimal set of changes you proposed, it makes `ReplicaExchangeMC` work nicely with `SimpleStepSizeAdaption`. I'm not sure what other nested kernels might be expected to work with `ReplicaExchangeMC`, so I'm not sure how far close this is to
> REMC...be[ing] stackable on top of any onion of kernels
In any case, let me know what you think and how this could be further improved!

CI seems to fail with a timeout&mdash; could be the additional kernel tested in the `ReplicaExchangeMC` tests :thinking:

tensorflow_probability/python/mcmc/internal/util.py
```diff
@@ -728,3 +728,37 @@ def index_remapping_gather(params,
...
+  if attr is not None:
+    return attr
+  accepted_results = getattr(kernel_results, 'accepted_results', None)
+  if accepted_results is None:
+    raise TypeError('Cannot extract {} from {}'.format(
+        field_name, kernel_results))
+  attr = getattr(accepted_results, field_name, None)
+  if attr is None:
+    raise TypeError('Cannot extract {} from {}'.format(
+        field_name, kernel_results))
```
axch, line 747: This should probably be generalized to `get_field(accepted_results, field_name)`, to support MH(Something(HMC)).  We don't currently have any of the latter, but it shouldn't hurt.  (Unless I'm missing something and that doesn't actually work?)

tensorflow_probability/python/mcmc/internal/util.py
```diff
@@ -728,3 +728,37 @@ def index_remapping_gather(params,
...
+    return kernel_results._replace(**{field_name: value})
+  accepted_results = getattr(kernel_results, 'accepted_results', None)
+  if accepted_results is None:
+    raise TypeError('Cannot set {} in {}'.format(
+        field_name, kernel_results))
+  if not hasattr(accepted_results, field_name):
+    raise TypeError('Cannot set {} in {}'.format(
+        field_name, accepted_results))
+  return kernel_results._replace(
+      accepted_results=accepted_results._replace(**{field_name: value}))
```
axch, line 764: Similar recursion here.  I think `return kernel_results._replace(accepted_results=update_field(accepted_results, field_name, value))` ought to work.

tensorflow_probability/python/mcmc/internal/util_test.py
```diff
@@ -580,5 +580,61 @@ def testNoWarn(self, tensor_callable):
...
+    'FakeKernelAcceptedResults',
+    'accepted_results')
+
+
+FakeAcceptedResults = collections.namedtuple(
+    'FakeAcceptedResults', 'some_field')
+
+
+FakeIncompleteKernelResults = collections.namedtuple(
+    'FakeIncompleteKernelResults', 'some_field')
```
axch, line 596: Do you need these two?  They look the same as `FakeKernelResults` to me?

tensorflow_probability/python/mcmc/replica_exchange_mc.py
```diff
@@ -969,46 +962,19 @@ def _swap_then_retemper(x):
...
-        'request with the TensorFlow Probability team.')
+  kr = _update_field(kr, 'target_log_prob', _swap_then_retemper(
+      _get_field(kr, 'target_log_prob')))
+  try:
+    new_grads_target_log_prob = _swap_then_retemper(
+        _get_field(kr, 'grads_target_log_prob'))
+    kr = _update_field(
+        kr, 'grads_target_log_prob', new_grads_target_log_prob)
+  except NotImplementedError:
+    pass
```
axch, line 973: This merits at least a comment saying where you expect that `NotImplementedError` to come from.

tensorflow_probability/python/mcmc/replica_exchange_mc_test.py
```diff
@@ -804,8 +825,8 @@ def target_log_prob(x, y):
       # z = matmul(inv(chol(true_cov)), [x, y] - true_mean)
       xy = tf.stack([x, y], axis=-1) - true_mean
       z = linop.solvevec(xy)
-      return -0.5 * tf.reduce_sum(z**2., axis=-1)
 
+      return -0.5 * tf.reduce_sum(z**2., axis=-1)
```
axch, line 829: Typo here?  There should be a blank between the end of the previous function and the start of the next one, anyway.

tensorflow_probability/python/mcmc/replica_exchange_mc.py
simeoncarstens: I added a detailed comment. But
> This merits **at least** a comment .... (emphasis mine)

sounds like this might still not be optimal?

simeoncarstens:
Thanks for the review and the comments with all of which I fully agree and which are being addressed in the most recent commits.

As for the test time: I tested it locally and turns out that the additional tests in `replica_exchange_mc_test.py` indeed increase the time the tests need to run by a factor of ~1.4. The reason for this is most certainly the fact that in `test_normal()`, the additional test adds to the elements of a the second-innermost loop in the nested loop building up the parameters list. That means that the number of parameters that test function is run with increases drastically. I have no idea what to do about that&mdash;I do think that the tests I added there are necessary, because they concern a common use case, but cutting back on other tests surely isn't right either.

tensorflow_probability/python/mcmc/replica_exchange_mc.py
axch: Well, there's optimal, and there's worth the time to perfect.  This looks good to me as it is.

PR #1046, opened 2020-08-06
Adding an Expectations Reducer to tfp.experimental.mcmc

rupei:
Bulking the suite of reducers with one that computes arbitrary expectations over a series of callable functions (that accept the incoming sample and kernel results).

### DIFF
--- a/tensorflow_probability/python/experimental/mcmc/expectations_reducer.py
+++ b/tensorflow_probability/python/experimental/mcmc/expectations_reducer.py
@@ -24,10 +24,8 @@ import collections
 import tensorflow.compat.v2 as tf
 
 from tensorflow_probability.python.experimental.mcmc import reducer as reducer_base
-from tensorflow_probability.python.internal import nest_util
-from tensorflow_probability.python.internal import prefer_static as ps
+from tensorflow_probability.python.experimental.stats import sample_stats
 from tensorflow_probability.python.mcmc.internal import util as mcmc_util
-from tensorflow.python.util import nest  # pylint: disable=g-direct-tensorflow-import
 
 
 __all__ = [
@@ -35,17 +33,21 @@ __all__ = [
 ]
 
 
+def _get_sample(sample_and_kr):
+  return sample_and_kr[0]
+
+
 ExpectationsReducerState = collections.namedtuple(
-    'ExpectationsReducerState', 'num_samples, running_total')
+    'ExpectationsReducerState', 'init_state, expectation_state')
 
 
 class ExpectationsReducer(reducer_base.Reducer):
   """`Reducer` that computes a running expectation.
 
   `ExpectationsReducer` calculates expectation over some arbitrary structure
-  of `callables`. A callable is a function that accepts a Markov chain sample
-  and kernel results, and outputs the relevant value for expectation
-  calculation. In other words, if we denote a callable as f(x,y),
+  of `callables`. A callable is a function that accepts a tuple of a Markov
+  chain sample and kernel results, and outputs the relevant value for
+  expectation calculation. In other words, if we denote a callable as f(x,y),
   `ExpectationsReducer` computes E[f(x,y)] for all provided functions. The
   finalized expectation will also identically mimic the structure of
   `callables`.
@@ -56,7 +58,7 @@ class ExpectationsReducer(reducer_base.Reducer):
   calls.
   """
 
-  def __init__(self, callables=lambda sample, kr: sample, name=None):
+  def __init__(self, callables=_get_sample, name=None):
     """Instantiates this object.
 
     Args:
@@ -89,14 +91,12 @@ class ExpectationsReducer(reducer_base.Reducer):
       initial_chain_state = tf.nest.map_structure(
           tf.convert_to_tensor,
           initial_chain_state)
-      return ExpectationsReducerState(
-          num_samples=tf.zeros((), dtype=tf.int32),
-          running_total=tf.nest.map_structure(
-              lambda _: tf.zeros(
-                  initial_chain_state.shape, initial_chain_state.dtype),
-              self.callables
-          )
+      stream = sample_stats.RunningExpectations(
+          shape=initial_chain_state.shape,
+          callables=self.callables,
+          dtype=initial_chain_state.dtype
       )
+      return ExpectationsReducerState(initial_chain_state, stream.initialize())
 
   def one_step(
       self,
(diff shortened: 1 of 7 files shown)


tensorflow_probability/python/experimental/mcmc/expectations_reducer.py
```diff
@@ -0,0 +1,227 @@
...
+  finalized expectation will also identically mimic the structure of
+  `callables`.
+
+  As with all reducers, ExpectationsReducer does not hold state information;
+  rather, it stores supplied metadata. Intermediate calculations are held in
+  a state object, which is returned via `initialize` and `one_step` method
+  calls.
+  """
+
+  def __init__(self, callables=lambda sample, kr: sample, name=None):
```
SiegeLordEx, line 59: lambdas as default arguments don't render well in docstrings in, e.g., Colab notebooks. A style we've adopted elsewhere is to move this out to a freestanding function with a descriptive name, e.g.:

```python
def _get_sample(sample, kr):
  return sample
```

You can see the difference if you try this in IPython:

```python
def foo():
    return 0
    
def baz(a=lambda: 0):
    return a()
    
def bar(a=foo):
    return a()
    
help(bar) # Arguably this has more information
help(baz)
```

tensorflow_probability/python/experimental/mcmc/expectations_reducer.py
```diff
@@ -0,0 +1,227 @@
...
+    Returns:
+      state: `ExpectationsReducerState` representing a stream of no inputs.
+    """
+    with tf.name_scope(
+        mcmc_util.make_name(self.name, 'expectations_reducer', 'initialize')):
+      initial_chain_state = tf.nest.map_structure(
+          tf.convert_to_tensor,
+          initial_chain_state)
+      return ExpectationsReducerState(
+          num_samples=tf.zeros((), dtype=tf.int32),
```
SiegeLordEx, line 93: Hmm, why not add a RunningMean to sample_stats.py and use that here? RunningMean is generally useful, and it'd follow the pattern of the CovarianceReducer.

tensorflow_probability/python/experimental/mcmc/expectations_reducer.py
```diff
@@ -0,0 +1,227 @@
...
+
+    Args:
+      final_reducer_state: `ExpectationsReducerState` that represents the
+        final reducer state.
+
+    Returns:
+      expectation: an estimate of the expectation with identical structure to
+        `final_reducer_state.running_total`.
+    """
+    if final_reducer_state.num_samples == 0:
```
SiegeLordEx, line 187: This won't work when num_samples isn't known. I'd drop this entire branch.

tensorflow_probability/python/experimental/mcmc/expectations_reducer.py
rupei: Yeah, that's a good point. I thought expectations was trivial enough to not warrant a separate algorithm in sample_stats, but it doesn't hurt!

SiegeLordEx: Certainly this algorithm is simple, but if we were to use the same algorithm as running covariance (i.e. `new_mean = state.mean + chunk_n * delta_mean / new_n`), then that's less trivial, and possibly more robust (certainly to overflow).

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -269,3 +273,169 @@ def __init__(self, shape=(), dtype=tf.float32):
...
+  which includes intermediate calculations, are held in a
+  `RunningExpectationsState` as returned via `initialize` and `update` method
+  calls.
+
+  `RunningExpectations` is meant to serve general streaming expectations.
+  For a specialized version that fits streaming over MCMC samples, see
+  `ExpectationsReducer` in `tfp.experimental.mcmc`.
+  """
+
+  def __init__(self, shape, callables, dtype=tf.float32):
```
SiegeLordEx, line 307: Why does this API need callables? I think the value transformation should be responsibility of the user code (just like it is with RunningCovariance).

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -269,3 +273,169 @@ def __init__(self, shape=(), dtype=tf.float32):
         By default, the dtype is `tf.float32`.
     """
     super(RunningVariance, self).__init__(shape, event_ndims=0, dtype=dtype)
+
+
+RunningExpectationsState = collections.namedtuple(
+    'RunningExpectationsState', 'num_samples, expectation')
+
+
+class RunningExpectations(object):
```
SiegeLordEx, line 282: Why not RunningMean?

tensorflow_probability/python/experimental/mcmc/expectations_reducer.py
```diff
@@ -0,0 +1,189 @@
...
+        mcmc_util.make_name(self.name, 'expectations_reducer', 'one_step')):
+      new_chain_state = tf.nest.map_structure(
+          tf.convert_to_tensor,
+          new_chain_state)
+      stream = sample_stats.RunningExpectations(
+          shape=new_chain_state.shape,
+          callables=self.callables,
+          dtype=new_chain_state.dtype
+      )
+      if previous_kernel_results is None:
```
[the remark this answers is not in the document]
rupei: You're absolutely right, in this context it should never be None. Good catch!

tensorflow_probability/python/experimental/mcmc/expectations_reducer.py
```diff
@@ -0,0 +1,189 @@
...
+  finalized expectation will also identically mimic the structure of
+  `callables`.
+
+  As with all reducers, ExpectationsReducer does not hold state information;
+  rather, it stores supplied metadata. Intermediate calculations are held in
+  a state object, which is returned via `initialize` and `one_step` method
+  calls.
+  """
+
+  def __init__(self, callables=_get_sample, name=None):
```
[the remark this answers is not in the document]
rupei: `value_fn` seems a bit too thin of a name. Do you think "transform" in tfp context has a connotation associated with bijectors? If not, `transform_fn` sounds good. `

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: Yeah, that makes sense.

tensorflow_probability/python/experimental/mcmc/expectations_reducer.py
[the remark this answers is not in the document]
SiegeLordEx: I don't think we use `transform` for bijectors.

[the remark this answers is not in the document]
rupei: I was thinking about how TransformedTransitionKernel is basically a bijector application wrapper. But if that's not an issue, then I think I'll go ahead and use `transform_fn` for now.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -266,6 +271,111 @@ def __init__(self, shape=(), dtype=tf.float32):
...
+
+  def __init__(self, shape, dtype=tf.float32):
+    """Instantiates this object.
+
+    Args:
+      shape: Python `Tuple` or `TensorShape` representing the shape of
+        incoming samples.
+      dtype: Dtype of incoming samples and the resulting statistics.
+        By default, the dtype is `tf.float32`. Any integer dtypes will also
+        be treated as `tf.float32` (to not lose significant precision).
```
axch, line 311: This comment isn't quite accurate.  A 32-bit float has only 23 bits of precision for representing integers, whereas a 32-bit int has 31.  And, of course, an int64 has 63 bits of precision for representing integers.  What you meant was that we need floating-point Tensors because the running mean should be doing floating-point rather than integer division, so say that.  You might also consider a more thoughtful casting strategy that uses 64-bit floating-point Tensors for running means on 64-bit integers.

PR #1047, opened 2020-08-07
Adding `TracingReducer` and a recreated `sample_chain` to tfp.experimental

rupei:
`TracingReducer` allows for accumulating arbitrary functions of samples and kernel results in the streaming MCMC framework. We can now also rewrite `sample_chain` in terms of `TracingReducer` and `sample_fold`, which is also included.

### DIFF
--- a/tensorflow_probability/python/experimental/mcmc/BUILD
+++ b/tensorflow_probability/python/experimental/mcmc/BUILD
@@ -435,7 +435,6 @@ py_library(
 py_test(
     name = "sample_fold_test",
     size = "small",
-    timeout = "long",
     srcs = ["sample_fold_test.py"],
     shard_count = 5,
     deps = [
--- a/tensorflow_probability/python/experimental/mcmc/BUILD
+++ b/tensorflow_probability/python/experimental/mcmc/BUILD
@@ -46,6 +46,7 @@ py_library(
         ":sample_fold",
         ":sample_sequential_monte_carlo",
         ":sequential_monte_carlo_kernel",
+        ":tracing_reducer",
         ":weighted_resampling",
         ":with_reductions",
         # tensorflow dep,
@@ -422,9 +423,11 @@ py_library(
     deps = [
         ":sample",
         ":sample_discarding_kernel",
+        ":tracing_reducer",
         ":with_reductions",
         # numpy dep,
         # tensorflow dep,
+        "//tensorflow_probability/python/mcmc/internal:util",
     ],
 )
 
@@ -432,6 +435,7 @@ py_library(
 py_test(
     name = "sample_fold_test",
     size = "small",
+    timeout = "long",
     srcs = ["sample_fold_test.py"],
     shard_count = 5,
     deps = [
@@ -459,6 +463,30 @@ py_test(
     name = "sample_discarding_kernel_test",
     size = "small",
     srcs = ["sample_discarding_kernel_test.py"],
+    deps = [
+        # numpy dep,
+        # tensorflow dep,
+        "//tensorflow_probability",
+        "//tensorflow_probability/python/internal:test_util",
+    ],
+)
+
+py_library(
+    name = "tracing_reducer",
+    srcs = ["tracing_reducer.py"],
+    srcs_version = "PY2AND3",
+    deps = [
+        # tensorflow dep,
+        ":reducer",
+        "//tensorflow_probability/python/mcmc/internal:util",
+    ],
+)
+
+# py2and3
+py_test(
+    name = "tracing_reducer_test",
+    size = "small",
+    srcs = ["tracing_reducer_test.py"],
     deps = [
         # numpy dep,
         # tensorflow dep,
--- a/tensorflow_probability/python/experimental/mcmc/sample_fold.py
+++ b/tensorflow_probability/python/experimental/mcmc/sample_fold.py
@@ -18,15 +18,21 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import collections
+import warnings
+
 # Dependency imports
 import tensorflow.compat.v2 as tf
 from tensorflow_probability.python.experimental.mcmc import sample
 from tensorflow_probability.python.experimental.mcmc import sample_discarding_kernel
 from tensorflow_probability.python.experimental.mcmc import with_reductions
+from tensorflow_probability.python.experimental import mcmc
+from tensorflow_probability.python.mcmc.internal import util as mcmc_util
 from tensorflow.python.util import nest  # pylint: disable=g-direct-tensorflow-import
 
 
 __all__ = [
+    'sample_chain',
     'sample_fold',
 ]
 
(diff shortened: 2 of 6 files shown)


tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -133,3 +139,174 @@ def sample_fold(
     return (reduction_results,
             end_state,
             final_kernel_results.inner_results.inner_results)
+
+
+class StatesAndTrace(
```
axch, line 144: Why copy these from tfp.mcmc instead of just reusing them?

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -133,3 +139,174 @@ def sample_fold(
...
+    num_steps_between_results=0,
+    trace_fn=lambda current_state, kernel_results: kernel_results,
+    return_final_kernel_results=False,
+    parallel_iterations=10,
+    seed=None,
+    name=None,
+):
+  """Implements Markov chain Monte Carlo via repeated `TransitionKernel` steps.
+
+  This function samples from an Markov chain at `current_state` and whose
```
axch, line 197:
```suggestion
  This function samples from a Markov chain at `current_state` whose
```

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -133,3 +139,174 @@ def sample_fold(
...
+  stationary distribution is governed by the supplied `TransitionKernel`
+  instance (`kernel`).
+
+  This function can sample from multiple chains, in parallel. (Whether or not
+  there are multiple chains is dictated by the `kernel`.)
```
axch, line 202: What does this parenthetical comment mean?

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -133,3 +139,174 @@ def sample_fold(
...
+        current_state=current_state,
+        previous_kernel_results=previous_kernel_results,
+        kernel=kernel,
+        reducer=tracing_reducer,
+        num_burnin_steps=num_burnin_steps,
+        num_steps_between_results=num_steps_between_results,
+        parallel_iterations=parallel_iterations,
+        seed=seed,
+        name=name,
+    )
```
axch, line 300: Neat!  I was hoping this would turn out to be this simple :)

tensorflow_probability/python/experimental/mcmc/BUILD
```diff
@@ -422,16 +423,19 @@ py_library(
...
         # tensorflow dep,
+        "//tensorflow_probability/python/mcmc/internal:util",
     ],
 )
 
 # py2and3
 py_test(
     name = "sample_fold_test",
     size = "small",
+    timeout = "long",
```
axch, line 438: Let's try to invest some energy in this not happening.

If the timeout is rising because of too many medium-size test methods adding up, increase the shard count instead.  If there's a single very long test method, see whether it can be avoided.  For instance, there's a `num_samples = 3000` at the end -- maybe don't run that one in all execution regimes?

(BTW, it's reasonable to have two test classes, one decorated with all execution regimes and one with something less profligate.)

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -133,3 +139,174 @@ def sample_fold(
...
+    else:
+      no_trace = False
+
+    if trace_fn is sample_chain.__defaults__[4]:
+      warnings.warn('Tracing all kernel results by default is deprecated. Set '
+                    'the `trace_fn` argument to None (the future default '
+                    'value) or an explicit callback that traces the values '
+                    'you are interested in.')
+
+    tracing_reducer = mcmc.TracingReducer(
```
axch, line 285: Could you import the module for TracingReducer directly above, instead of trying to go through the package?  IIRC, importing the package caused problems in Python 2.

tensorflow_probability/python/experimental/mcmc/tracing_reducer.py
```diff
@@ -0,0 +1,201 @@
...
+  accumulates all results. By default, all kernel results are traced but in the
+  future the default will be changed to no results being traced, so plan
+  accordingly.
+
+  If wrapped in a `tfp.experimental.mcmc.WithReductions` Transition Kernel,
+  TracingReducer will not accumulate the kernel results of `WithReductions`.
+  Rather, the top level kernel results will be that of `WithReductions`' inner
+  kernel.
+
+  As with all reducers, TracingReducer does not hold state information;
```
axch, line 57:
```suggestion
  As with all reducers, `TracingReducer` does not hold state information;
```

tensorflow_probability/python/experimental/mcmc/tracing_reducer.py
```diff
@@ -0,0 +1,201 @@
...
+        with shape and dtype compatible with those used to initialize the
+        `TracingState`.
+      current_reducer_state: `TracingState`s representing all previously traced
+        results.
+      previous_kernel_results: A (possibly nested) structure of `Tensor`s
+        representing internal calculations made in a related
+        `TransitionKernel`.
+
+    Returns:
+      new_reducer_state: `TracingState` with updated trace. Its 'trace_state'
```
axch, line 156:
```suggestion
      new_reducer_state: `TracingState` with updated trace. Its `trace_state`
```

tensorflow_probability/python/experimental/mcmc/sample_fold.py
rupei: good question, reusing is probably better

tensorflow_probability/python/experimental/mcmc/sample_fold.py
rupei: I was actually confused about this too... it's in the Docstring of `sample_chain` so I assumed it applied to some other kernels in mcmc that I haven't looked at in depth

tensorflow_probability/python/experimental/mcmc/sample_fold.py
rupei: yes, absolutely! I made this PR before I asked about this weird spec so I didn't know about the py2 incompatibility

rupei:
check that `sample_chain` doesn't nest any more kernel results

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -18,15 +18,20 @@
 from __future__ import division
 from __future__ import print_function
 
+import warnings
+
 # Dependency imports
 import tensorflow.compat.v2 as tf
-from tensorflow_probability.python.experimental.mcmc import sample
+from tensorflow_probability.python.experimental.mcmc import sample as exp_sample
```
SiegeLordEx, line 25: We usually call these '_lib', i.e. `sample_lib`

tensorflow_probability/python/experimental/mcmc/sample_fold.py
```diff
@@ -133,3 +138,145 @@ def sample_fold(
...
+    else:
+      no_trace = False
+
+    if trace_fn is sample_chain.__defaults__[4]:
+      warnings.warn('Tracing all kernel results by default is deprecated. Set '
+                    'the `trace_fn` argument to None (the future default '
+                    'value) or an explicit callback that traces the values '
+                    'you are interested in.')
+
+    # `WithReductions` assumes all its reducers want to reducer over the
```
SiegeLordEx, line 250: Want to *reduce*

tensorflow_probability/python/experimental/mcmc/tracing_reducer.py
```diff
@@ -0,0 +1,202 @@
...
+  Rather, the top level kernel results will be that of `WithReductions`' inner
+  kernel.
+
+  As with all reducers, `TracingReducer` does not hold state information;
+  rather, it stores supplied metadata. Intermediate calculations are held in
+  a `TracingState` named tuple, which is returned via `initialize` and
+  `one_step` method calls.
+  """
+
+  # TODO(Ru): bring the lambda fn out
```
SiegeLordEx, line 63: Don't forge this TODO

tensorflow_probability/python/experimental/mcmc/sample_fold.py
rupei: gotcha, I called it `exp_sample_lib` just to put emphasis on the fact that it's sample from experimental (the normal mcmc.sample is imported as well)

tensorflow_probability/python/experimental/mcmc/tracing_reducer.py
```diff
@@ -0,0 +1,205 @@
...
```
SiegeLordEx, line 67: There's still the comment left.

PR #1053, opened 2020-08-12
Fix typo in real_nvp_default_template documentation

Jimpachnet:
Missing closure of inline code env.


Jimpachnet:
@googlebot I signed it!

Issue #1054, opened 2020-08-13
`tfp.mcmc.potential_scale_reduction` incorrectly implements Gelman and Rubin (1992)'s r-hat algorithm


rupei:
In implementing the streaming version of r-hat, I discovered that our non-streaming diagnostic incorrectly implements the referenced r-hat algorithm. The actual procedure is outlined both in Gelman and Rubin (1992) and in the earlier parts of Brooks and Gelman (1998). Both papers are referenced in the Docstring.

More precisely, two mistakes occur in the `_potential_scale_reduction_single_state` helper function.

1. Instead of taking the sample variance for the chain samples, the population variance is used instead (note the `biased=True` kwarg when calling `_reduce_variance`) to compute `w` -- the within-chain variance in the language of Brooks and Gelman (1988).
2. An incorrect calculation of `sigma_2_plus` is used. The \frac{n-1}{n} term before `w` is dropped.

My hypothesis is that the author realized that taking the population variance would simplifies the calculation of `sigma_2_plus`. However, the original definition of `w` is used again in the final (over)estimate of the statistic; yet, no adjustment is made in the implementation. Hence, the computation is actually incorrectly implemented.

It's also worth noting that this has slipped under the radar as tests for potential scale reduction check to see if the statistic is close to 1 (chain has converged) or greater than 1.2 (chain has not converged). There are no checks to ensure the actual statistic is calculated correctly.

SiegeLordEx:
I believe you're right, although I'd just phrase this as the final r_hat formula missing the `(n - 1) / n` factor in front of `w`, the actual value of `sigma_2_plus` is fine. That said, perhaps in fixing this we could just use the original formulation to aid future readers.

> It's also worth noting that this has slipped under the radar as tests for potential scale reduction check to see if the statistic is close to 1 (chain has converged) or greater than 1.2 (chain has not converged). There are no checks to ensure the actual statistic is calculated correctly.

If the error was in transcribing the formula from the paper, presumably the same error would be made in the test as well. Such correctness checks are only possible when we have some analytic way of computing a quantity, which we don't really have in this case.

rupei:
Thanks for the feedback Pavel! You're right... my explanation is probably more verbose than required and not entirely precise. I'll go ahead and edit for clarity. In terms of commenting on the testing I was more so foreshadowing a couple of tests that I was going to add that checks the r hat against a manual computation using a combination of numpy operations.

PR #1055, opened 2020-08-16
Adding `RhatReducer` and streaming rhat to tfp.experimental

rupei:
This PR accomplishes three things:

1. Add a base implementation of Gelman and Rubin (1992)'s potential scale reduction (or rhat) in streaming fashion to `tfp.experimental.stats` in fashion similar to `tfp.mcmc.potential_scale_reduction`, but without the use of certain flags.
2. Add a RhatReducer to `tfp.experimental.mcmc`
3. Update `tfp.mcmc.potential_scale_reduction` to correctly implement Gelman and Rubin (1992)'s formula. This addresses and fixes https://github.com/tensorflow/probability/issues/1054

### DIFF
--- a/tensorflow_probability/python/experimental/mcmc/rhat_reducer.py
+++ b/tensorflow_probability/python/experimental/mcmc/rhat_reducer.py
@@ -1,0 +1,203 @@
+# Copyright 2020 The TensorFlow Probability Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ============================================================================
+"""RhatReducer for calculating streaming potential scale reduction."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import collections
+
+# Dependency imports
+import tensorflow.compat.v2 as tf
+
+from tensorflow_probability.python.experimental.mcmc import reducer as reducer_base
+from tensorflow_probability.python.experimental.stats import sample_stats
+from tensorflow_probability.python.internal import prefer_static as ps
+from tensorflow_probability.python.mcmc.internal import util as mcmc_util
+
+
+__all__ = [
+    'RhatReducer',
+]
+
+
+RhatReducerState = collections.namedtuple(
+    'RhatReducerState', 'sample_shape, rhat_state')
+
+
+class RhatReducer(reducer_base.Reducer):
+  """`Reducer` that computes a running rhat diagnostic statistic.
+
+  `RhatReducer` assumes that all provided chain states include samples from
+  multiple independent Markov chains, and that all of these chains are to
+  be included in the same calculation. `RhatReducer` also assumes that
+  incoming samples have shape `[Ni, Ci1, Ci2,...,CiD]`. Dimension `0` indexes
+  the `Ni > 1` result steps of the Markov Chain. Dimensions `1` through `D`
+  index the `Ci1 x ... x CiD` independent chains to be tested for convergence
+  to the same target. A possible chunking dimension can also be specified via
+  the `axis` parameter of the `one_step` method. This dimension indexes multiple
+  dependent samples from the same Markov chain, but disregards the first sample
+  dimension.
+
+  To illustrate, with no chunking enabled, a chain state of shape `[5, 2, 3, 4]`
+  can be inferred as shape `[2, 3, 4]` samples drawn from 5 independent Markov
+  chains. If chunking is enabled and `axis=1` then the same chain state can be
+  inferred as 3 samples of shape `[2, 4]` from 5 independent Markov chains. Note
+  how the chunking dimension indexes into only the event dimensions, and not the
+  batch dimension.
+
+  As with all reducers, RhatReducer does not hold state information;
+  rather, it stores supplied metadata. Intermediate calculations are held in
+  a state object, which is returned via `initialize` and `one_step` method
+  calls.
+
+  `RhatReducer` is meant to fit into the larger Streaming MCMC framework.
+  `RunningPotentialScaleReduction` in `tfp.experimental.stats` is
+  better suited for more generic streaming covariance needs. More precise
+  algorithmic details can also be found by referencing
+  `RunningPotentialScaleReduction`.
+  """
+
+  def __init__(self, num_parallel_chains, name=None):
+    """Instantiates this object.
+
+    Args:
+      num_parallel_chains: Integer number of independent Markov chains
+        ran in parallel.
+      name: Python `str` name prefixed to Ops created by this function.
+        Default value: `None` (i.e., 'rhat_reducer').
+
+    Raises:
+      ValueError: if `num_parallel_chains < 2`. This results in undefined
+        intermediate variance calculations.
+    """
+    if num_parallel_chains < 2:
+      raise ValueError('At least 2 parallel chains are required for '
+                       'well-defined between_chain variances')
+    self._parameters = dict(
+        num_parallel_chains=num_parallel_chains,
+        name=name or 'rhat_reducer'
+    )
+
+  def initialize(self, initial_chain_state, initial_kernel_results=None):
+    """Initializes a `RhatReducerState` using previously defined metadata.
+
+    For calculation purposes, the `initial_chain_state` does not count as a
+    sample. This is a deliberate decision that ensures consistency across
+    sampling procedures (i.e. `tfp.mcmc.sample_chain` follows similar
+    semantics).
+
+    Args:
+      initial_chain_state: A (possibly nested) structure of `Tensor`s or Python
+        `list`s of `Tensor`s representing the current state(s) of the Markov
+        chain(s). It is used to infer the shape and dtype of future samples.
+      initial_kernel_results: A (possibly nested) structure of `Tensor`s
+        representing internal calculations made in a related `TransitionKernel`.
+        For streaming rhat, this argument has no influence on the
+        computation; hence, it is `None` by default. However, it's
+        still accepted to fit the `Reducer` base class.
+
+    Returns:
+      state: `RhatReducerState` with `rhat_state` field representing
+        a stream of no inputs.
+    """
+    with tf.name_scope(
+        mcmc_util.make_name(self.name, 'rhat_reducer', 'initialize')):
+      initial_chain_state = tf.convert_to_tensor(initial_chain_state)
+      if ps.rank(initial_chain_state) == 1:
+        sample_shape = ()
+      else:
+        sample_shape = ps.shape(initial_chain_state)[1:]
+      running_rhat = sample_stats.RunningPotentialScaleReduction(
+          shape=sample_shape,
+          num_chains=self.num_parallel_chains,
+          dtype=initial_chain_state.dtype
+      )
+      return RhatReducerState(sample_shape, running_rhat.initialize())
+
+  def one_step(
+      self,
+      new_chain_state,
+      current_reducer_state,
+      previous_kernel_results=None,
+      axis=None):
+    """Update the `current_reducer_state` with a new chain state.
+
+    Args:
+      new_chain_state: A (possibly nested) structure of incoming chain state(s)
+        with shape and dtype compatible with those used to initialize the
+        `current_reducer_state`.
+      current_reducer_state: `RhatReducerState` representing the current
+        state of the running rhat statistic.
+      previous_kernel_results: A (possibly nested) structure of `Tensor`s
+        representing internal calculations made in a related
+        `TransitionKernel`. For streaming rhat, this argument has no
+        influence on computation; hence, it is `None` by default. However, it's
+        still accepted to fit the `Reducer` base class.
+      axis: If chunking is desired, this is a (possibly nested) structure of
+        integers that specifies the axis with chunked samples. For individual
+        samples, set this to `None`. By default, samples are not chunked
+        (`axis` is None).
+
+    Returns:
+      new_reducer_state: `RhatReducerState` with updated running
+        statistics.
+    """
+    with tf.name_scope(
+        mcmc_util.make_name(self.name, 'rhat_reducer', 'one_step')):
+      new_chain_state = tf.convert_to_tensor(new_chain_state)
+      running_rhat = sample_stats.RunningPotentialScaleReduction(
+          shape=current_reducer_state.sample_shape,
+          num_chains=self.num_parallel_chains,
+          dtype=new_chain_state.dtype
+      )
+      new_rhat_state = running_rhat.update(
+          current_reducer_state.rhat_state,
+          new_chain_state,
+          chunk_axis=axis)
+      return RhatReducerState(
+          current_reducer_state.sample_shape, new_rhat_state)
+
+  def finalize(self, final_reducer_state):
+    """Finalizes covariance calculation from the `final_reducer_state`.
+
+    Args:
+      final_reducer_state: `RhatReducerState` that represents the
+        final state of the running rhat statistic.
+
+    Returns:
+      rhat: an estimate of the rhat.
+    """
+    with tf.name_scope(
+        mcmc_util.make_name(self.name, 'rhat_reducer', 'finalize')):
+      running_rhat = sample_stats.RunningPotentialScaleReduction(
+          shape=final_reducer_state.sample_shape,
+          num_chains=self.num_parallel_chains,
+          dtype=final_reducer_state.rhat_state.chain_var[0].mean.dtype
+      )
+      return running_rhat.finalize(final_reducer_state.rhat_state)
+
+  @property
+  def parameters(self):
+    return self._parameters
+
+  @property
+  def num_parallel_chains(self):
+    return self._parameters['num_parallel_chains']
+
+  @property
+  def name(self):
+    return self._parameters['name']
--- a/tensorflow_probability/python/experimental/mcmc/rhat_reducer_test.py
+++ b/tensorflow_probability/python/experimental/mcmc/rhat_reducer_test.py
@@ -1,0 +1,206 @@
+# Copyright 2020 The TensorFlow Probability Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ============================================================================
+"""Tests for RhatReducer."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import collections
+
+# Dependency imports
+
+import numpy as np
+
+import tensorflow.compat.v2 as tf
+import tensorflow_probability as tfp
+from tensorflow_probability.python.internal import test_util
+
+
+TestTransitionKernelResults = collections.namedtuple(
+    'TestTransitionKernelResults', 'counter_1, counter_2')
+
+
+class TestTransitionKernel(tfp.mcmc.TransitionKernel):
+  """Fake deterministic Transition Kernel."""
+
+  def __init__(self, shape=(), target_log_prob_fn=None, is_calibrated=True):
+    self._is_calibrated = is_calibrated
+    self._shape = shape
+    # for composition purposes
+    self.parameters = dict(
+        target_log_prob_fn=target_log_prob_fn)
+
+  def one_step(self, current_state, previous_kernel_results, seed=None):
+    return (current_state + tf.ones(self._shape),
+            TestTransitionKernelResults(
+                counter_1=previous_kernel_results.counter_1 + 1,
+                counter_2=previous_kernel_results.counter_2 + 2))
+
+  def bootstrap_results(self, current_state):
+    return TestTransitionKernelResults(
+        counter_1=tf.zeros(()),
+        counter_2=tf.zeros(()))
+
+  @property
+  def is_calibrated(self):
+    return self._is_calibrated
+
+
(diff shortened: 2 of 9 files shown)


tensorflow_probability/python/mcmc/diagnostic.py
```diff
@@ -492,6 +493,18 @@ def potential_scale_reduction(chains_states,
 def _potential_scale_reduction_single_state(state, independent_chain_ndims,
                                             split_chains, validate_args):
   """potential_scale_reduction for one single state `Tensor`."""
+  # casting integers to floats for floating-point division
+  # check to see if the `state` is a numpy object for the numpy test suite
+  if type(state).__module__ == np.__name__:
```
rupei, line 498: This seems very kluge-like to me of a fix. Would be nice if there was a general check for both numpy and tf ints

tensorflow_probability/python/experimental/mcmc/rhat_reducer.py
```diff
@@ -0,0 +1,203 @@
...
+    'RhatReducer',
+]
+
+
+RhatReducerState = collections.namedtuple(
+    'RhatReducerState', 'sample_shape, rhat_state')
+
+
+class RhatReducer(reducer_base.Reducer):
+  """`Reducer` that computes a running rhat diagnostic statistic.
```
axch, line 42:
```suggestion
  """`Reducer` that computes a running R-hat diagnostic statistic.
```

tensorflow_probability/python/experimental/mcmc/rhat_reducer.py
```diff
@@ -0,0 +1,203 @@
...
+  batch dimension.
+
+  As with all reducers, RhatReducer does not hold state information;
+  rather, it stores supplied metadata. Intermediate calculations are held in
+  a state object, which is returned via `initialize` and `one_step` method
+  calls.
+
+  `RhatReducer` is meant to fit into the larger Streaming MCMC framework.
+  `RunningPotentialScaleReduction` in `tfp.experimental.stats` is
+  better suited for more generic streaming covariance needs. More precise
```
axch, line 69:
```suggestion
  better suited for more generic streaming R-hat needs. More precise
```

tensorflow_probability/python/experimental/mcmc/rhat_reducer.py
```diff
@@ -0,0 +1,203 @@
...
+  algorithmic details can also be found by referencing
+  `RunningPotentialScaleReduction`.
+  """
+
+  def __init__(self, num_parallel_chains, name=None):
+    """Instantiates this object.
+
+    Args:
+      num_parallel_chains: Integer number of independent Markov chains
+        ran in parallel.
```
axch, line 79: Is there a reason this is different from `tfp.mcmc.potential_scale_reduction_factor`?  There, the integer gives the number of leading Tensor _dimensions_ to be interpreted as independent chains, and the number of chains is inferred from the sizes of those dimensions.

tensorflow_probability/python/experimental/mcmc/rhat_reducer.py
```diff
@@ -0,0 +1,203 @@
...
+    'RhatReducerState', 'sample_shape, rhat_state')
+
+
+class RhatReducer(reducer_base.Reducer):
+  """`Reducer` that computes a running rhat diagnostic statistic.
+
+  `RhatReducer` assumes that all provided chain states include samples from
+  multiple independent Markov chains, and that all of these chains are to
+  be included in the same calculation. `RhatReducer` also assumes that
+  incoming samples have shape `[Ni, Ci1, Ci2,...,CiD]`. Dimension `0` indexes
```
axch, line 47: Why not also allow an arbitrary "event shape", as in `tfp.mcmc.potential_scale_reduction_factor`?

tensorflow_probability/python/experimental/mcmc/rhat_reducer.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    if num_parallel_chains < 2:
+      raise ValueError('At least 2 parallel chains are required for '
+                       'well-defined between_chain variances')
+    self._parameters = dict(
+        num_parallel_chains=num_parallel_chains,
+        name=name or 'rhat_reducer'
+    )
+
+  def initialize(self, initial_chain_state, initial_kernel_results=None):
```
axch, line 95: Does it make sense to define potential scale reduction for a transformation of the latents?  Or is this unlike, say, covariance in that respect?  @SiegeLordEx ?

tensorflow_probability/python/experimental/mcmc/rhat_reducer.py
```diff
@@ -0,0 +1,203 @@
...
+        computation; hence, it is `None` by default. However, it's
+        still accepted to fit the `Reducer` base class.
+
+    Returns:
+      state: `RhatReducerState` with `rhat_state` field representing
+        a stream of no inputs.
+    """
+    with tf.name_scope(
+        mcmc_util.make_name(self.name, 'rhat_reducer', 'initialize')):
+      initial_chain_state = tf.convert_to_tensor(initial_chain_state)
```
axch, line 119: This is not consistent with the claim that the chain state can be an arbitrary structure of Tensors.  `tfp.mcmc.potential_scale_reduction_factor` seems to support at least a list of Tensors, which suggests that an arbitrary structure should not be difficult.  Can we do that?

tensorflow_probability/python/experimental/mcmc/rhat_reducer_test.py
```diff
@@ -0,0 +1,206 @@
...
+# Dependency imports
+
+import numpy as np
+
+import tensorflow.compat.v2 as tf
+import tensorflow_probability as tfp
+from tensorflow_probability.python.internal import test_util
+
+
+TestTransitionKernelResults = collections.namedtuple(
```
axch, line 32: These should come from `test_fixtures` when that merges, right?

tensorflow_probability/python/experimental/mcmc/rhat_reducer_test.py
```diff
@@ -0,0 +1,206 @@
...
+        num_parallel_chains=3,
+    )
+    state = rhat_reducer.initialize(offset_samples[0])
+    for sample in offset_samples:
+      state = rhat_reducer.one_step(sample, state)
+    rhat = self.evaluate(rhat_reducer.finalize(state))
+    self.assertAllEqual((4,), rhat.shape)
+    self.assertAllEqual(np.ones_like(rhat).astype(bool), rhat > 1.2)
+
+  def test_with_hmc(self):
```
axch, line 181: Can we also add a test with more than one latent Tensor, following some of the tests of `tfp.mcmc.potential_scale_reduction_factor`?

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -22,15 +22,19 @@
...
 import tensorflow.compat.v2 as tf
-
+from tensorflow_probability.python.internal import nest_util
 from tensorflow_probability.python.internal import prefer_static as ps
+from tensorflow_probability.python.mcmc import diagnostic
+from tensorflow.python.util import nest  # pylint: disable=g-direct-tensorflow-import
 
 
 __all__ = [
     'RunningMean',
```
axch, line 32: Alphabetical order here too, please.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +393,185 @@ def finalize(self, state):
...
     """
     return state.mean
+
+
+RunningPotentialScaleReductionState = collections.namedtuple(
+    'RunningPotentialScaleReductionState', 'chain_var')
+
+
+class RunningPotentialScaleReduction(object):
+  """Holds metadata for and computes a running rhat diagnostic statistic.
```
axch, line 403:
```suggestion
  """Holds metadata for and computes a running R-hat diagnostic statistic.
```

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +393,185 @@ def finalize(self, state):
...
+
+  `RunningPotentialScaleReduction` uses Gelman and Rubin (1992)'s potential
+  scale reduction (or rhat) for chain convergence [1]. This object also
```
axch, line 406:
```suggestion
  scale reduction (also known as R-hat) for chain convergence [1]. This object also
```

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +393,185 @@ def finalize(self, state):
...
+  supports both batching and chunking.
+
+  If multiple rhat computations are desired (batching), one should use a
```
axch, line 409:
```suggestion
  If multiple R-hat computations are desired (batching), one should use a
```

axch: Also elsewhere.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +393,185 @@ def finalize(self, state):
...
+  (possibly nested) collection for initialization parameters `num_chains`
+  and `shape`. Subsequent chain states used to update the streaming rhat
+  should mimic their identical structure.
+
+  In computation, samples can be provided individually or in chunks. A
+  "chunk" of size M implies incorporating M samples into a single expectation
```
axch, line 415:
```suggestion
  "chunk" of size M implies incorporating M samples into a single R-hat
```

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +393,185 @@ def finalize(self, state):
...
+  """
+
+  def __init__(self, shape, num_chains, dtype=tf.float32):
+    """Instantiates this object.
+
+    Args:
+      shape: Python `Tuple` or `TensorShape` representing the shape of
+        incoming samples. Using a collection enables batching and implies
+        that future samples will mimic that exact structure.
+      num_chains: A (possibly nested) collection of integers representing
```
axch, line 441: ??  Is this code batching over different Tensors in a list structure?  That's very unconventional in TFP; generally we batch over Tensor dimensions.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +393,185 @@ def finalize(self, state):
...
+
+    Returns:
+      state: `RunningPotentialScaleReductionState` representing a stream
+        of no inputs.
+    """
+    def _initialize_for_one_state(num_chains, shape):
+      """Initializes a running variance state for one group of Markov chains."""
+      var_stream = RunningVariance(shape, self.dtype)
+      return [var_stream.initialize()
+              for _ in range(num_chains)]
```
axch, line 470: ???  Why not rely on batching in the underlying variance reducer?

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +393,185 @@ def finalize(self, state):
...
+  def update(self, state, new_sample, chain_axis=0, chunk_axis=None):
+    """Update the `RunningPotentialScaleReductionState` with a new sample.
+
+    Args:
+      state: `RunningPotentialScaleReductionState` that represents the
+        current state of running statistics.
+      new_sample: Incoming `Tensor` sample or (possibly nested) collection of
+        `Tensor`s with shape and dtype compatible with those used to form the
+        `RunningPotentialScaleReductionState`.
+      chain_axis: The sample axis that indexes into independent Markov chains
```
axch, line 490: This looks different from `tfp.mcmc.potential_scale_reduction_factor`.  Is it?  Is there a reason to be different?

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +393,185 @@ def finalize(self, state):
...
+    def _update_for_one_state(
+        num_chains, chain_var, new_sample, chain_axis, chunk_axis):
+      """Updates the running variance for one group of Markov chains."""
+      sample_rank = ps.rank(new_sample)
+      if sample_rank >= 2:
+        # make the axis denoting independent chains the leading dimension
+        perm = [chain_axis] + list(
+            range(0, chain_axis)) + list(
+                range(chain_axis + 1, sample_rank))
+        new_sample = tf.transpose(new_sample, perm)
```
axch, line 520: What weakness of the underlying API makes this transposition necessary?

tensorflow_probability/python/experimental/mcmc/rhat_reducer_test.py
rupei: That's the idea

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: Yeah... good point. Not entirely sure what I was thinking with this

tensorflow_probability/python/experimental/mcmc/rhat_reducer.py
rupei: Pretty sure that's supported but I didn't explicitly state it in the Docstring... oops

tensorflow_probability/python/experimental/mcmc/rhat_reducer.py
rupei: Well I was thinking in the context of streaming MCMC, it would always be the leading Tensor dimension that contains the batched dimension/the independent chains (please correct me if I'm wrong) so I didn't see the need of some `independent_chain_ndims` kwarg. In hindsight, I probably don't need the user to provide the `num_parallel_chains` either.

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: It kind of is. Idea is that in the `sample_stats` version we don't force the user to conform to any sample shape semantics (i.e. [number of samples, number of chains, event shape]). Hence, the axis that specifies the independent chains can be any (I didn't think there would be a use-case where this is actually multiple axes, but again, may not have the right intuition on this).

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: that would be having a `chain_axis` argument -- we don't know which dimension after the leading in the sample actually represents the chain axis. We could simply streamline everything at make the incoming samples conform to a certain shape, like in `tfp.mcmc.potential_scale_reduction` and the reducer if you think this is more annoying than convenient (i.e. code complexity doesn't really make sense for a relatively unhelpful utility)

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: Ah, I remember now. My thought was that if we wanted to truly batch multiple R-hat calculations for whatever reason (as supported by `tfp.mcmc.potential_scale_reduction` -- one can imagine in a more complex way than `test_batching` in `sample_stats_test`), then you can't really pack all of that into a `Tensor` as the latents may be completely different from one another. Let me know if I started elaborating on a completely different point than the one you were trying to get at.

tensorflow_probability/python/experimental/mcmc/rhat_reducer_test.py
rupei: independent R-hat calculations on multiple states

tensorflow_probability/python/mcmc/diagnostic.py
SiegeLordEx: We have dtype_util.is_integer

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -22,15 +22,19 @@
...
 
 
 __all__ = [
-    'RunningMean',
-    'RunningMeanState',
     'RunningCovariance',
     'RunningCovarianceState',
+    'RunningMean',
+    'RunningMeanState',
+    'RunningPotentialScaleReduction',
```
SiegeLordEx, line 36: Eh... this is kind of MCMC specific. E.g. you can do a RunningCovariance on any dataset, but PotentialScaleReduction is basically only relevant to MCMC chains.

In this case it would have been fine to just implement everything in the reducer. I guess it's fine to leave it as you did it though, the sample_stats API is a bit less assumptive.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +393,169 @@ def finalize(self, state):
...
+
+  def finalize(self, state):
+    """Finalizes potential scale reduction computation for the `state`.
+
+    Args:
+      state: `RunningPotentialScaleReductionState` that represents
+        the current state of running statistics.
+
+    Returns:
+      R-hat: An estimate of the R-hat.
```
SiegeLordEx, line 533: `rhat:`, we pretend that it's a Python variable.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +393,169 @@ def finalize(self, state):
...
+  """
+
+  def __init__(self, shape, num_chains, dtype=tf.float32):
+    """Instantiates this object.
+
+    Args:
+      shape: Python `Tuple` or `TensorShape` representing the shape of
+        incoming samples. Using a collection implies that future samples
+        will mimic that exact structure.
+      num_chains: A (possibly nested) collection of integers representing
```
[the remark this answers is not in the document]
rupei: tfp.mcmc.potential_scale_reduction supports a list of independent R-hat calculations. Wanted to support at least that so naturally generalized to nests.

rupei:
A note on the updated scope of streaming rhat. This PR currently doesn't support chunking or reducing over some arbitrary `transform_fn` for the sake of time; though both are not difficult to add.

tensorflow_probability/python/mcmc/diagnostic.py
```diff
@@ -492,6 +494,12 @@ def potential_scale_reduction(chains_states,
 def _potential_scale_reduction_single_state(state, independent_chain_ndims,
                                             split_chains, validate_args):
   """potential_scale_reduction for one single state `Tensor`."""
+  # casting integers to floats for floating-point division
+  # check to see if the `state` is a numpy object for the numpy test suite
+  if state.dtype is tf.int64 or state.dtype is np:
```
SiegeLordEx, line 499: ?

tensorflow_probability/python/mcmc/diagnostic_test.py
```diff
@@ -410,6 +410,36 @@ def use_static_shape(self):
...
+    # manual computation with numpy operations
+    b_div_n = np.var(np.mean(chain_state, axis=0), ddof=1)
+    w = np.mean(np.var(chain_state, axis=0, ddof=1))
+    sigma_2_plus = 0.8 * w + b_div_n
+    n = 5
+    m = 3
+    manual_rhat = ((m + 1.) / m) * sigma_2_plus / w - (n - 1.) / (m * n)
+    self.assertNear(manual_rhat, rhat, err=1e-6)
+
+  def testIntegerSamples(self):
```
SiegeLordEx, line 429: Check the rhat dtype is what you expect.

tensorflow_probability/python/mcmc/diagnostic.py
SiegeLordEx: Maybe use dtype_util.as_numpy_dtype(state.dtype) is np.float64 as a generic check.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +395,175 @@ def finalize(self, state):
...
+        `tf.float32`), as intermediate calculations should be performing
+        floating-point division.
+    """
+    self.shape = shape
+    self.independent_chain_ndims = independent_chain_ndims
+    def _infer_one_chain_shape(shape, chain_ndims):
+      return ((functools.reduce(
+          (lambda x, y: x * y), (shape[:chain_ndims])),)
+              + shape[chain_ndims:])
+    self.one_chain_ndims_shape = nest.map_structure_up_to(
```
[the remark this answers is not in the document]
rupei: We're reshaping for when `independent_chain_ndims > 2`. Then, we'll just compress all the dimensions that represent the independent chains to take advantage of variance batching.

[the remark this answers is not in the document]
rupei: If there's a better way to do it that's implementable in a day, I'd definitely be open to hearing it!

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +395,175 @@ def finalize(self, state):
...
+        self.independent_chain_ndims,
+        _infer_one_chain_shape,
+        self.shape,
+        self.independent_chain_ndims,
+    )
+    def _cast_dtype(dtype):
+      if dtype is tf.int64 or tf.float64:
```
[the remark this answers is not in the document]
rupei: Ah yes, this was an oversight... I wasn't even trying to compare it against `tf.float64`. May have just been too tired to notice; thanks for catching this one!

tensorflow_probability/python/mcmc/diagnostic.py
rupei: Another oversight (meant to have np.int64 there)

tensorflow_probability/python/experimental/stats/sample_stats.py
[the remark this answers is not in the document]
SiegeLordEx: The way to do it is to copy the code here: https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/mcmc/diagnostic.py#L554-L561

But I'll take a TODO to consider changing this in the future as a fine resolution for now.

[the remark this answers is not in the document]
rupei: Makes a lot of sense! Also thanks so much for taking the time to explain why reshaping is bad, I would've never even known one should avoid it where possible.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +395,175 @@ def finalize(self, state):
...
+    """Finalizes potential scale reduction computation for the `state`.
+
+    Args:
+      state: `RunningPotentialScaleReductionState` that represents
+        the current state of running statistics.
+
+    Returns:
+      rhat: An estimate of the R-hat.
+    """
+    def _finalize_for_one_state(shape, chain_ndims, chain_var):
```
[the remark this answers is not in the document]
rupei: it is now haha

tensorflow_probability/python/experimental/stats/sample_stats.py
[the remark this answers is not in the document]
SiegeLordEx: Sorry, I was mistaken about the reshapes being bad per se (it's transposes that are the actual issue), the issue here was purely of clarity. Thanks for making the change either way!

PR #1056, opened 2020-08-17
Markdown typo

JacobFV:
Reading the markdown rendered docstring from Colab, I noticed inverted backticks.


davmre:
I just merged your fix, though it may take some time (likely the next TFP release?) to propagate to Colab. Thanks for reporting this and for making a PR!

PR #1059, opened 2020-08-17
Ensure that the BijectorCache reset method can be called

### LINKED ISSUE #1058: BijectorCache.reset method throws an exception
The `BijectorCache` class `reset` method throws the following exception:
`line 483, in reset`
`    type(self.forward), type(self.inverse))`
`TypeError: __init__() takes from 3 to 4 positional arguments but 5 were given`

This seems to be because the implementation of the `reset` method assumes that the `cache_type` function for the forward and inverse caches can be different, but the constructor uses the same function for both caches.

leroidauphin:
fix #1058

### DIFF
--- a/tensorflow_probability/python/internal/cache_util.py
+++ b/tensorflow_probability/python/internal/cache_util.py
@@ -115,7 +115,8 @@ class HashableWeakRef(weakref.ref):
     """
     if isinstance(referrent, np.generic):
       raise ValueError('Unable to weakref np.generic')
-    elif isinstance(referrent, np.ndarray):
+
+    if isinstance(referrent, np.ndarray):
       referrent.flags.writeable = False
     super(HashableWeakRef, self).__init__(referrent, callback)
     self._hash = hash_structure(referrent)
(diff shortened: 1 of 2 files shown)


tensorflow_probability/python/internal/cache_util.py
```diff
@@ -115,7 +115,8 @@ def __init__(self, referrent, callback=None):
     """
     if isinstance(referrent, np.generic):
       raise ValueError('Unable to weakref np.generic')
-    elif isinstance(referrent, np.ndarray):
```
johnamcleod, line 118: This change was in response to a linting error.

leroidauphin, line 118: This change was in response to a linting error.

emilyfertig:
Hi @leroidauphin -- thanks for the PR. In the last couple months we made major changes to bijector caching, so that bijectors access a common cache instead of maintaining instance-level caches. With these changes, we replaced the `reset` method with a `clear` method (which has better test coverage). Since the bug is fixed at head with `clear`, I'll close this (and the related issue).

PR #1061, opened 2020-08-17
Adding transform_fn capabilities to `tfp.experimental.mcmc.CovarianceReducer`

rupei:
Like with `tfp.experimental.mcmc.ExpectationsReducer`, I'm adding a kwarg to CovarianceReducer to allow the user to specify an arbitrary structure of functions to evaluate on the chain state and kernel results for covariance computation. This makes it much easier to compute covariance over the kernel results.

### DIFF
--- a/tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
+++ b/tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
@@ -41,8 +41,22 @@ CovarianceReducerState = collections.namedtuple(
     'CovarianceReducerState', 'init_structure, cov_state')
 
 
+def _get_sample(current_state, kernel_results):
+  del kernel_results
+  return current_state
+
+
 class CovarianceReducer(reducer_base.Reducer):
   """`Reducer` that computes a running covariance.
+
+  `CovarianceReducer` computes a running covariance on samples evaluated on
+  some arbitrary structure of `transform_fn`s. A `trasform_fn` is defined as
+  any function that accepts the chain state and kernel results, and ouptuts
+  a `Tensor` or nested structure of `Tensor`s to compute the covariance over.
+  To be explicit, if we denote a `transform_fn` as f(x, y), then
+  `CovarianceReducer` will compute Cov(f(x, y)). If some structure of
+  `transform_fn`s are given, the final statistic (as returned by the
+  `finalize` method) will mimic that exact structure.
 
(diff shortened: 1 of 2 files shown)


tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -41,9 +41,23 @@
...
 
+def _get_sample(current_state, kernel_results):
+  del kernel_results
+  return current_state
+
+
 class CovarianceReducer(reducer_base.Reducer):
   """`Reducer` that computes a running covariance.
 
+  `CovarianceReducer` computes a running covariance on samples evaluated on
```
axch, line 52: Around here and in the method docstring, say what the default means (i.e., covariance directly over the latent state, ignoring kernel results).

tensorflow_probability/python/experimental/mcmc/covariance_reducer.py
```diff
@@ -229,7 +258,7 @@ def finalize(self, final_reducer_state):
 
     Returns:
       covariance: an estimate of the covariance with identical structure to
```
axch, line 260: Identical structure to the _results_ of `self.transform_fn`, here and elsewhere.

To illustrate the difference, if `self.transform_fn = lambda x: [x, x+1]`, this code should compute the covariance of both `x` and `x+1`, even though there was nominally only one `transform_fn`.

tensorflow_probability/python/experimental/mcmc/covariance_reducer_test.py
```diff
@@ -131,17 +146,88 @@ def test_nested_with_batching_and_chunking(self):
...
+    x_reshaped = x.reshape(100, 10)
+    final_cov_reshaped = tf.reshape(final_cov, (10, 10))
+    self.assertEqual((5, 2), final_mean.shape)
+    self.assertAllClose(np.mean(x, axis=0), final_mean, rtol=1e-5)
+    self.assertEqual((5, 2, 5, 2), final_cov.shape)
+    self.assertAllClose(np.cov(x_reshaped.T, ddof=0),
+                        final_cov_reshaped,
+                        rtol=1e-5)
+
+  def test_latent_state_with_multiple_transform_fn(self):
```
axch, line 206: Also test that a transform_fn that returns a structure works as expected (in a tf.while_loop, no less).

If there is a reason to make it work differently than suggested in my previous comment, you will probably discover it.

PR #1062, opened 2020-08-17
Specifying semantics for parallel sampling in `sample_chain` and friends

rupei:
Current documentation doesn't explicitly specify that the `current_state` cannot be a `List` of `Tensor`s if parallel sampling is desired. Moreover, if one tries using a `List` of `Tensor`s, the error message is not really helpful. To make it more clear that the `current_state` should be a `Tensor` with a leading batch dimension, this PR simply explicitly states that in the Docstring of `sample_chain` and friends.

### DIFF
--- a/tensorflow_probability/python/mcmc/sample.py
+++ b/tensorflow_probability/python/mcmc/sample.py
@@ -100,10 +100,13 @@ def sample_chain(
   instance (`kernel`).
 
   This function can sample from multiple chains, in parallel. (Whether or not
-  there are multiple chains is dictated by the `kernel`.)
+  there are multiple chains is dictated by the `kernel`).
 
   The `current_state` can be represented as a single `Tensor` or a `list` of
-  `Tensors` which collectively represent the current state.
+  `Tensors` which collectively represent the current state. Note, if one desires
+  to sample from multiple chains in parallel, the `current_state` must be a
+  `Tensor` with a leading batch dimension representing the number of parallel
+  chains.
 
   Since MCMC states are correlated, it is sometimes desirable to produce
   additional intermediate states, and then discard them, ending up with a set of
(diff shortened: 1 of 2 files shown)


tensorflow_probability/python/mcmc/sample.py
```diff
@@ -100,10 +100,13 @@ def sample_chain(
...
   This function can sample from multiple chains, in parallel. (Whether or not
-  there are multiple chains is dictated by the `kernel`.)
+  there are multiple chains is dictated by the `kernel`).
 
   The `current_state` can be represented as a single `Tensor` or a `list` of
-  `Tensors` which collectively represent the current state.
+  `Tensors` which collectively represent the current state. Note, if one desires
+  to sample from multiple chains in parallel, the `current_state` must be a
+  `Tensor` with a leading batch dimension representing the number of parallel
+  chains.
```
brianwa84, line 109: The chain shape for parallel sampling is determined by the shape of `target_log_prob(*state)`. So the overall assertion that sample_chain can't do parallel sampling of multiple chains with a list of states is not correct.

rupei: Ah got it. Just for my own understanding, for a list of `Tensor`s to work, that means the `target_log_prob` would've had to be defined to accept an arbitrary number of args, right?

I guess I'm wondering now if you think a more accurate Docstring edit could be useful. At least personally, it took some time for me to debug why parallel sampling wasn't working.

axch: To be clear, @brianwa84 , @rupei is trying to say that to run a batch of chains, the batching has to be across a Tensor dimension, not across a list.  List structure is always interpreted as part of the latent state of one chain, never as indexing a batch dimension.

axch:
The purpose of this PR has been served by 5710987.

PR #1065, opened 2020-08-19
Add streaming arbitrary central moments to `tfp.experimental.stats`

rupei:
This PR adds an object that facilitates the computation of arbitrary central moments to `tfp.experimental.stats`. All intermediate moments (i.e. lower moments) are required for computation, and hence, the user has the choice to define if they want to receive all moments or simply their moment of choice via the `finalize` method.

### DIFF
--- a/tensorflow_probability/python/experimental/stats/sample_stats.py
+++ b/tensorflow_probability/python/experimental/stats/sample_stats.py
@@ -402,14 +402,16 @@ RunningCentralMomentsState = collections.namedtuple(
 class RunningCentralMoments(object):
   """Holds metadata for and computes running central moments.
 
-  `RunningCentralMoments` will compute an arbitrary central moment in
+  `RunningCentralMoments` will compute arbitrary central moments in
   streaming fashion following the formula proposed by Philippe Pebay
   (2008) [1]. For reference, the formula we refer to is the incremental
   version of arbitrary moments (equation 2.9). Since the algorithm computes
   moments as a function of lower ones, even if not requested, all lower
-  moments will be computed as well. Note, while any arbitrarily high
-  central moment is theoretically supported, `RunningCentralMoments` cannot
-  guarantee numerical stability for all moments.
+  moments will be computed as well. The moments that are actually returned
+  is specified by the `moment` parameter at initialization. Note, while
+  any arbitrarily high central moment is theoretically supported,
+  `RunningCentralMoments` cannot guarantee numerical stability for all
+  moments.
 
   `RunningCentralMoments` objects do not hold state information. That
   information, which includes intermediate calculations, are held in a
@@ -423,14 +425,14 @@ class RunningCentralMoments(object):
         https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2008/086212.pdf
   """
 
-  def __init__(self, shape, max_moment, dtype=tf.float32):
+  def __init__(self, shape, moment, dtype=tf.float32):
     """Instantiates this object.
 
     Args:
       shape: Python `Tuple` or `TensorShape` representing the shape of
         incoming samples.
-      max_moment: Integer that represents the highest central moment of
-        interest.
+      moment: Integer or shallow collection of integers that represent the
+        desired moments to return.
       dtype: Dtype of incoming samples and the resulting statistics.
         By default, the dtype is `tf.float32`. Any integer dtypes will be
         cast to corresponding floats (i.e. `tf.int32` will be cast to
@@ -438,7 +440,14 @@ class RunningCentralMoments(object):
         floating-point division.
     """
     self.shape = shape
-    self.max_moment = max_moment
+    if isinstance(moment, collections.abc.Iterable):
+      # we want to support numpy arrays too, but must convert to a list to not
+      # confuse `tf.nest.map_structure` in `finalize`
+      self.moment = list(moment)
+      self.max_moment = max(self.moment)
+    else:
+      self.moment = moment
+      self.max_moment = moment
     if dtype is tf.int64:
       dtype = tf.float64
     elif dtype.is_integer:
@@ -454,7 +463,7 @@ class RunningCentralMoments(object):
     The `RunningCentralMomentsState` contains a `RunningMeanState` and
     a `Tensor` representing the sum of exponentiated residuals. The sum
     of exponentiated residuals is a `Tensor` of shape
-    (`self.max_moment - 1`, `self.shape`) which contains the sum of
+    (`self.max_moment - 1`, `self.shape`), which contains the sum of
     residuals raised to the nth power, for all `2 <= n <= self.max_moment`.
 
     Returns:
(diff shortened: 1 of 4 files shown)


tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +392,152 @@ def finalize(self, state):
...
+      # having a `(n-1)` denominator.
+      adj_term = (((delta_mean / n) ** p) * (n - 1) *
+                  ((n - 1) ** (p - 1) + (-1) ** p))
+      new_sum_pth_residual = old_res[p - 1] + summation + adj_term
+      new_sum_exponentiated_residuals.append(new_sum_pth_residual)
+
+    return RunningCentralMomentsState(
+        num_samples=n_1 + 1,
+        mean=new_mean,
+        sum_exponentiated_residuals=tf.convert_to_tensor(
```
SiegeLordEx, line 509: tf.stack for clarity?

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +392,152 @@ def finalize(self, state):
...
+    ), new_sample).mean
+    old_res = state.sum_exponentiated_residuals
+    new_sum_exponentiated_residuals = [tf.zeros(self.shape, self.dtype)]
+
+    # the following two nested for loops calculate equation 2.9 in Pebay's
+    # 2008 paper from smallest moment to highest.
+    for p in range(2, self.moment + 1):
+      summation = tf.zeros(self.shape, self.dtype)
+      for k in range(1, p - 1):
+        adjusted_old_res = ((-delta_mean / n) ** k) * old_res[p-k-1]
```
SiegeLordEx, line 496: Add spaces around binary ops.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +392,152 @@ def finalize(self, state):
...
+    Returns:
+      state: `RunningCentralMomentsState` representing a stream of no
+        inputs.
+    """
+    mean_stream = RunningMean(
+        self.shape, self.dtype
+    )
+    return RunningCentralMomentsState(
+        num_samples=0,
+        mean=mean_stream.initialize().mean,
```
SiegeLordEx, line 461: Why not just keep the mean_stream here? You seem to keep reconstructing it.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +392,152 @@ def finalize(self, state):
...
+
+    return RunningCentralMomentsState(
+        num_samples=n_1 + 1,
+        mean=new_mean,
+        sum_exponentiated_residuals=tf.convert_to_tensor(
+            new_sum_exponentiated_residuals
+        )
+    )
+
+  def finalize(self, state, return_all_moments=True):
```
SiegeLordEx, line 514: I don't know if this arg really pulls its weight. The user can just as easily write the `r.finalize(state)[r.moment - 1]` themselves.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +392,152 @@ def finalize(self, state):
...
+
+
+class RunningCentralMoments(object):
+  """Holds metadata for and computes running central moments.
+
+  `RunningCentralMoments` will compute an arbitrary central moment in
+  streaming fashion following the formula proposed by Philippe Pebay
+  (2008) [1]. For reference, the formula we refer to is the incremental
+  version of arbitrary moments (equation 2.9). Since the algorithm computes
+  moments as a function of lower ones, even if not requested, all lower
```
SiegeLordEx, line 409: Maybe we should embrace this fact on the API level, and alter the names of things like `moment` to `max_moment` and just accept that we're returning a bunch of moments. The class name is already in plural form rather than singular, which is already a step in this direction.

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: Ah right. Sorry for continuously missing these, it's a personal preference of mine and my pylint doesn't yell at me when I don't do it

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: Yeah, it's not a critical argument to have -- I just thought it would be a nice convenience. Do you think it's better to remove it for API simplicity?

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: Yup, that's definitely an option. I'll go ahead and adopt that then

tensorflow_probability/python/experimental/stats/sample_stats.py
rupei: Good point. Would seem kind of weird to keep it in the state so I'm going to make it a class attribute.

SiegeLordEx: I mean... there's nothing wrong with composition. In tfp.mcmc we have 'foreign' kernel results in the outer kernel results all the time.

rupei: Ohhh are you referring to the `RunningMeanState` and not the actual `RunningMean` object?

If you did mean the former, I thought that nesting was kind of awkward as the `RunningMeanState` is simply the mean and the number of samples. If we kept the entire state, we'd be duplicating storing the number of samples in two places.

If you mean the latter, I just thought it was weird having an actual object in the kernel results (and not another state object or set of kernel results). I could be wrong and there definitely could be precedent somewhere else in tfp I haven't stumbled upon.

SiegeLordEx: I meant the former. You're right that it's awkward. You could store the `num_samples` inside RMS exclusively (i.e. remove it from RCMS), or replace the mean in RMS state with an empty tuple. I guess what you had here isn't the worst thing either, but it just speaks to the awkwardness in this API. In FunMC I just accept that I'm going to have duplicate state in a few places (with an option to do one of the above things if I really thought it was a performance issue).

rupei: Gotcha. Yeah, I agree just having the state works. I'll simply store the `num_samples` inside RMS.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +392,133 @@ def finalize(self, state):
...
+
+    Returns:
+      state: `RunningCentralMomentsState` with updated calculations.
+    """
+    n_2 = 1
+    n_1 = state.mean_state.num_samples
+    n = tf.cast(n_1 + n_2, dtype=self.dtype)
+    delta_mean = new_sample - state.mean_state.mean
+    new_mean_state = self.mean_stream.update(state.mean_state, new_sample)
+    old_res = state.sum_exponentiated_residuals
```
axch, line 481: Explain somewhere what the type of `sum_exponentiated_residuals` is, and what those values mean.

tensorflow_probability/python/experimental/stats/sample_stats.py
```diff
@@ -389,3 +392,133 @@ def finalize(self, state):
...
+    new_sum_exponentiated_residuals = [tf.zeros(self.shape, self.dtype)]
```
axch, line 482: These zeros don't seem to ever be adjusted.  Is there a reason to store them in the output, since they are always zero?

rupei: What ends up happening is the `new_sum_exponentiated_residuals` gets converted to a tensor with a leading dimension that indexes into the (i+1)th central moment (i.e. the 0th index is the first central moment). Since the first central moment is always 0 I prepend this list with a corresponding `tf.zeros` to simplify the `Tensor` of moments that's returned.

PR #1070, opened 2020-08-30
add argument `line_search_kwargs` to BFGS

### LINKED ISSUE #1043: BFGS user should be able to pass parameters to the line search
This should just be a matter of accepting a dict named `line_search_kwargs` or such and passing it through to the actual line search.

This was a feature request that came out of discussing #1034.

rushabh-v:
Fixes: https://github.com/tensorflow/probability/issues/1043

### DIFF
--- a/tensorflow_probability/python/optimizer/bfgs.py
+++ b/tensorflow_probability/python/optimizer/bfgs.py
@@ -81,7 +81,8 @@ def minimize(value_and_gradients_function,
              stopping_condition=None,
              validate_args=True,
              max_line_search_iterations=50,
-             name=None):
+             name=None,
+             line_search_kwargs={}):
   """Applies the BFGS algorithm to minimize a differentiable function.
 
   Performs unconstrained minimization of a differentiable function using the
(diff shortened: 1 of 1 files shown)


tensorflow_probability/python/optimizer/bfgs.py
```diff
@@ -81,7 +81,8 @@ def minimize(value_and_gradients_function,
              stopping_condition=None,
              validate_args=True,
              max_line_search_iterations=50,
-             name=None):
+             name=None,
+             line_search_kwargs={}):
```
axch, line 85: We have a pretty strong convention in TFP of letting the `name` argument be last.

Also please use `None` as a default value and explicitly set it to `{}` if `None` in the body, because `{}` is not safe as a default argument (see lint error on Travis).

tensorflow_probability/python/optimizer/bfgs.py
```diff
@@ -276,10 +277,9 @@ def _body(state):
           state, inverse_hessian_estimate=actual_inv_hessian)
 
       next_state = bfgs_utils.line_search_step(
-          current_state,
-          value_and_gradients_function, actual_search_direction,
+          current_state, value_and_gradients_function, actual_search_direction,
           tolerance, f_relative_tolerance, x_tolerance, stopping_condition,
-          max_line_search_iterations)
+          max_line_search_iterations, **line_search_kwargs)
```
axch, line 282: The purpose is for the user to be able to control the parameters of the actual line search algorithm, namely `linesearch.hager_zhang`.  All of the parameters that `line_search_step` itself consumes are already being passed in explicitly.  Please pass the dictionary through to that depth, and add a unit test that you can, indeed, affect the behavior of the algorithm by controlling at least one of those parameters (e.g., `initial_step_size`).

axch (changes requested):

brianwa84:
@rushabh-v will you be following up on the test @axch requested?

rushabh-v:
I am sorry, I tried at that time but couldn't figure out a way to test that (I could not figure out changing which argument would make what change in the output, because I don't understand the math behind it). And currently, I am really low on time. Should I close this PR?

hasan-malik:
FYI — I've opened #2023 that picks this up and addresses @axch's feedback (the argument ordering and the unit test).

Happy to defer if you'd prefer to revive this one.

PR #1079, opened 2020-09-11
add example of conditional density estimation to docs

Padarn:
I was reading the docs here: https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/AutoregressiveNetwork and couldn't immediately see how to use `conditional`. So after figuring it out I decided it would be useful to submit an example.

Simply adds an example of conditional density estimation to the doc string.

### DIFF
--- a/tensorflow_probability/python/bijectors/masked_autoregressive.py
+++ b/tensorflow_probability/python/bijectors/masked_autoregressive.py
@@ -684,20 +684,21 @@ class AutoregressiveNetwork(tf.keras.layers.Layer):
 
   ```python
   # Generate data as the mixture of two distributions.
-  n = 2000
+  n = 10000
   c = np.r_[
     np.zeros(n//2),
     np.ones(n//2)
   ]
+  mean_0, mean_1 = 0, 5
   x = np.r_[
-    np.random.randn(n//2).astype(dtype=np.float32) * 2,
-    np.random.randn(n//2).astype(dtype=np.float32) + 5
+    np.random.randn(n//2).astype(dtype=np.float32) + mean_0,
+    np.random.randn(n//2).astype(dtype=np.float32) + mean_1
   ]
 
   # Density estimation with MADE.
   made = tfb.AutoregressiveNetwork(
     params=2, 
-    hidden_units=[10, 10],
+    hidden_units=[2, 2],
     event_shape=(1,),
     conditional=True, 
     conditional_event_shape=(1,)
@@ -714,19 +715,21 @@ class AutoregressiveNetwork(tf.keras.layers.Layer):
   model = tfk.Model([x_, c_], log_prob_)
 
   model.compile(optimizer=tf.optimizers.Adam(),
-                 loss=lambda _, log_prob: -log_prob)
+                loss=lambda _, log_prob: -log_prob)
 
   batch_size = 25
   model.fit(x=[x, c],
             y=np.zeros((n, 0), dtype=np.float32),
             batch_size=batch_size,
-            epochs=1,
-            steps_per_epoch=1,  # Usually `n // batch_size`.
+            epochs=2,
+            steps_per_epoch=n // batch_size,  
             shuffle=True,
             verbose=True)
 
   # Use the fitted distribution to sample condition on c = 1
-  distribution.sample((3,), bijector_kwargs={'conditional_input': np.ones((3, 1))})
+  n_samples = 1000
+  cond = 1
+  samples = distribution.sample((n_samples,), bijector_kwargs={'conditional_input': cond * np.ones((n_samples, 1))})
   ```
 
   #### Examples: Handling Rank-2+ Tensors
--- a/tensorflow_probability/python/bijectors/masked_autoregressive.py
+++ b/tensorflow_probability/python/bijectors/masked_autoregressive.py
@@ -686,16 +686,22 @@ class AutoregressiveNetwork(tf.keras.layers.Layer):
   # Generate data as the mixture of two distributions.
   n = 2000
   c = np.r_[
-  np.zeros(n//2),
-  np.ones(n//2)
+		np.zeros(n//2),
+		np.ones(n//2)
   ]
   x = np.r_[
-  np.random.randn(n//2).astype(dtype=np.float32) * 2,
-  np.random.randn(n//2).astype(dtype=np.float32) + 5
+		np.random.randn(n//2).astype(dtype=np.float32) * 2,
+		np.random.randn(n//2).astype(dtype=np.float32) + 5
   ]
 
   # Density estimation with MADE.
-  made = tfb.AutoregressiveNetwork(params=2, hidden_units=[10, 10])
+  made = tfb.AutoregressiveNetwork(
+		params=2, 
+		hidden_units=[10, 10],
+		event_shape=(1,),
+    conditional=True, 
+		conditional_event_shape=(1,)
+	)
 
   distribution = tfd.TransformedDistribution(
     distribution=tfd.Sample(tfd.Normal(loc=0., scale=1.), sample_shape=[1]),
@@ -707,8 +713,10 @@ class AutoregressiveNetwork(tf.keras.layers.Layer):
   log_prob_ = distribution.log_prob(x_, bijector_kwargs={'conditional_input': c_})
   model = tfk.Model([x_, c_], log_prob_)
 
-  model.compile(optimizer=tf.optimizers.Adam(),
-              loss=lambda _, log_prob: -log_prob)
+  model.compile(
+		optimizer=tf.optimizers.Adam(),
+ 	  loss=lambda _, log_prob: -log_prob
+	)
 
   batch_size = 25
   model.fit(x=[x, c],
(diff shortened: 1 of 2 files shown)


tensorflow_probability/python/bijectors/masked_autoregressive.py
```diff
@@ -676,6 +679,50 @@ class AutoregressiveNetwork(tf.keras.layers.Layer):
...
   ```
 
+  The `conditional` argument can be used to instead build a conditional density.  
+  estimator. To do this the conditioning variable must be passed as a `kwarg`:
+
+  ```python
+  # Generate data as the mixture of two distributions.
+  n = 2000
+  c = np.r_[
+  np.zeros(n//2),
```
SiegeLordEx, line 689: Could you indent the lines inside the brackets?

tensorflow_probability/python/bijectors/masked_autoregressive.py
```diff
@@ -676,6 +679,50 @@ class AutoregressiveNetwork(tf.keras.layers.Layer):
...
+  np.ones(n//2)
+  ]
+  x = np.r_[
+  np.random.randn(n//2).astype(dtype=np.float32) * 2,
```
SiegeLordEx, line 693: Here too.

tensorflow_probability/python/bijectors/masked_autoregressive.py
```diff
@@ -676,6 +679,50 @@ class AutoregressiveNetwork(tf.keras.layers.Layer):
...
+  np.random.randn(n//2).astype(dtype=np.float32) + 5
+  ]
+
+  # Density estimation with MADE.
+  made = tfb.AutoregressiveNetwork(params=2, hidden_units=[10, 10])
```
SiegeLordEx, line 698: It's not actually conditional?

tensorflow_probability/python/bijectors/masked_autoregressive.py
```diff
@@ -676,6 +679,50 @@ class AutoregressiveNetwork(tf.keras.layers.Layer):
...
+    bijector=tfb.MaskedAutoregressiveFlow(made))
+
+  # Construct and fit model.
+  x_ = tfkl.Input(shape=(1,), dtype=tf.float32)
+  c_ = tfkl.Input(shape=(1,), dtype=tf.float32)
+  log_prob_ = distribution.log_prob(x_, bijector_kwargs={'conditional_input': c_})
+  model = tfk.Model([x_, c_], log_prob_)
+
+  model.compile(optimizer=tf.optimizers.Adam(),
+              loss=lambda _, log_prob: -log_prob)
```
SiegeLordEx, line 711: Two extra spaces too.

tensorflow_probability/python/bijectors/masked_autoregressive.py
```diff
@@ -676,6 +679,50 @@ class AutoregressiveNetwork(tf.keras.layers.Layer):
...
+  x = np.r_[
+  np.random.randn(n//2).astype(dtype=np.float32) * 2,
+  np.random.randn(n//2).astype(dtype=np.float32) + 5
+  ]
+
+  # Density estimation with MADE.
+  made = tfb.AutoregressiveNetwork(params=2, hidden_units=[10, 10])
+
+  distribution = tfd.TransformedDistribution(
+    distribution=tfd.Sample(tfd.Normal(loc=0., scale=1.), sample_shape=[1]),
```
SiegeLordEx, line 701: I guess it's okay for a test, but MAF doesn't really do anything useful when it works on length=1 vectors. It'll just learn an unconditional shift + scale.

tensorflow_probability/python/bijectors/masked_autoregressive.py
```diff
@@ -676,6 +679,50 @@ class AutoregressiveNetwork(tf.keras.layers.Layer):
...
+  model.fit(x=[x, c],
+          y=np.zeros((n, 0), dtype=np.float32),
+          batch_size=batch_size,
+          epochs=1,
+          steps_per_epoch=1,  # Usually `n // batch_size`.
+          shuffle=True,
+          verbose=True)
+
+  # Use the fitted distribution to sample condition on c = 1
+  distribution.sample((3,), bijector_kwargs={'conditional_input': np.ones((3, 1))})
```
SiegeLordEx, line 723: Maybe take a few more samples, and assert that their mean is close to 5?

SiegeLordEx, line 723: There is an important point to mention here: conditional_input must broadcast with shapes like: sample_shape + conditional_event_shape.

SiegeLordEx (changes requested):
This looks good. I'm a little concerned about this only existing in the docstring though. Would you mind adding a test case in masked_autoregressive_test.py with (roughly) the same snippet?

Padarn:
Thanks for the comments @SiegeLordEx - It was late at night and I made some mistakes in adapting from my local example. Will add a test case and fix the formatting.

tensorflow_probability/python/bijectors/masked_autoregressive.py
Padarn: Agreed. In this case, I'm just trying to make obvious how the feature works. Perhaps I could add a more complete example to a tutorial?

tensorflow_probability/python/bijectors/masked_autoregressive.py
Padarn: fixed

Padarn:
@SiegeLordEx I've updated it with a test case and fixed some of the formattings. 

I haven't added any comment about the batch dimension - I agree this is important but perhaps should be added earlier in the docstring rather within the example. I'm happy to work on this, but I'd need your advice on an example of a docstring in the codebase which follows the canonical format you'd like to follow so I can match the style.

I had to fiddle around with the sample size and training epochs to make sure the unit test would pass. I set an `np.seed` to make sure this works repeatedly. Let me know if you think there is a better way to do this.

tensorflow_probability/python/bijectors/masked_autoregressive.py
SiegeLordEx: Let's keep it as is, I was just thinking out loud.

tensorflow_probability/python/bijectors/masked_autoregressive.py
```diff
@@ -676,6 +679,62 @@ class AutoregressiveNetwork(tf.keras.layers.Layer):
...
+            epochs=2,
+            steps_per_epoch=n // batch_size,
+            shuffle=True,
+            verbose=True)
+
+  # Use the fitted distribution to sample condition on c = 1
+  n_samples = 1000
+  cond = 1
+  samples = distribution.sample(
+    (n_samples,),
```
SiegeLordEx, line 734: The lint is complaining about the trailing space here.

tensorflow_probability/python/bijectors/masked_autoregressive_test.py
```diff
@@ -729,6 +729,60 @@ def test_doc_string(self):
...
+              verbose=True)
+
+    # Use the fitted distribution to sample condition on c = 1
+    n_samples = 1000
+    cond = 1
+    samples = distribution.sample(
+      (n_samples,),
+      bijector_kwargs={'conditional_input': cond * np.ones((n_samples, 1))})
+    # Assert mean is close to conditional mean
+    self.assertAllClose(tf.reduce_mean(samples).numpy(), mean_1, atol=0.2)
```
SiegeLordEx, line 784: You should be able to drop the `.numpy()` here, as `assertAllClose` knows how to handle TensorFlow values.

tensorflow_probability/python/bijectors/masked_autoregressive_test.py
```diff
@@ -729,6 +729,60 @@ def test_doc_string(self):
     self.assertAllEqual(
         (3,), distribution.log_prob(np.ones((3, 2), dtype=np.float32)).shape)
 
+  def test_doc_string_2(self):
+    np.random.seed(123)
```
SiegeLordEx, line 733: No need to set the seed, it's set somewhere deep in TensorFlow's base testcase.

SiegeLordEx (changes requested):
Thanks for the changes. I think this is basically good to go, just a few minor changes. Let's try to get the lint  passing on Travis and maybe the test itself (if Travis doesn't time out). If not, I'll just run our internal tests.

tensorflow_probability/python/bijectors/masked_autoregressive_test.py
Padarn: Ah yes, I wondered about this but didn't want to assume. Thanks

Padarn:
Thanks - I've gotten the lint passing, but had to also resolve some linting around python 3 `super()`. Do we support python 2 still? If so maybe this is going to cause some issues.

SiegeLordEx (approved):

SiegeLordEx:
That looks great. I'll run this through our internal review system, and we'll get this merged shortly. Thanks again!

Padarn:
Thanks! Happy to contribute, hope to see you in another MR soon :-)

Issue #372, opened 2019-04-21
Feature request: Negative binomial with mean and dispersion parameterization


gokceneraslan:
For many people working with count data (especially in ecology and genomics), it is very common to parameterize negative binomial with mean (`total_count*exp(logits)`) and dispersion (`1/total_count`). Would it be possible to add this reparametrization of NB to tfp?

brianwa84:
It seems straightforward to do the reparameterization since one can infer
the existing parameters from those you identify. Are there any numerical
stability benefits or downsides to this parameterization? e.g. It seems
like you might lose precision on total_count if it were represented by it's
inverse between 0 and 1.

https://github.com/tensorflow/probability/blob/c6c14ff4e684ebc2613833d71cb476ba5c27e2a8/tensorflow_probability/python/distributions/negative_binomial.py#L148
is where we do sampling.

gokceneraslan:
There is no consensus over how dispersion is defined actually, some people use `number of failures` directly, some people use the inverse. Former approaches `+inf` in Poisson regime, so I'm not entirely sure if this is as bad as it approaching zero.

R uses both parameterization in following way: `rnbinom(n, size, prob, mu)` where size is the target for number of successful trials, or dispersion parameter (the shape parameter of the gamma mixing distribution) and mu is the mean. Either prob or mu has to be specified by the user, but not both. Details are [here](https://stat.ethz.ch/R-manual/R-patched/library/stats/html/NegBinomial.html).

I think having both parameterizations would be very convenient in tfp too.

brianwa84:
I agree, if it's a fairly common parameterization, we should support it. It seems like R does support this, but scipy does not. 

If you'd like to send a PR to add either alternative constructor args, I think we'd be open to it (make sure to raise if >1 parameterization is actually used). It would be good if you could could put a reference or two [in the docstring](https://github.com/tensorflow/probability/blob/master/STYLE_GUIDE.md#tensorflow-probability-style) showing how the new parameterization is used in practice in a couple compbio/eco papers? I found this one: https://stackoverflow.com/questions/40846992/alternative-parametrization-of-the-negative-binomial-in-scipy

We try to be sparing with new arg names, but dispersion seems fairly commonly used for this and is somewhat analagous to `concentration`, which we use widely. @jvdillon thoughts?

jvdillon:
We've discussed mean parameterization of distributions. Unfortunately there's a few usability questions which need discussion. Can we discuss this as a team and report back?

[...]

PR #1082, opened 2020-09-16
Add NegativeBinomial mean and dispersion parametrization

adriang133:
An attempt at #372.

Not really sure how to handle dtypes in this case, it seems like a waste to call `dtype_util.common_dtype` twice, but I couldn't find a good way to reconcile the `DeferredTensor` dtypes otherwise.


brianwa84 (approved):
I'm going to pull this in, but rename the function to `experimental_from_mean_dispersion` while doing so, to give us more time for a discussion about whether we want general or specific naming for these. Thanks for kicking this off!

PR #1084, opened 2020-09-17
Fix: `index_points` coming from `self._get_index_points(index_points)`

### LINKED ISSUE #1083: tfd.VariationalGaussianProcess class: _batch_shape and _batch_shape_tensor with index_points in input
While extending the class, I just noticed the methods:

```python
  def _batch_shape(self, index_points=None):
    index_points = self._get_index_points(index_points)
    batch_shapes = [
        self.kernel.batch_shape,
        self.variational_inducing_observations_loc.shape[:-1]]
    if self.index_points is not None:
      batch_shapes.append(
          self.index_points.shape[:-(self.kernel.feature_ndims + 1)])
    return functools.reduce(tf.broadcast_static_shape, batch_shapes)

  def _batch_shape_tensor(self, index_points=None):
    index_points = self._get_index_points(index_points)
    batch_shape_tensors = [
        self.kernel.batch_shape_tensor(),
        tf.shape(self.variational_inducing_observations_loc)[:-1]]
    if self.index_points is not None:
      batch_shape_tensors.append(
          tf.shape(self.index_points)[:-(self.kernel.feature_ndims + 1)])
    return functools.reduce(tf.broadcast_dynamic_shape, batch_shape_tensors)
```

which are able to take `index_points` as input. However, the local variable `index_points` is never used, instead, `self.index_points` is used to understand the batch_shape (which is wrong when the index points are different than the initialisation). 

I believe this to be not intended, and the fix to simply be replacing `self.index_points` with `index_points` in both methods.
Let me know if that was instead intended.

fdtomasi:
This fixes #1083 .

`index_points = self._get_index_points(index_points)` were not used in the VariationalGaussianProcess class, in both `_batch_shape` and `_batch_shape_tensor` methods. Now the index points are correctly used to understand the batch_shape (without relying on `self.index_points`).

### DIFF
--- a/tensorflow_probability/python/distributions/variational_gaussian_process.py
+++ b/tensorflow_probability/python/distributions/variational_gaussian_process.py
@@ -934,9 +934,9 @@ class VariationalGaussianProcess(gaussian_process.GaussianProcess):
     batch_shapes = [
         self.kernel.batch_shape,
         self.variational_inducing_observations_loc.shape[:-1]]
-    if self.index_points is not None:
+    if index_points is not None:
       batch_shapes.append(
-          self.index_points.shape[:-(self.kernel.feature_ndims + 1)])
+          index_points.shape[:-(self.kernel.feature_ndims + 1)])
     return functools.reduce(tf.broadcast_static_shape, batch_shapes)
 
   def _batch_shape_tensor(self, index_points=None):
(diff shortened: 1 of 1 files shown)


csuter (approved):
Thanks!

csuter:
Thanks! I'll pull this in to our internal VCS and it'll pop back out here in a bit.

Issue #1080, opened 2020-09-16
Difference returning dtype of tfp.layers.IndependentNormal.params_size on Windows and Google Colab (Linux )


elda27:
I found the following similarity example from the document of TensorFlow probability.

```
model = tf.keras.Sequential([
    tfp.layers.DenseFlipout(800, activation=tf.nn.tanh),
    tfp.layers.DenseFlipout(tfp.layers.IndependentNormal.params_size(1)), # This line raise InvalidArgumentError (Original exception throw from tf.concat function.)
    # tfp.layers.DenseFlipout(np.int32(tfp.layers.IndependentNormal.params_size(1))), # This line is working both environments.
    tfp.layers.IndependentNormal(1)
])
```

The example is working on Windows but it is NOT working on Google Colab (Linux) by the eager execution.
Is this problem well known or unknown?
Or my usage is wrong?
I seem correct behavior is the Windows environment because of the "int" type will interpret "int32" on the numpy. 

I made a small example, please verify the usage and the problem.
https://gist.github.com/elda27/54989bb20383ae29dc981bf36bb29a73

csuter:
It seems this arises because np.prod is upcasting int types from int32 to int64. Our params_size code tries to evaluate the size by getting the shape statically, and then doing an np.prod on it, at which point it becomes of type np.int64. I think we can patch this in TFP Layers by doing an explicit `.astype(np.int32)` after that reduce_prod.

elda27:
I agree with your suggestion.

I check the dtype of tf.reduce_prod and np.prod.
The first one is int32 but the second one is int64 on Linux (Both ones are int32).

And I seem to be able to set dtype by an argument at np.prod.
I seem the change is only the following lines.

https://github.com/tensorflow/probability/blob/a9cff2537033b12c1400484beb7bad24a1c38fa6/tensorflow_probability/python/layers/distribution_layer.py#L95

Can I contribute to the problem?

elda27:
@csuter Can you review my PR?
I fixed returning unexpected dtype on the IndependentNormalDistribution.params_size and IndependentLogisticDistribution.params_size.
And also I modified a unit test.

PR #1090, opened 2020-09-27
Fix dtype error independent normal distribution

elda27:
Please check my pull request about #1080 .
I solved this problem and modified a test.

I always cast Python integer to numpy.int32 because the value of numpy.int32 will upcast with computing Python integer on the Linux.

### DIFF
--- a/tensorflow_probability/python/layers/distribution_layer.py
+++ b/tensorflow_probability/python/layers/distribution_layer.py
@@ -896,7 +896,7 @@ class IndependentLogistic(DistributionLambda):
     with tf.name_scope(name or 'IndependentLogistic_params_size'):
       event_shape = tf.convert_to_tensor(
           event_shape, name='event_shape', dtype_hint=tf.int32)
-      return 2 * _event_size(
+      return np.int32(2) * _event_size(
           event_shape, name=name or 'IndependentLogistic_params_size')
 
   def get_config(self):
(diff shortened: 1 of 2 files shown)


elda27:
@googlebot I signed it!

tensorflow_probability/python/layers/distribution_layer.py
```diff
@@ -896,7 +896,7 @@ def params_size(event_shape=(), name=None):
     with tf.name_scope(name or 'IndependentLogistic_params_size'):
       event_shape = tf.convert_to_tensor(
           event_shape, name='event_shape', dtype_hint=tf.int32)
-      return 2 * _event_size(
+      return np.int32(2) * _event_size(
```
ColCarroll, line 899: Will tests pass without this cast (and the one on line 1014)?

elda27: This line will fail in my environment. (I added the line)
https://github.com/tensorflow/probability/blob/58a5fcba16d89a9a17f1a83d97a2732808dd2733/tensorflow_probability/python/layers/distribution_layer_test.py#L829

Please see this issue.
https://github.com/numpy/numpy/issues/5297

I understood the 0-dim array or scalar of np.int32 will upcast np.int64 with computing python integer.
(Above issue is discussed np.float32 and np.float64 but the essence is the same.)
So I need the always cast the np.int32(2).

elda27: Supplements of mine comment. There is the result dtype on my environment (on Python 3.6.5 and Ubuntu 20.04). 
Please try these lines on your environment too.

In [1]: import numpy as np
In [2]: np.__version__
Out[2]: '1.19.2'

In [3]: print((np.array([10, ], dtype=np.int32) * 4).dtype) # NumPy 1-d array
int32

In [4]: print((np.int32(10) * 4).dtype) # NumPy scalar
int64

In [5]: print((np.array(10, dtype=np.int32) * 4).dtype) # NumPy 0-d array
int64

ColCarroll (approved):

PR #1104, opened 2020-10-05
add weights support for histogram

mayou36:
Add weights support for the `stats.histogram` function. This is a crucial feature for zfit, a likelihood model fitting library for High Energy Physics, as there are always weights involved.

Since weights are supported already in the internally used `count_integer` and basically just needs to be passed through.

The requirement and behavior of the `weights` is equivalent to `count_integers`.

TODO:

- [ ] Approve changes by TFP
- [ ] Explicitly fail if dynamic shapes don't match?
- [x] Add unittest
- [x] Update Docstring

### DIFF
--- a/tensorflow_probability/python/stats/quantiles.py
+++ b/tensorflow_probability/python/stats/quantiles.py
@@ -399,7 +399,7 @@ def histogram(x,
           raise ValueError('Shape of `x` and `weights` have to coincide.  '
                            'Found: x.shape={} weights.shape={}'
                            ''.format(x.shape, weights.shape))
-        weights = _move_dims_to_flat_end(weights, axis, x_ndims,
+        weights = _move_dims_to_flat_end(weights, axis, weights_ndims,
                                          right_end=False)
 
     # bins.shape = x.shape = [n_samples] + E,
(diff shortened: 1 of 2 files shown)


mayou36:
@googlebot I signed it!

brianwa84 (approved):

brianwa84:
@jvdillon asks:
I wonder if this added capability is really necessary. Can the user just rescale the histogram after it's computed?

mayou36:
No, you can't. It's not that easy because you do not have the weights _per bin_ but the _weight per event_, so you have to know which weight goes to which bin. You would need to to a `find_bin` for the actual values, use them to index the weights and the values... basically do everything manually.

It's just consistent as well with the `count_integer` that is used (which takes a `weights` argument. Basically, `weights` here is "just passed through".

mayou36:
Any news on this, can this be approved & merged @jvdillon ?

brianwa84:
It got merged, but I guess the PR didn't get updated: https://github.com/tensorflow/probability/commit/e359af405d620a02e59422513eb0f30becfcff08

Should be in tfp-nightly and in r0.12.0rc1 !

PR #1123, opened 2020-10-08
Introduce Ascending bijector equivalent to Invert(Ordered())

### LINKED ISSUE #765: are `tfb.Ordered.forward` and `tfb.Ordered.inverse` the wrong way round?
Hi,

Please forgive me if I am being a bit slow with this.

The current behaviour is well documented in the library, but given the behaviour of other bijectors (e.g. `tfb.Exp().forward(x)` calculates `exp(x)`) I really expected `tfb.Ordered().forward(x)` to "order" x - but that is not the case. Is there a way of thinking that I am missing or is this inconsistent?

Currently (I think) this would mean to do e.g. MCMC with parameters that were constrained to be ordered we would need to use:

``` python
tfp.mcmc.TransformedTransitionKernel(..., bijector=[tfb.Invert(tfb.Ordered())])
```

Which is no big deal but feels odd to me...

This would be needed in https://github.com/tensorflow/probability/pull/753.

Thanks

st--:
Resolves #765

### DIFF
--- a/tensorflow_probability/python/bijectors/ascending.py
+++ b/tensorflow_probability/python/bijectors/ascending.py
@@ -86,6 +86,6 @@ class Ascending(bijector.Bijector):
   def _assertions(self, t):
     if not self.validate_args:
       return []
-    return [assert_util.assert_positive(
-        t[..., 1:] - t[..., :-1],
+    return [assert_util.assert_greater(
+        t[..., 1:], t[..., :-1],
         message="Inverse transformation input must be strictly increasing.")]
--- a/tensorflow_probability/python/bijectors/ascending.py
+++ b/tensorflow_probability/python/bijectors/ascending.py
@@ -49,7 +49,7 @@ class Ascending(bijector.Bijector):
   def __init__(self, validate_args=False, name="ascending"):
     parameters = dict(locals())
     with tf.name_scope(name) as name:
-      super().__init__(
+      super(Ascending, self).__init__(
           forward_min_event_ndims=1,
           validate_args=validate_args,
           parameters=parameters,
--- a/tensorflow_probability/python/bijectors/ascending.py
+++ b/tensorflow_probability/python/bijectors/ascending.py
@@ -12,11 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ============================================================================
-"""Ordered bijector."""
-
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
+"""Ascending bijector."""
 
 import tensorflow.compat.v2 as tf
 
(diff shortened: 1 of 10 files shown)


tensorflow_probability/python/bijectors/ascending.py
```diff
@@ -0,0 +1,91 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ============================================================================
+"""Ascending bijector."""
+
+import tensorflow.compat.v2 as tf
```
st--, line 17: What's the intent behind all the backwards-compatibility shims, if tfp 0.11 doesn't even support TF < 2.3? And TF 2 no longer supports python 2.7 either as far as I can see.

tensorflow_probability/python/bijectors/ascending.py
```diff
@@ -0,0 +1,91 @@
...
+from tensorflow_probability.python.bijectors import bijector
+from tensorflow_probability.python.internal import assert_util
+
+
+__all__ = [
+    "Ascending",
+]
+
+
+class Ascending(bijector.Bijector):
```
st--, line 28: Happy to rename it if the tfp maintainers have a preference for a different name.

tensorflow_probability/python/bijectors/ascending_test.py
```diff
@@ -0,0 +1,100 @@
...
+
+  def setUp(self):
+    self._rng = np.random.RandomState(42)
+
+  def testName(self):
+    ascending = tfb.Ascending()
+    self.assertStartsWith(ascending.name, "ascending")
+
+  def testBijectorVector(self):
+    ordered = tfb.Invert(tfb.Ascending())
```
st--, line 40: I was admittedly a bit lazy to carefully go through the test and swap all x and y and directions; as Ascending() effectively implements Invert(Ordered()), then Invert(Ascending()) is equivalent to Ordered() and I simply copied the tests from ordered_test.py

tensorflow_probability/python/bijectors/ordered.py
```diff
@@ -30,13 +30,12 @@
 
 
 class Ordered(bijector.Bijector):
-  """Maps a vector of increasing elements to an unconstrained vector.
+  """Deprecated. Use bijectors.Invert(bijectors.Ascending()) instead.
```
st--, line 33: Should the class raise (well, warn with) a DeprecationWarning? Though I note no-where else in the tfp code is that made use of.

jeffpollock9: I think the best thing to do here is to use `tensorflow.python.util.deprecation` see e.g. https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/bijectors/affine_scalar.py#L55-L59

tensorflow_probability/python/bijectors/ascending.py
```diff
@@ -0,0 +1,91 @@
...
+    return tf.reduce_sum(x[..., 1:], axis=-1)
+
+  def _inverse_log_det_jacobian(self, y):
+    with tf.control_dependencies(self._assertions(y)):
+      return -tf.reduce_sum(tf.math.log(y[..., 1:] - y[..., :-1]), axis=-1)
+
+  def _assertions(self, t):
+    if not self.validate_args:
+      return []
+    return [assert_util.assert_positive(
```
jeffpollock9, line 89: I think it'd read better to use `assert_util.assert_greater` here, I'd guess slightly more efficient too.

st--: I've just copied from `ordered.py` but happy to update this. Should I update ordered.py accordingly as well ?

st--: though I'm surprised to hear that `assert_greater` should be more efficient than `assert_positive` (which is simply implemented as `assert_less(0, ...)`)...

tensorflow_probability/python/bijectors/ascending.py
```diff
@@ -0,0 +1,91 @@
...
+    with tf.name_scope(name) as name:
+      super().__init__(
+          forward_min_event_ndims=1,
+          validate_args=validate_args,
+          parameters=parameters,
+          name=name)
+
+  def _forward(self, x):
+    y0 = x[..., :1]
+    yk = tf.exp(x[..., 1:])
```
st--, line 60: This means that practically the domain isn't [-inf, inf] but actually just [-709, 709] in float64, and less in float32... larger numbers overfloat the exp() - should this be reflected in the docstring ?

tensorflow_probability/python/bijectors/ascending.py
jeffpollock9: I think the advantage of using `assert_greater` (or `assert_less` if we swap the arguments) is that we don't need to do the subtraction `t[..., 1:] - t[..., :-1]` and then a comparison -> we just directly do the comparison.

So the code could just be something like:

``` python
assert_util.assert_greater(t[..., 1:], t[..., :-1], ...
```

Although I haven't dug into the implementations.

st--: ahh! that makes a lot of sense of course, how did I not think of that!

st--:
@googlebot I signed it!

st--:
Force-pushed to update e-mail address to satisfy CLA checker.

tensorflow_probability/python/bijectors/ascending.py
brianwa84: Just legacy, we only dropped 2.7 support internally at Google several weeks back.

tensorflow_probability/python/bijectors/ordered.py
```diff
@@ -22,21 +22,24 @@
...
+from tensorflow.python.util import deprecation  # pylint: disable=g-direct-tensorflow-import
 
 
 __all__ = [
     "Ordered",
 ]
 
 
+@deprecation.deprecated(
+      '2020-10-09',
```
brianwa84, line 34: can you change this to +3mo from today?

brianwa84:
Looks great, I'll pull it in once we get the deprecation date updated

tensorflow_probability/python/bijectors/ascending.py
st--: So this is going to get cleaned up going forward ?

csuter: i think this compat shim is still needed for lingering TF1 users internal to google. we probably could clean it up on our code export rewrites.

csuter: (That said, let's leave it here for now)

st--:
I believe all tests are passing now; the build failure seems to be a single shard (https://travis-ci.org/github/tensorflow/probability/jobs/734338414) that timed out - is there anything else I need to do on this PR?

brianwa84 (approved):
I'll get this pulled in.

PR #1140, opened 2020-10-26
add gev distribution and test

blacksde:
Resolve issue #874

### DIFF
--- a/tensorflow_probability/python/distributions/gev.py
+++ b/tensorflow_probability/python/distributions/gev.py
@@ -23,14 +23,15 @@ import numpy as np
 import tensorflow.compat.v2 as tf
 
 from tensorflow_probability.python.bijectors import gev_cdf as gev_cdf_bijector
-from tensorflow_probability.python.bijectors import identity as identity_bijector
 from tensorflow_probability.python.bijectors import invert as invert_bijector
+from tensorflow_probability.python.bijectors import softplus as softplus_bijector
 from tensorflow_probability.python.distributions import transformed_distribution
 from tensorflow_probability.python.distributions import uniform
 from tensorflow_probability.python.internal import distribution_util
 from tensorflow_probability.python.internal import dtype_util
+from tensorflow_probability.python.internal import parameter_properties
 from tensorflow_probability.python.internal import tensor_util
-
+from tensorflow_probability.python.internal import prefer_static as ps
 
 class GeneralizedExtremeValue(transformed_distribution.TransformedDistribution):
   """The scalar GeneralizedExtremeValue distribution
@@ -42,7 +43,7 @@ class GeneralizedExtremeValue(transformed_distribution.TransformedDistribution):
 
   ```none
   pdf(x; loc, scale, conc) = t(x; loc, scale, conc) ** (1 + conc) * exp(
-  - t(x; loc, scale, conc) ) / scale
+  -t(x; loc, scale, conc) ) / scale
   where t(x) =
     * (1 + conc * (x - loc) / scale) ) ** (-1 / conc) when conc != 0;
     * exp(-(x - loc) / scale) when conc = 0.
@@ -77,7 +78,7 @@ class GeneralizedExtremeValue(transformed_distribution.TransformedDistribution):
   dist.cdf(1.)
 
   # Define a batch of two scalar valued generalized extreme values.
-  # The first has mean 1 and scale 11, the second 2 and 22.
+  # The first has loc 1 and scale 11, the second 2 and 22.
   dist = tfd.GeneralizedExtremeValue(loc=[1, 2.], scale=[11, 22.])
 
   # Evaluate the pdf of the first distribution on 0, and the second on 1.5,
@@ -92,7 +93,7 @@ class GeneralizedExtremeValue(transformed_distribution.TransformedDistribution):
 
   ```python
   # Define a batch of two scalar valued Logistics.
-  # Both have mean 1, but different scales.
+  # Both have location 1, but different scales.
   dist = tfd.GeneralizedExtremeValue(loc=1., scale=1, concentration=[0, 0.9])
 
   # Evaluate the pdf of both distributions on the same point, 3.0,
@@ -116,7 +117,7 @@ class GeneralizedExtremeValue(transformed_distribution.TransformedDistribution):
     that supports broadcasting (e.g. `loc + scale` + `concentration` is valid).
 
     Args:
-      loc: Floating point tensor, the means of the distribution(s).
+      loc: Floating point tensor, the location parameter of the distribution(s).
       scale: Floating point tensor, the scales of the distribution(s).
         scale must contain only positive values.
       concentration: Floating point tensor, the concentration of
@@ -172,15 +173,16 @@ class GeneralizedExtremeValue(transformed_distribution.TransformedDistribution):
           parameters=parameters,
           name=name)
 
-  @staticmethod
-  def _param_shapes(sample_shape):
-    return dict(
-        zip(('loc', 'scale', 'concentration'),
-            ([tf.convert_to_tensor(sample_shape, dtype=tf.int32)] * 3)))
-
   @classmethod
-  def _params_event_ndims(cls):
-    return dict(loc=0, scale=0, concentration=0)
+  def _parameter_properties(cls, dtype, num_classes=None):
+    # pylint: disable=g-long-lambda
+    return dict(
+        loc=parameter_properties.ParameterProperties(),
+        scale=parameter_properties.ParameterProperties(
+            default_constraining_bijector_fn=(
+                lambda: softplus_bijector.Softplus(low=dtype_util.eps(dtype)))),
+        concentration=parameter_properties.ParameterProperties())
+    # pylint: enable=g-long-lambda
 
   @property
   def loc(self):
@@ -198,9 +200,11 @@ class GeneralizedExtremeValue(transformed_distribution.TransformedDistribution):
     return self._gev_bijector.concentration
 
   def _entropy(self):
-    # Use broadcasting rules to calculate the full broadcast sigma.
-    scale = self.scale * tf.ones_like(self.loc)
-    return 1. + tf.math.log(scale) + np.euler_gamma*(1. + self.concentration)
+    scale = tf.broadcast_to(self.scale,
+                            ps.broadcast_shape(ps.shape(self.scale),
+                                               ps.shape(self.loc)))
+    euler_gamma = tf.constant(np.euler_gamma, self.dtype)
+    return 1. + tf.math.log(scale) + euler_gamma * (1. + self.concentration)
 
   def _log_prob(self, x):
     with tf.control_dependencies(self._gev_bijector._maybe_assert_valid_x(x)):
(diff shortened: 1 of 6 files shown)


blacksde:
@srvasude @brianwa84 Could you help review this pr? Thx~

brianwa84:
Pavel, could you take a look?

tensorflow_probability/python/distributions/gev.py
```diff
@@ -0,0 +1,265 @@
...
+               allow_nan_stats=True,
+               name='GeneralizedExtremeValue'):
+    """Construct generalized extreme value distributions with location `loc`,
+    scale `scale`, and concentration `concentration`.
+
+    The parameters `loc`, `scale`, and `concentration` must be shaped in a way
+    that supports broadcasting (e.g. `loc + scale` + `concentration` is valid).
+
+    Args:
+      loc: Floating point tensor, the means of the distribution(s).
```
SiegeLordEx, line 119: It's not the actual mean. Maybe just call it "the location parameter of the distribution(s)"

tensorflow_probability/python/distributions/gev.py
```diff
@@ -0,0 +1,265 @@
...
+              high=tf.ones(batch_shape, dtype=dtype),
+              allow_nan_stats=allow_nan_stats),
+          # The GEV bijector encodes the CDF function as the forward,
+          # and hence needs to be inverted.
+          bijector=invert_bijector.Invert(
+              self._gev_bijector, validate_args=validate_args),
+          parameters=parameters,
+          name=name)
+
+  @staticmethod
```
SiegeLordEx, line 175: `_param_shapes` and `_param_event_ndims` methods, we have a new mechanism for describing this. In this case, you can do:

```python
  @classmethod
  def _parameter_properties(cls, dtype, num_classes=None):
    # pylint: disable=g-long-lambda
    return dict(
        loc=parameter_properties.ParameterProperties(),
        scale=parameter_properties.ParameterProperties(
            default_constraining_bijector_fn=(
                lambda: softplus_bijector.Softplus(low=dtype_util.eps(dtype)))))
        concentration=parameter_properties.ParameterProperties(),
    # pylint: enable=g-long-lambda
```

This new style lets us automatically constrain the parameters to the distribution.

tensorflow_probability/python/distributions/gev.py
```diff
@@ -0,0 +1,265 @@
...
+    return self._gev_bijector.scale
+
+  @property
+  def concentration(self):
+    """Distribution parameter for shape."""
+    return self._gev_bijector.concentration
+
+  def _entropy(self):
+    # Use broadcasting rules to calculate the full broadcast sigma.
+    scale = self.scale * tf.ones_like(self.loc)
```
SiegeLordEx, line 202: Do `tf.broadcast_to(self.scale, ps.broadcast_shape(ps.shape(self.scale), ps.shape(self.loc))` instead, it's a little more self-descriptive (you can drop the comment) and doesn't waste flops.

`ps` should be defined as `from tensorflow_probability.python.internal import prefer_static as ps` up top.

tensorflow_probability/python/distributions/gev.py
```diff
@@ -0,0 +1,265 @@
...
+    equal_zero = tf.equal(conc, 0.)
+
+    # Use broadcasting rules to calculate the full broadcast sigma.
+    mode_z = tf.where(equal_zero,
+                      tf.zeros_like(conc),
+                      tf.math.expm1(-conc * tf.math.log1p(conc)) /conc)
+
+    return self.loc + self.scale * mode_z
+
+  def _default_event_space_bijector(self):
```
SiegeLordEx, line 259: Drop this entirely, the superclass has a good default implementation for this distribution (it'll use the composition of sigmoid + GevCDF). Identity is particularly bad, since this distribution often isn't supported on the entire real line.

tensorflow_probability/python/distributions/gev.py
```diff
@@ -0,0 +1,265 @@
...
+
+  @property
+  def concentration(self):
+    """Distribution parameter for shape."""
+    return self._gev_bijector.concentration
+
+  def _entropy(self):
+    # Use broadcasting rules to calculate the full broadcast sigma.
+    scale = self.scale * tf.ones_like(self.loc)
+    return 1. + tf.math.log(scale) + np.euler_gamma*(1. + self.concentration)
```
SiegeLordEx, line 203: - Please add spaces around binary operators
- We try our best to avoid numpy type upcasting, so please write this as `tf.constant(np.euler_gamma, self.dtype)`

tensorflow_probability/python/distributions/gev.py
```diff
@@ -0,0 +1,265 @@
...
+
+  def _log_prob(self, x):
+    with tf.control_dependencies(self._gev_bijector._maybe_assert_valid_x(x)):
+      scale = tf.convert_to_tensor(self.scale)
+      z = (x - self.loc) / scale
+
+      conc = tf.convert_to_tensor(self.concentration)
+      equal_zero = tf.equal(conc, 0.)
+      log_t = tf.where(equal_zero, -z,
+                       -tf.math.log1p(z * conc) / conc)
```
SiegeLordEx, line 213: This needs to be written somewhat differently, because as is it'll produce NaN gradients when conc is 0, see https://github.com/tensorflow/probability/blob/master/discussion/where-nan.pdf for details. In this case, what you can do it something like:

```python
safe_conc = tf.where(equal_zero, 1., conc)
log_t = tf.where(equal_zero, -z,
        -tf.math.log1p(z * safe_conc) / safe_conc)
```

tensorflow_probability/python/distributions/gev.py
```diff
@@ -0,0 +1,265 @@
...
+
+      return (conc + 1) * log_t - tf.exp(log_t) - tf.math.log(scale)
+
+  def _mean(self):
+    conc = tf.convert_to_tensor(self.concentration)
+    equal_zero = tf.equal(conc, 0.)
+    less_than_one = tf.less(conc, 1.)
+
+    mean_z = tf.where(equal_zero,
```
SiegeLordEx, line 222: Same comment about NaN gradients as above (with the same solution, I imagine, although maybe conc=1 isn't a good "safe value").

tensorflow_probability/python/distributions/gev.py
```diff
@@ -0,0 +1,265 @@
...
+                      np.euler_gamma*tf.ones_like(conc),
```
SiegeLordEx, line 223: This can be `tf.fill(tf.shape(conc), tf.constant(np.euler_gamma, self.dtype))`

tensorflow_probability/python/distributions/gev.py
```diff
@@ -0,0 +1,265 @@
...
+                      tf.where(less_than_one,
+                               tf.math.expm1(tf.math.lgamma(1. - conc)) / conc,
+                               np.inf*tf.ones_like(conc))
```
SiegeLordEx, line 226: Same suggestion about using `tf.fill` instead.

tensorflow_probability/python/distributions/gev.py
```diff
@@ -0,0 +1,265 @@
...
+
+  def _stddev(self):
+    conc = tf.convert_to_tensor(self.concentration)
+    equal_zero = tf.equal(conc, 0.)
+    less_than_half = tf.less(conc, 0.5)
+
+    g1_square = tf.exp(tf.math.lgamma(1. - conc)) ** 2
+    g2 = tf.exp(tf.math.lgamma(1. - 2.*conc))
+
+    std_z = tf.where(equal_zero,
```
SiegeLordEx, line 239: Same comment about `NaN` gradients.

tensorflow_probability/python/distributions/gev.py
```diff
@@ -0,0 +1,265 @@
...
+
+    return self.loc + self.scale * mean_z
+
+  def _stddev(self):
+    conc = tf.convert_to_tensor(self.concentration)
+    equal_zero = tf.equal(conc, 0.)
+    less_than_half = tf.less(conc, 0.5)
+
+    g1_square = tf.exp(tf.math.lgamma(1. - conc)) ** 2
+    g2 = tf.exp(tf.math.lgamma(1. - 2.*conc))
```
SiegeLordEx, line 237: Spaces around binary ops, please.

tensorflow_probability/python/distributions/gev.py
```diff
@@ -0,0 +1,265 @@
...
+                     )
+
+    return self.scale * tf.ones_like(self.loc) * std_z
+
+  def _mode(self):
+    conc = tf.convert_to_tensor(self.concentration)
+    equal_zero = tf.equal(conc, 0.)
+
+    # Use broadcasting rules to calculate the full broadcast sigma.
+    mode_z = tf.where(equal_zero,
```
SiegeLordEx, line 253: Same comment about `NaN` gradients.

tensorflow_probability/python/distributions/gev_test.py
```diff
@@ -0,0 +1,330 @@
...
+import tensorflow.compat.v1 as tf1
+import tensorflow.compat.v2 as tf
+import tensorflow_probability as tfp
+
+from tensorflow_probability.python.internal import test_util
+
+tfd = tfp.distributions
+
+
+class _GEVTest(object):
```
SiegeLordEx, line 34: Since I had all those concerns about NaN gradients, would you mind adding a test for those? See https://github.com/tensorflow/probability/blob/612a9cc27cc96f6dad433d604c8255eb9daf5e2e/tensorflow_probability/python/distributions/generalized_normal_test.py#L168-L186 for an example. I especially care about the branch points of concentration=1, 0.5, 0 but other spots may be good to check too.

tensorflow_probability/python/distributions/gev_test.py
```diff
@@ -0,0 +1,330 @@
...
+
```
SiegeLordEx, line 35: Would you also mind adding a test for broadcasting? Something like this: https://github.com/tensorflow/probability/blob/612a9cc27cc96f6dad433d604c8255eb9daf5e2e/tensorflow_probability/python/distributions/multivariate_student_t_test.py#L36-L108

SiegeLordEx (changes requested):
Looks good, thanks! I glanced over the math, it seemed okay. I mainly have concerns about the gradients in a few spots.

tensorflow_probability/python/distributions/gev.py
blacksde: Thx for pointing this out, I will also fix all other places of misusing mean and loc.

blacksde:
> Looks good, thanks! I glanced over the math, it seemed okay. I mainly have concerns about the gradients in a few spots.

Hi @SiegeLordEx , Thanks a lot for all your valuable comments. I have already fixed all of them(although I didn't reply one by one inline :) ). There are two places to point out:

1. for the broadcasting test case you suggested, I referred to the one in half_student_t_test as its test case class is the same as what I used previously. I don't want to change the basic structure too much so I was hoping this one would be OK.

2. for the gradient NaN issue, I also updated some logic in gev_cdf as the cdf of gev relies on this bijector.

Let me know if there is anything else needed to be updated.

SiegeLordEx (approved):

SiegeLordEx:
Looks good, thanks. I'll send this along to our internal review and that'll get this merged.

PR #1154, opened 2020-11-08
Adding an example notebook of variational GP

pengluyaoyao:
Demonstrating how variational GP is used in multiclass-classification


pengluyaoyao:
@googlebot I signed it!

PR #1160, opened 2020-11-12
Fix log_prob example in RealNVP docstring

hartikainen:


tensorflow_probability/python/bijectors/real_nvp.py
```diff
@@ -105,7 +105,7 @@ class RealNVP(bijector_lib.Bijector):
 
   x = nvp.sample()
   nvp.log_prob(x)
-  nvp.log_prob(0.)
```
hartikainen, line 108: `nvp.log_prob(0.)` raises:
```
ipdb> nvp.log_prob(0.)
2020-11-12 19:34:57.888647: W tensorflow/core/framework/op_kernel.cc:1767] OP_REQUIRES failed at strided_slice_op.cc:108 : Invalid argument: Index out of range using input dim 0; input has only 0 dims
*** tensorflow.python.framework.errors_impl.InvalidArgumentError: Index out of range using input dim 0; input has only 0 dims [Op:StridedSlice] name: real_nvpMultivariateNormalDiag/l
```

tensorflow_probability/python/bijectors/real_nvp.py
```diff
@@ -105,7 +105,7 @@ class RealNVP(bijector_lib.Bijector):
...
+  nvp.log_prob((0.0, 0.0, 0.0))
```
hartikainen, line 108:
```
ipdb> nvp.log_prob((0.0, 0.0, 0.0))
<tf.Tensor: shape=(), dtype=float32, numpy=-2.7568154>
```

SiegeLordEx: Would you mind changing this to be a list rather than a tuple? We (as a style point) prefer to use lists whenever possible for arrays of numbers.

SiegeLordEx (changes requested):
Thanks!

tensorflow_probability/python/bijectors/real_nvp.py
hartikainen: Changed. Thanks!

SiegeLordEx (approved):

Issue #1169, opened 2020-11-16
Autoregressive flow on jax


mattwescott:
The docstring for `MaskedAutoregressiveFlow` suggests `AutoregressiveNetwork`, which fails on the jax substrate due to Keras dependencies. If there are still missing pieces, are you interested in contributions?

@sharadmv @SiegeLordEx

brianwa84:
We haven't yet settled on a "right" way to do NNs in JAX.
For now, an approach you can use is passing kwargs to log_prob, something
like below.

If you have a proposal for how to integrate these bijectors (also thinking
about RealNVP) with flax/objax/haiku/oryx, it'd be great to share that.
Could perhaps send a PR to add a notebook under discussion/examples.

np.random.seed(213)

n = 2000
x2 = np.random.randn(n).astype(dtype=np.float32) * 2.
x1 = np.random.randn(n).astype(dtype=np.float32) + (x2 * x2 / 4.)
data = np.stack([x1, x2], axis=-1)
scatter(*data.T);title('data');show()

def make_bijector(y, layers=None):
for layer in layers[:-1]:
# print(y.shape, [l.shape for l in layer])
y = jnp.tanh(jnp.einsum('...b,ba->...a', y, layer[0]) + layer[1])
# print(y.shape, [l.shape for l in layers[-1]])
y = jnp.einsum('...b,ba->...a', y, layers[-1][0]) + layers[-1][1]
# print('final',y.shape)
return tfb.Shift(y[..., :1])(tfb.Scale(log_scale=y[..., 1:]))

layers=[(np.random.normal(size=[2, 20]), np.random.normal(size=[20])),
(np.random.normal(size=[20, 10]), np.random.normal(size=[10])),
(np.random.normal(size=[10, 10]), np.random.normal(size=[10])),
(np.random.normal(size=[10, 2]), np.random.normal(size=[2]))]
total_steps = 0

made = tfp.bijectors.MaskedAutoregressiveFlow(bijector_fn=make_bijector)

# print(made.forward(np.random.uniform(size=[5, 2]), layers=layers))
# print(made.inverse(np.random.uniform(size=[2000, 2]), layers=layers))

distribution = tfd.TransformedDistribution(
distribution=tfd.Sample(tfd.Normal(loc=0., scale=1.), sample_shape=[2]),
bijector=made)
# print(distribution.sample(len(data), bijector_kwargs=dict(layers=layers),
seed=jax.random.PRNGKey(123)).shape)
[...]

PR #1176, opened 2020-11-23
Add example autoregressive flow on JAX

mattwescott:
Motivated by #1169


tensorflow_probability/examples/jupyter_notebooks/autoregressive_flow_on_jax.ipynb
```diff
@@ -0,0 +1,485 @@
+{
```
SiegeLordEx, line 1: <p>I think Haiku made a release the other day, could we install it from Pypi?</p><p> </p><p>Also, you don't use the psd_kernels anywhere.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/probability/pull/1176/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

mattwescott:
This approach is more limited than I'd realized. From  @tomhennigan in deepmind/dm-haiku#84:

>Haiku (and other OO JAX libraries) need you to use wrapped versions of JAX transforms if you are using JAX transforms that close over the usage of Modules without explicitly passing state. In Haiku this means inside your hk.transformed function you need to use hk.jit etc if you are wrapping code that makes use of a module.
>
> For modules that only make use of parameters/rng in Haiku (e.g. if you are using hk.transform and not hk.transform_with_state) there is a "trick" you can use to handle these inner JAX transforms, which is making use of the jax.core.eval_context() context manager inside your modules). This will cause values to not be eagerly "staged out" to the inner transform (unless they have been explicitly passed in to it) and will stop parameters becoming (deeper) tracers.
> 
> Here is an updated copy of your notebook with the trick applied: https://colab.research.google.com/gist/tomhennigan/94ce417598baeb0b8db703981ea83992/autoregressive_flow_on_jax-ipynb-txt.ipynb. I've wrapped this up as a decorator (run_in_eval_context) to make the delta from your original code small (defining and using the decorator is the only meaningful difference).

In this example, the loop in `MaskedAutoregressiveFlow` is now unrolled to avoid any inner transforms. There may be other bijectors with inner transformations that aren't so amenable. The `eval_context` trick might support more bijectors, but it seems opaque and precludes using state. Manually threading the parameters seems tedious. Supporting Haiku-wrapped transforms within `tfp` doesn't sit right. There must be a better way. What am I missing? Maybe Oryx supports a cleaner implementation?

MilesCranmer:
Also very interested in this! Thanks very much for the implementation @mattwescott.

Issue #1161, opened 2020-11-13
Could higher level inference/fitting functions be part of TFP?


jeffpollock9:
Hi,

Most of the time when I see comparisons of TFP to other probabilistic frameworks, one of the downsides to TFP is how hard it can be to write code to do inference (see for example https://colcarroll.github.io/ppl-api).

I am wondering if there is any scope to add a higher level API to TFP which wraps up some of the common patterns and provides sensible defaults? There is already the [fit_with_hmc](https://github.com/tensorflow/probability/blob/0b05d7ddfb3f59510060717685557c307d056b35/tensorflow_probability/python/sts/fitting.py#L209) function for structural time series models which does this and I think it's really useful.

If a user codes up some sort of `JointDistribution` then I think it encodes almost everything you need to call a simple function `do_inference(joint_distribution)` which behaves sensibly by default. I tried a small example with [MAP](https://en.wikipedia.org/wiki/Maximum_a_posteriori_estimation) estimation for a simple linear regression and it seems to work. I've added the full code as a [gist](https://gist.github.com/jeffpollock9/ac4dac806741f8be8c962fa4f5be17cd) but the main parts are:

``` python
@tfd.JointDistributionCoroutine
def joint_dist():
    intercept = yield Root(tfd.Normal(loc=0.0, scale=1.0, name="intercept"))
    coefficients = yield Root(
        tfd.MultivariateNormalDiag(
            loc=0.0, scale_diag=tf.ones([NUM_COEFFICIENTS]), name="coefficients"
        )
    )
    scale = yield Root(tfd.Exponential(rate=1.0, name="scale"))
    loc = intercept[..., tf.newaxis] + tfl.matvec(design_matrix, coefficients)
    yield tfd.MultivariateNormalLinearOperator(
        loc=loc,
        scale=tfl.LinearOperatorScaledIdentity(
            num_rows=NUM_OBSERVATIONS, multiplier=scale
        ),
        name="observations",
    )

estimates, opt = find_map(
    joint_distribution=joint_dist,
    condition_on={"observations": observations},
    batch_shape=[10],
)

best_estimate = {k: v[tf.argmin(opt.objective_value)] for k, v in estimates.items()}

truths = [random_intercept, random_coefficients, random_scale]

for (name, estimate), truth in zip(best_estimate.items(), truths):
    print(f"{name}:")
    print(f"\testimate: {estimate.numpy()}")
    print(f"\ttruth:    {truth.numpy()}")

# intercept:
[...]

Issue #570, opened 2019-09-26
Laplace approximation


dionman:
Is there any function computing Laplace approximation (similar to ed.Laplace) supported at tensorflow probability at the moment?

jeffpollock9:
I'm not aware of any Laplace approximation functionality but would definitely be keen to help with contributing it.

brianwa84:
Building a cookie-cutter laplace_approx(target_density, location) shouldn't
be too hard.
Could probably do something like
tfd.MVNLinearOperator(LinearOperatorInvert(LinearOperstorFull(tf.hessians(target_density(xs),
xs))))
A couple thoughts:
tf.hessians d/n work in tf2, here's a workaround:
https://github.com/tensorflow/tensorflow/issues/29781#issuecomment-504980912
The hessian is not necessarily PD unless the location is in a local minimum.

Bigger challenge may be how are you going to find the location.

Brian Patton | Software Engineer | bjp@google.com

junpenglao:
I would try using lbgfs to find the location - it should be even possible to find multi mode using tfp.optimizer.converged_all when the log_prob is batch-friendly.

Padarn:
Would it be valuable to have `laplace_approx(target_density, location)` exposed, or should this look more like the Edward version that does the location finding.

From looking at the Edward implementation, there is a Base class `Inference` for which tfp doesn't seem to have an equivalent of - or am I missing something looking through the code?

jeffpollock9:
If at all useful to anyone coming across this issue, I coded up some Laplace approximation stuff recently. Usage is in this comment, which also contains a gist of the implementation: https://github.com/tensorflow/probability/pull/1178#issuecomment-737183260

Would be happy to spend some time on this making a more formal proposal for adding something like this to TFP if that was of interest.

PR #1178, opened 2020-11-24
changes to rethinking.quap.py

jeffpollock9:
Hello, as per https://github.com/tensorflow/probability/issues/1161 I had a look at the `quap` function. I've added a few noddy changes to it, I hope that it's helpful and not annoying (if the latter please let me know and I will change back).

@ColCarroll is there anything in particular you would like me to attempt to improve in the `quap` function further? I've updated my version of `find_map` (gist here: https://gist.github.com/jeffpollock9/0cc298641722277cf30e438f04eb8609) so please let me know if there is anything you'd like to take. I wasn't sure the best way to change `quap` to use this stuff without changing the API, which I would prefer not to do unless someone else is happy with that.

### DIFF
--- a/tensorflow_probability/examples/statistical_rethinking/rethinking/quap.py
+++ b/tensorflow_probability/examples/statistical_rethinking/rethinking/quap.py
@@ -55,7 +55,7 @@ def quap(joint_dist, data=None, max_tries=20, initial_position=None, name=None):
     joint_dist: A `JointDistributionNamed` or `JointDistributionSequential`
       model. Also works with auto batched versions of the same.
     data: Optional `dict` of data to condition the joint_dist with. The return
-      value will be conditioned o this data. If this is `None`, the return
+      value will be conditioned on this data. If this is `None`, the return
       value will be a quadratic approximation to the distribution itself.
     max_tries: Optional `int` of number of times to run the optimizer internally
       before raising a `RuntimeError`. Default is 20.
--- a/tensorflow_probability/examples/statistical_rethinking/rethinking/quap.py
+++ b/tensorflow_probability/examples/statistical_rethinking/rethinking/quap.py
@@ -58,7 +58,7 @@ def quap(joint_dist, data=None, max_tries=20, initial_position=None, name=None):
       value will be conditioned o this data. If this is `None`, the return
       value will be a quadratic approximation to the distribution itself.
     max_tries: Optional `int` of number of times to run the optimizer internally
-      before raising a `RuntimeError`. Default is 10.
+      before raising a `RuntimeError`. Default is 20.
     initial_position: Optional `dict` to initialize the optimizer. Keys should
       correspond to names in the JointDistribution. Defaults to random draws
       from `joint_dist`.
--- a/tensorflow_probability/examples/statistical_rethinking/rethinking/quap.py
+++ b/tensorflow_probability/examples/statistical_rethinking/rethinking/quap.py
@@ -59,7 +59,7 @@ def quap(joint_dist, data=None, max_tries=20, initial_position=None, name=None):
       value will be a quadratic approximation to the distribution itself.
     max_tries: Optional `int` of number of times to run the optimizer internally
       before raising a `RuntimeError`. Default is 10.
-      initial_position: Optional `dict` to initialize the optimizer. Keys should
+    initial_position: Optional `dict` to initialize the optimizer. Keys should
       correspond to names in the JointDistribution. Defaults to random draws
       from `joint_dist`.
     name: Python `str` name prefixed to ops created by this function.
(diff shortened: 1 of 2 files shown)


ColCarroll:
Thanks for this! It looks fine as is, but also right now is a great time to change the API there, since it is not publicly exposed or even used in the notebooks yet. 

Would it be possible to:
- Rename `find_map` to `_bfgs_optimize`, and just `return opt` out of the final `control_dependencies`
- Add a `find_MAP` calling the above and using your logic to return a dictionary
- Add a `quap` calling the above and using my logic to return a JDNamed

What do you think?

jeffpollock9:
Yes that sounds much better than the ideas I had for this - thanks!

I'll try a few ideas along those lines and see how it looks. I'd also like to support Laplace approximations from this sort of code in some way, as per https://github.com/tensorflow/probability/issues/570, but not sure if that would be well placed in the rethinking examples.

ColCarroll:
I was playing with this yesterday, and another problem is that the Laplace approximation is basis-dependent, so the (approximate) hessian you get back from `lbfgs` depends on the bijectors being used. It seems like there's some choice, then (see [this](https://link.springer.com/article/10.1023/A:1007558615313), for example) over what gets returned.

jeffpollock9:
Thanks for the reference, I'll try to go through it soon.

I think the MAP estimate also has problems being basis-dependent, I recall a lot of discussion on the [stan forums](https://discourse.mc-stan.org/) as to whether or not running stan in `optimizing` mode should include the bijector jacobian adjustment but I can't find it now and don't think there was a definite answer :(

My favourite document on this stuff is https://mc-stan.org/users/documentation/case-studies/mle-params.html, thought I'd link it on the off chance it is useful to anyone, this is the summary:

> The moral of the story is that full Bayesian inference is insensitive to parameterization as long as the appropriate Jacobian adjustment is applied. In contrast, posterior modes (i.e., maximum a posteriori estimates) do change under reparameterization, and thus are no true Bayesian quantity.

I guess I'll think a little harder about how the Laplace stuff could work and go on as I was with the MAP stuff, although any comments about any of this would be great.

ColCarroll:
Ah, thanks for the link!

Yes, I imagine the docstring would include a discussion and some of these references, and probably there should be a `bijector` argument that defaults to `default_event_space_bijector`, and makes this nice warning about the output being "no true Bayesian quantity."

Does it make sense to you to merge the changes on this PR into `quap.py`, and you can think about a PR into the main repo with a design closer to the one in the gist?

jeffpollock9:
Yes, that sounds good to me.

jeffpollock9:
Hi @ColCarroll, I've tried to merge all the ideas together into one function, and have uploaded it here https://gist.github.com/jeffpollock9/2a6950164711d8f813561b8253c9cfa9 as I wasn't sure where to put it as I'm hoping this could perhaps be added to the TFP API (shall I open a separate PR with doc + tests?) instead of the rethinking examples. In that case perhaps we could leave the `quap` function or merge in the small changes (mostly for the docs) in this PR. 

This new attempt is different from both of our original functions as it returns a transformed multivariate normal. I think this is quite nice as the object naturally contains all the information that (I think/hope) will be useful for folks. I've not tested it too much but I think this can handle more cases than the current `quap` function and also keeps the full estimate of the MVN covariance matrix instead of just the diagonal. 

An example of the usage:

``` python
print(tf.__version__)
# ==> 2.5.0-dev20201201

print(tfp.__version__)
# ==> 0.12.0-dev20201201


@tfd.JointDistributionCoroutineAutoBatched
def joint_dist():
    yield tfd.Normal(loc=1.0, scale=1.0, name="a")
    yield tfd.Normal(loc=[2.0, 2.0], scale=2.0, name="b")
    yield tfd.Gamma(concentration=[2.0, 2.0, 2.0], rate=10.0, name="c")
    yield tfd.CholeskyLKJ(dimension=3, concentration=1.0, name="d")


# without conditioning
approximation = laplace_approximation(joint_dist)

names = joint_dist._flat_resolve_names()

point_estimate = approximation.bijector(approximation.distribution.mode())

for name, estimate in zip(names, point_estimate):
    print(f"{name}:")
    print(estimate.numpy())
    print("")

# ==>
# a:
# 0.9999998

# b:
# [2.000008  1.9999926]

# c:
# [0.09999622 0.10000692 0.09999667]

# d:
# [[ 1.000000e+00  0.000000e+00  0.000000e+00]
#  [ 4.016183e-07  1.000000e+00  0.000000e+00]
#  [ 5.092372e-01 -4.885684e-01  7.085044e-01]]

# with conditioning, also shows that sampling from the approximation works
approximation = laplace_approximation(joint_dist, data={"d": approximation.sample()[3]})

point_estimate = approximation.bijector(approximation.distribution.mode())

for name, estimate in zip(names, point_estimate):
    print(f"{name}:")
    print(estimate.numpy())
    print("")

# ==>
# a:
# 0.9999998

# b:
# [1.9999962 1.999986 ]

# c:
# [0.10000057 0.10000078 0.09999969]
```

If you (or anyone else) has any comments that would be great but no rush of course.

Thanks again.

brianwa84:
I think we'd probably be open to a laplace approximation under `tfp.experimental.distributions` if you want to send a PR to put something like this in there. @ColCarroll wdyt?

ColCarroll:
+1 that'd be great -- happy to help review/advise, too.

jeffpollock9:
@brianwa84 @ColCarroll thanks - that sounds great. I'll try to send over a PR for a Laplace approximation in `tfp.experimental.distributions` when I have some time.

PR #1179, opened 2020-11-25
Add GitHub actions

SiegeLordEx:
Sample run: https://github.com/SiegeLordEx/probability/actions/runs/382285993


testing/run_travis_lints.sh
```diff
@@ -18,13 +18,13 @@ set -v  # print commands as they are executed
 set -e  # fail and exit on any command erroring
```
ColCarroll, line 19: maybe filename changes to, `run_github_lints.sh` or `run_ci_lints.sh`?

PR #1187, opened 2020-12-04
Remove local rademachers

hartikainen:
I noticed that `{dense,conv}_variational.py` modules were using local rademacher functions. Rademacher is now available at `tfp.random.rademacher`, so I thought it would be nice to have the variational layers use that instead of local ones.


ColCarroll:
Looks good to me - thanks, @hartikainen!

ColCarroll (approved):

PR #1212, opened 2021-01-07
UnitNormal distribution

jeffpollock9:
Hello,

This PR adds a special case of the `Normal` distribution with `loc = 0.` and `scale = 1.`, the motivation for this is as described in [the stan manual](https://mc-stan.org/docs/2_20/functions-reference/normal-distribution.html) which describes their `std_normal` distribution. In short, it allows a slight simplification of typical code like:

``` python
tfd.Normal(loc=tf.zeros(num_groups, dtype=tf.float64), scale=1.)
```
to:
``` python
tfd.UnitNormal(batch_shape=num_groups, dtype=tf.float64)
```

And as far as I can tell in some cases it is slightly more efficient.

Let me know if there is anything else I can do, or if this addition is not of interest (absolutely fine!).

Thanks!

### DIFF
--- a/tensorflow_probability/python/distributions/BUILD
+++ b/tensorflow_probability/python/distributions/BUILD
@@ -141,6 +141,7 @@ multi_substrate_py_library(
         ":truncated_cauchy",
         ":truncated_normal",
         ":uniform",
+        ":unit_normal",
         ":variational_gaussian_process",
         ":vector_exponential_diag",
         ":von_mises",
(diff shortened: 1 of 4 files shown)


brianwa84:
What about changing this to a method on `tfd.Normal`? (and using the more familiar "standard" nomenclature)

```python
@classmethod
def standard(cls, shape, dtype=tf.float32):
  return cls(loc=tf.zeros(shape, dtype), scale=1.)
```

My sense is there's not appetite to support a new class, but a classmethod is fairly maintainable. Eventually we could replace the impl with something more efficient like you propose, but we'd want to pull that efficiency out into a more general `AddBatchShape` or `BroadcastBatch` meta-distribution.

jeffpollock9:
@brianwa84 thanks - that sounds sensible. I'm not super keen on the classmethod as I can't see myself ever using it - so I think I'll close this for now. Hope that's ok.

PR #1220, opened 2021-01-15
Add missing dtype in `diagonal_mass_matrix_adaptation`

chrism0dwk:
Currently, `tf.experimental.mcmc.DiagonalMassMatrixAdaptation` will not work with float64 variance and state parts.  This is because a `tf.ones` tensor is constructed without an explicit `dtype` in `_make_momentum_distribution`.  Here's a simple suggested fix.


chrism0dwk:
@googlebot I signed it!

ColCarroll (approved):

ColCarroll:
Thanks, Chris!

PR #1228, opened 2021-02-04
[WIP] experimental Laplace approximation

jeffpollock9:
Hello, as per the discussion in #1178 I've started to code up how a Laplace approximation over a joint distribution could look. It's still very much WIP as these things are missing/I'm not sure about:

1. The design - do we want users to input a joint distribution and data? An alternative would be allowing them to create their own target log prob function like e.g. mcmc does? There is also the possibility of creating a `LaplaceApproximation` distribution class and use that instead of the current transformed MVN?
2. This doesn't work in eager mode without the `tf.function` annotation, I couldn't quite understand why. There are also (related?) issues with running this with jax and numpy modes so those tests fail. I can dig deeper into this but a small nudge in the right direction would really help.
3. The documentation needs to be written. I was hoping to do this last to avoid having to re-write if anything changes.

@ColCarroll just tagging you as you said you would be happy to have a look - hope that is ok. Any comments 

Thanks!


tensorflow_probability/python/experimental/distributions/laplace_approximation.py
```diff
@@ -0,0 +1,216 @@
...
+from tensorflow_probability.python.bijectors import chain
+from tensorflow_probability.python.bijectors import joint_map
+from tensorflow_probability.python.distributions import transformed_distribution
+from tensorflow_probability.python.distributions import mvn_tril
+from tensorflow_probability.python.math import value_and_gradient
+from tensorflow_probability.python.optimizer import bfgs_utils
+from tensorflow_probability.python.optimizer import bfgs
+
+
+def _common_dtype(dtypes, dtype_hint=None):
```
ColCarroll, line 34: should be able to use `from tensorflow_probability.python.internal import dtype_util` and `dtype_util.common_dtype(dist._get_single_sample_distributions())`

tensorflow_probability/python/experimental/distributions/laplace_approximation.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def _bfgs_minimize(
+    joint_dist,
+    bijector,
+    initial_values,
+    initial_inverse_hessian_estimate,
+    validate_convergence,
+):
+  """Minimize the -ve log prob of a joint distribution using BFGS."""
```
ColCarroll, line 85: typo? `ve`

tensorflow_probability/python/experimental/distributions/laplace_approximation.py
```diff
@@ -0,0 +1,216 @@
...
+
+  names = joint_dist._flat_resolve_names()
+
+  def loss(unconstrained_values):
+    constrained_values = bijector.forward(unconstrained_values)
+    return -_log_prob(joint_dist, **dict(zip(names, constrained_values)))
+
+  @tf.function(autograph=False)
```
ColCarroll, line 93: We try to avoid adding the function annotation internally, and let users add it where necessary. Would that be terrible here?

tensorflow_probability/python/experimental/distributions/laplace_approximation_test.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+  def testJointDistributionSequentialAutoBatchedWithDefaults(self):
+    joint_dist = tfd.JointDistributionSequentialAutoBatched([
+        tfd.MultivariateNormalDiag(loc=0.123, scale_diag=[1., 2., 3.]),
+        tfd.StudentT(loc=2.5, scale=3., df=3.),
+    ])
+
+    approx = tfde.laplace_approximation(joint_dist)
+    mean = approx.bijector(approx.distribution.mean())
```
ColCarroll, line 109: why does this need to go through the bijector here? (I _think_ the bijector should be the identity here anyways, right? so maybe it doesn't...)

ColCarroll (commented):
This looks great! 

It'd be nice to see another test or two that the hessian approximation recovers (approximately) the covariance, and maybe a test that shows that providing different bijectors results in different laplace approximations (which should probably be mentioned in the docstring).

I guess the other TODO is filling out the docs for laplace approximation.

Let me know if you want help with any of this, or won't have time for a while -- this looks like some careful work, and I don't want it to get lost!

tensorflow_probability/python/experimental/distributions/laplace_approximation.py
jeffpollock9: ah cool, was wondering how to use that function, done in 891de2793

tensorflow_probability/python/experimental/distributions/laplace_approximation.py
jeffpollock9: fixed in 0f11fdc86

tensorflow_probability/python/experimental/distributions/laplace_approximation.py
jeffpollock9: yes I didn't want to add this, but you get an error about `None` gradients without it :( I wasn't sure if this is related to the same issue with gradients not working in jax and numpy modes (I don't have much understanding in that area). If you have any idea about how to fix that would be great. If you e.g. remove the annotation and debug one of the tests you can see:

``` python
(Pdb) loss_and_gradient(initial_position)
(<tf.Tensor: shape=(10,), dtype=float32, numpy=
array([4.7611666, 8.704665 , 4.0540056, 4.3684416, 3.7730975, 3.951736 ,
       4.1365232, 3.8006756, 4.6959515, 3.9844842], dtype=float32)>, None)
```

tensorflow_probability/python/experimental/distributions/laplace_approximation_test.py
jeffpollock9: here the bijector only splits up the MVN samples according to the event shape of the distributions in the joint distribution (`[3]` and `[]` in this case):

``` python
>>> approx.distribution.mean()
<tf.Tensor: shape=(4,), dtype=float32, numpy=array([0.12300005, 0.12300101, 0.12300146, 2.500002  ], dtype=float32)>
>>> approx.bijector(approx.distribution.mean())
_TupleWrapper((<tf.Tensor: shape=(3,), dtype=float32, numpy=array([0.12300005, 0.12300101, 0.12300146], dtype=float32)>, <tf.Tensor: shape=(), dtype=float32, numpy=2.500002>))
```

so have just used it for some convenience.

jeffpollock9:
> This looks great!
> 
> It'd be nice to see another test or two that the hessian approximation recovers (approximately) the covariance, and maybe a test that shows that providing different bijectors results in different laplace approximations (which should probably be mentioned in the docstring).
> 
> I guess the other TODO is filling out the docs for laplace approximation.
> 
> Let me know if you want help with any of this, or won't have time for a while -- this looks like some careful work, and I don't want it to get lost!

Cool!

Yes those sound like good tests, I'll add them to my TODO list at the top of the PR. If you think the API looks reasonable then I would be happy to start filling out the documentation? The main thing I need help with is figuring out what to do with this `None` gradient error that you get without the `@tf.function` annotation in TF mode or when running with jax (as mentioned above with an example) - so any input there would be amazing.

I should have some time for this here and there over the next couple of weeks.

ColCarroll:
Hey @jeffpollock9 -- I used some of the ideas from your `_transform_reshape_split` bijector in the [new windowed sampler](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/experimental/mcmc/windowed_sampling.py#L93) -- (re) using some of that code might be helpful here.

jeffpollock9:
> Hey @jeffpollock9 -- I used some of the ideas from your `_transform_reshape_split` bijector in the [new windowed sampler](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/experimental/mcmc/windowed_sampling.py#L93) -- (re) using some of that code might be helpful here.

@ColCarroll cool, thanks for sharing, perhaps we can pull this out as a new bijector for all to use? I'm not sure when I'll get more time to spend on this stuff but I am sure I will get to it at some point. In the meantime if you'd like to edit any of my stuff in any way please feel free.

jeffpollock9:
@ColCarroll if at all interesting, I've spent a little more time (not as much as I would like) on this and added another feature that I hope is useful - the ability to turn on/off the jacobian adjustment as described a little in the [stan manual](https://mc-stan.org/docs/2_26/reference-manual/writing-models-for-optimization.html).

I've also added further testing of the covariance matrix as you suggested.

I'm not sure in what cases you'd want to turn on/off the jacobian adjustment, in stan it is always off, but in the first example I've tried I think I'd rather have it on? Here is the code and plots:

``` python
import matplotlib.pyplot as plt
import seaborn as sns
import numpy as np

tfd = tfp.distributions
tfb = tfp.bijectors
tfde = tfp.experimental.distributions

import laplace_approximation as tfde

ADD_LOG_DET_JACOBIAN = True

a = tfd.LogNormal(loc=np.array(0.), scale=1.)
b = tfd.Gamma(concentration=np.array(2.), rate=10.)

joint_dist = tfd.JointDistributionSequential([a, b])

initial_values = [np.linspace(0.01, 5., 10)] * 2

e_approx = tfde.laplace_approximation(
    joint_dist,
    bijectors=[tfb.Exp()] * 2,
    initial_values=initial_values,
    add_log_det_jacobian=ADD_LOG_DET_JACOBIAN)

s_approx = tfde.laplace_approximation(
    joint_dist,
    bijectors=[tfb.Softplus()] * 2,
    initial_values=initial_values,
    add_log_det_jacobian=ADD_LOG_DET_JACOBIAN)

num_samples = 100_000
e_samples = e_approx.sample(num_samples)
s_samples = s_approx.sample(num_samples)
true_samples = joint_dist.sample(num_samples)

all_samples = [e_samples, s_samples, true_samples]
labels = ["exp", "softplus", "truth"]

fig, axs = plt.subplots(2, 1)

axs[0].set_title("LogNormal(0, 1)")
axs[0].set_xlim(-1, 10)

axs[1].set_title("Gamma(2, 10)")
axs[1].set_xlim(-1, 1)

for s, l in zip(all_samples, labels):
  sns.kdeplot(s[0].numpy(), label=l, ax=axs[0])
  sns.kdeplot(s[1].numpy(), label=l, ax=axs[1])

handles, labels = axs[1].get_legend_handles_labels()
fig.legend(handles, labels, loc="upper center")
fig.show()
```

with adjustment:

with_ldj

without the adjustment:

without_ldj

PR #1231, opened 2021-02-07
Fix a typo in the description of fit_surrogate_posterior

IvanUkhov:
I suppose `result_fn` was renamed to `trace_fn` at some point; however, the docstring was updated only partially.


davmre:
Thanks for the fix! I've pulled this in internally; it should get officially checked in in the next day or two.

PR #1233, opened 2021-02-07
Add experimental `GibbsKernel`

chrism0dwk:
`GibbsKernel` is a `tfp.mcmc.TransitionKernel` that allows a sequence
of transitional kernels to be bundled together to sequentially update
different parts of a full state.  Currently the semantic is to map
one state part per Gibbs step, though this is definitely up for
discussion.

`GibbsKernel` currently refers to each state part by index.  This 
leads to poor readability of the code, particularly in large Gibbs
schemes where the `target_log_prob_fn` definition scrolls off the top
of the page.  The time seems right to PR this kernel, as discussions
around turnkey MCMC solutions are progressing!


ColCarroll:
Thanks, Chris! I started taking a look at this, but feel free to ping in a few days if there are no comments.

chrism0dwk:
@ColCarroll no worries!   I feel a Gibbs sampling capability is something badly needed in TFP currently.  Although the HMC (and derivatives) literature focuses on single-kernel chains, the wider class of models involving discontinuous `log_prob` functions is important -- not least in current event-time analysis in the COVID epidemic. 

I should also provide a description of the logic of `GibbsKernel`:

1. `GibbsKernel` is essentially a meta-kernel which sequentially calls instances of `tfp.mcmc.TransitionKernel` on different state parts;
2. `GibbsKernelResults` tracks the value of `target_log_prob` (on the target scale!), whilst keeping the individual steps' "results" structures stored in a list.  
3. `GibbsKernel.one_step` sequentially forwards `target_log_prob` to each `previous_results` structure before passing it to the respective Gibbs step's `one_step` function.
4. Instances of `tfp.mcmc.TransformedTransitionKernel` are detected, and the bijector is invoked in order to map `target_log_prob` --> `transformed_target_log_prob`.
5. The mixing of non-gradient based kernels (e.g. RWM) with gradient-based kernels poses an issue with forwarding gradient information.  Currently, we take the pragmatic approach of calling `k.bootstrap_results` to reset `target_log_prob` and associated gradients if `k`'s `previous_results` contains gradient information.  Doubtless we could be cleverer!
6. A design issue with this kernel is the presence of a Python `list` within the `GibbsKernelResults` `named_tuple`.  This breaks the recursive functions such as `get_innermost` and `get_outermost`.  Ideas welcome.

coatless:
@ColCarroll ping

tensorflow_probability/python/experimental/mcmc/gibbs_kernel.py
```diff
@@ -0,0 +1,411 @@
...
+        # Require to check if all kernel.is_calibrated is True
+        self._parameters = dict(
+            target_log_prob_fn=target_log_prob_fn,
+            kernel_list=kernel_list,
+            name=name,
+        )
+
+    @property
+    def is_calibrated(self):
+        return True
```
brianwa84, line 281: presumably only calibrated as long as each make_kernel_fn is also calibrated?

brianwa84:
A few thoughts after looking over this:
1. If you had a complete conditional (which we should for the MVN case?), how would you implement the `make_kernel_fn`? Could `make_kernel_fn` return a `tfd.Distribution`, and if we get back a distribution we literally call `dist.sample(seed=seed)`? Or could we add a `SamplingKernel` that takes a `Distribution` and `one_step` just returns another sample? Can you add this as an example?
2. Could you add an example that incorporates a discrete latent, and showcases using a different kernel for a different state part (MH for discrete, HMC for continuous)?
3. It is a bit limiting that we only deal with single state parts at a time, though I guess that could be generalized later. Also a bit limiting to address them by index (though I guess that's more a limitation of the overall TFP MCMC stack supporting only lists and tuples).
4. The gradient resetting stuff is icky (but we do similar things in REMC).
5. I don't quite follow the necessity of unwrapping `TransformedTransitionKernel` instances. Can you explain why that's necessary? We have some cases of joint distributions where the descendent event space bijectors depend on the ancestral sample values. e.g:
    ```python
    @tfd.JointDistributionCoroutine
    def model():
       low = yield tfd.Normal(0, 1)
       span = yield tfd.LogNormal(0, 1)
       x = yield tfd.Uniform(low, low+span)
       y = yield tfd.Normal(x, 1)
    ```
    In this case the joint bijector constraining `x` to its support needs the values of `low` and `low+span` to shift and scale the sigmoid bijection.
    My main pt here is that sometimes TTK can act across state parts and not only within a single state part.
6. Could we add support for "make_kernel_for_the_rest_of_the_states" where only the explicitly specified indices get a make_kernel_fn, and everything else can use (e.g.) a single TTK(HMC(..))? I guess this makes sense to defer to the future as well, alongside (3) above.
7. Can you clarify: What about `list` and `unnest` causes issues?

tensorflow_probability/python/experimental/mcmc/gibbs_kernel.py
chrism0dwk: @brianwa84 Just a few preliminary comments on the above:
1. That's a really good point, we'd thought about that too.  I'd vote for the `SamplingKernel` approach, as it simplifies `GibbsKernel` and explicitly marks the step out as a full conditional.  How about `FullConditionalKernel` which takes a `make_distribution_fn` function returning the correct distribution for that step of the MCMC chain?
2. Fair point.  I'll give this some careful thought -- probably along the lines of Fonnesbeck 2013 (which STAN uses a marginalised implementation for, but as far as I understand the original implementation mixed over a discrete space). 
3. (and 6) I guess this one's up to how the TFP semantics evolve.  For a `JointDistribution*`, the state parts map onto stochastic nodes in the respective DAG describing the model, which I think this makes sense for most applications.  The exception would be on-the-fly non-centering of models (e.g. [Xiao Li Meng's ASIS algorithm](https://www.tandfonline.com/doi/abs/10.1198/jcgs.2011.203main) or some of the partially-non-centred epidemics work such as [Neal and Roberts 2005](https://link.springer.com/article/10.1007/s11222-005-4074-7)).  What's the direction of travel with the semantics?
4. Yeah, couldn't see a way around that, for the same reasons as (5) next...
5. Unwrapping `TransformedTransitionKernel`s.  If I've understood your question correctly... Consider the sequence of Gibbs steps for single-parameter updates of `alpha`, `beta`, and `sigma` in the linear model
```
y = Normal(loc = alpha + beta*x, scale=sigma)
```
`alpha` and `beta` exist on an unconstrained space, whereas `sigma` is strictly positive and must be transformed to an unconstrained space for effective sampling.  The conditional posteriors for all parameters do not factorise into independent terms, so rather than re-calculating the `current_target_log_prob` at each Gibbs step, it's quicker just to recognise the presence of a transformed kernel and add or subtract the Jacobian term as required.  
6. (see 3)
7. `GibbsKernelResults.inner_results` is of type `list` (of inner `TransitionKernel` results namedtuples).  At the time when we wrote the code, the `unnest` recursive descent algorithm wasn't able to iterate over a list structure when it met it, and so threw an exception.  Not sure if this has changed?   I guess one possibility might be to dynamically construct a `namedtuple` for `GibbsKernelResults` which replaces the `list` with a bunch of fields. 

We'll do the easy bits first, and address (2) then (1).  

Maybe I could ask the team for their thoughts around how `state_parts` semantics might play into the future of this approach?  It seems a good time to talk about it with the commit of @ColCarroll's turnkey windowed sampling solution (which AFAIU uses restructuring and shape bijectors to concatenate state parts).

chrism0dwk: > presumably only calibrated as long as each make_kernel_fn is also calibrated?

And yes, we should try to instantiate each kernel by calling each `make_kernel_fn` and take logical AND of all `is_calibrated`

coatless:
@chrism0dwk I guess you gave up on Gibbs sampling in Tensorflow or moved onto PyTorch alternatives?

PR #1234, opened 2021-02-09
Updated ExpGamma formula to be correct

tmarkovich:
In prepping to implement the SigmoidBeta distribution, I reviewed the ExpGamma distribution to understand the rough code structure. I saw that the formula in the comments didn't match the code or the derivation using the change of variables formula. This PR makes the minor modifications to fix that. The formula now agrees with the log_prob formula.


srvasude (approved):

brianwa84:
Thanks for the fix!

PR #1242, opened 2021-02-19
Array-like equality comparison for TransformedVariable

### LINKED ISSUE #1241: TransformedVariable does not follow array convention for equality (==) comparison
A `tfp.util.TransformedVariable` generally behaves like any tensor-like object. For example, a less-than-equal comparison works element-wise, as expected:
```python
import tensorflow as tf
var = tf.Variable(1.0)
assert var <= 1.0 and var >= 1.0 and var == 1.0  # passes
t_var = tfp.util.TransformedVariable(1.0, tfp.bijectors.Identity())
assert t_var <= 1.0 and t_var >= 1.0  # passes
```
However, equality comparison does **not** behave likewise, and the following fails:
```python
assert t_var == 1.0  # FAILS
```
This is completely unexpected behaviour which can lead to very subtle downstream bugs.

The reason for this behaviour seems to be due to explicitly removing `__eq__` and `__ne__` in https://github.com/tensorflow/probability/blob/a91efeb17a6f6fdae97d3465878f5fc7fba12d30/tensorflow_probability/python/util/deferred_tensor.py#L86

What was the motivation for excluding equality comparisons? Would there be any objections to a PR that removes that line, thereby allowing tfp TransformedVariables to behave more expectedly like tf Variables?

st--:
Array-like behaviour for `TransformedVariable.__eq__` and `TransformedVariable.__ne__`.
Resolves #1241.


st--:
@googlebot I signed it!

st--:
It's clearly not quite as simple as changing this one line, but I'd really appreciate thoughts from the tensorflow-probability core developers as to what else might need changing to get the tests to pass and be able to close #1241!

PR #1243, opened 2021-02-21
Make deprecation warning clearer

abhinav-upadhyay:
The message to 'combine scale_identity_multiplier with scale_diag' seemed to suggest that the arguments `scale_identity_multiplier` and `scale_diag` should be combined together, but what is actually meant is that `scale_diag` should be used in favour of `scale_identity_multiplier`. The comment down below in the body of `__init__` was clearer.


emilyfertig:
Thanks for the clarification! I'll merge this.

PR #1254, opened 2021-02-27
Add `tfp.layers.MonteCarloDropout` layer.

### LINKED ISSUE #420: Always-on dropout layer
**Feature**

Add the ability to create "always-on" dropout layers as proposed [here](https://arxiv.org/abs/1703.04977).

**Current behavior/state.**

`tf.keras.layers.Dropout` operates differently at train and test time. It randomly drops node during training while simply passing through the inputs during testing.

Currently, implementing the suggested feature requires writing a custom dropout layer:

```py
class Dropout(tf.keras.layers.Layer):
    """Always-on dropout layer, i.e. does not respect the training flag
    set to true by model.fit but false by model.predict.
    Unlike tf.keras.layers.Dropout, this layer does not return input
    unchanged if training=true, but always randomly drops a fraction specified
    by self.size of the input nodes.
    """

    def __init__(self, rate, **kwargs):
        super().__init__(**kwargs)
        self.rate = rate

    def call(self, inputs):
        return tf.nn.dropout(inputs, self.rate)

    def get_config(self):
        """enables model.save and restoration through tf.keras.models.load_model"""
        config = super().get_config()
        config["rate"] = self.rate
        return config
```

**Who will benefit from this feature?**

Everyone using dropout to run Bayesian neural networks. Hence `tfp` users in particular may benefit from this feature.

**Additional info**

This issue is a follow up to [tf#28484](https://github.com/tensorflow/tensorflow/issues/28484).

frangaren:
Add `tfp.layers.MonteCarloDropout` layer as suggested in #420. This way, this pull request may resolve #420.

### DIFF
--- a/tensorflow_probability/python/layers/BUILD
+++ b/tensorflow_probability/python/layers/BUILD
@@ -34,6 +34,7 @@ py_library(
         ":distribution_layer",
         ":initializers",
         ":masked_autoregressive",
+        ":monte_carlo_dropout",
         ":util",
         ":variable_input",
     ],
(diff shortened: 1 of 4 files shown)


frangaren:
@googlebot I signed it!

PR #1270, opened 2021-03-16
added highway flow

gisilvs:


gisilvs:
@emilyfertig @davmre @sharadmv

gisilvs:
@googlebot I signed it!

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # set activation=False for last block
+    def __init__(self, width, activation=True, validate_args=False, name='tri_res_net'):
+        super(TriResNet, self).__init__(
+            validate_args=validate_args,
+            forward_min_event_ndims=0,
+            name=name)
+
+        self.width = width
+        self.W = tf.Variable(np.random.normal(0, 0.01, (self.width, self.width)), trainable=True, dtype=tf.float32)
```
davmre, line 26: In TFP classes we generally assume that `__init__` is a lightweight method that has no 'graph side effects', meaning that it doesn't create any new Tensors or tf.Variables. This is nice because it means it's always cheap to construct an instance, but it does put some constraints on designs.

Another TFP convention (#6 in the style guide: https://github.com/tensorflow/probability/blob/master/STYLE_GUIDE.md) is that we prefer to use descriptive, plain-English names in place of mathematical notation, even if doing so makes the code painfully verbose. For example, we'd prefer `weights` rather than `W`, `bias` rather than `b`, `diagonal` (or similar) rather than `d`, `residual_fraction` (or similar) rather than `lambda`, etc. This makes it easier for a future reader or maintainer of this code (who might not have full mathematical context) to glance at a line of code and at least have some idea of what's going on. (if needed, we can also add a comment to describe the math specifically).

Here I'd suggest that the bijector `__init__` could have a signature along the lines of `__init__(self, weights, bias, diagonal_elements, unconstrained_residual_fractions, activation_fn=tf.nn.sigmoid, validate_args=False, name='tri_res_net')`, where we pass in the variables explicitly using descriptive names (it'd also be nice to accept arbitrary activation functions or `None`, rather than just sigmoid, though that's not essential for a first pass).

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,91 @@
...
+'''
+
+
+class TriResNet(tfb.Bijector):
+
+    # set activation=False for last block
+    def __init__(self, width, activation=True, validate_args=False, name='tri_res_net'):
+        super(TriResNet, self).__init__(
+            validate_args=validate_args,
+            forward_min_event_ndims=0,
```
davmre, line 22: Assuming that the values transformed by this bijector (arguments `x` and `y` to the forward and inverse methods, respectively) are intended to be vectors, then this should be `forward_min_event_ndims=1`. (otherwise the base class will screw up your log-det-Jacobian calculations by trying to sum them).

emilyfertig (commented):
Hi Gianluigi -- Thanks for the PR, this looks like a great start! I left some initial comments.

tensorflow_probability/python/experimental/bijectors/highway_flow.py
davmre: To clarify re taking variables as `__init__` arguments: of course variables need to be created somewhere, but we'd probably move that to a wrapper function, which could have a signature close to what you have here. Something along the lines of:

```
def build_highway_flow_layer(width, activation):
  return TriResNet(
    weights=tf.Variable(np.random.normal(0, 0.01, (width, width)).astype(tf.float32)),
    bias=tf.Variable(np.random.normal(0, 0.01, (width,)).astype(tf.float32)),
    ...,
    activation=activation)
```

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,91 @@
...
+    def _inverse(self, y):
+        if self.activation:
+            y = self.inv_f(y)
+
+        y = tf.linalg.matmul(y - self.b, tf.linalg.inv(self.cu(self.get_U())))
+        y = tf.linalg.matmul(y, tf.linalg.inv(self.cu(self.get_L())))
+        return y
+
+    def _inverse_log_det_jacobian(self, y):
+        return self._forward_log_det_jacobian(y)  # TODO: is this correct?
```
davmre, line 84: If you don't implement an inverse_log_det_jacobian, the Bijector base class defaults to

inverse_log_det_jacobian(y) = - forward_log_det_jacobian(x=inverse(y))

(which is always correct), so there's no need for an explicit implementation unless you have an approach that's more efficient or more numerically stable.

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def cu(self, M):
+        # convex update
+        # same as doing as in the paper, but I guess easier to invert
+        I = tf.eye(self.width)
+        l = self.get_l()
+        return l * I + (1 - l) * M
+
+    def inv_f(self, y, N=20):
+        # inverse with Newton iteration
```
davmre, line 62: If you want to support activation functions other than sigmoid, we have root-finding methods that can generically invert scalar functions:
https://www.tensorflow.org/probability/api_docs/python/tfp/math/find_root_chandrupatla
https://www.tensorflow.org/probability/api_docs/python/tfp/math/find_root_secant

The experimental ScalarFunctionWithInferredInverse bijector https://www.tensorflow.org/probability/api_docs/python/tfp/experimental/bijectors/ScalarFunctionWithInferredInverse?version=nightly
is a wrapper around these methods that implements the bijector API, e.g., it'll automatically compute the log_det_jacobian (which for a scalar function is just the log absolute derivative) to give you an analogue of the derivative method that you wrote here for a sigmoid activation (`self.df`).

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # lambda before sigmoid
+        self.pre_l = tf.Variable(np.random.normal(0, 0.1, (1,)), trainable=True, dtype=tf.float32)
+
+        self.activation = activation
+
+        self.masku = tfe.numpy.tril(tf.ones((self.width, self.width)), -1)
+        self.maskl = tfe.numpy.triu(tf.ones((self.width, self.width)), 1)
+
+    def get_l(self):
```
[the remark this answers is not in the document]
emilyfertig: Update after reading Dave's comment: It could also work to pass predefined `TransformedVariables` into the constructor (or just not use them, if you and he prefer). For some more complex bijectors -- e.g. `tfb.Glow` -- we do define variables in the constructor, so that might be something we could consider.

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def get_U(self):
+        return self.W * self.masku + tf.linalg.diag(tf.math.softplus(self.d))
+
+    def df(self, x):
+        # derivative of activation
+        l = self.get_l()
+        return l + (1 - l) * tf.math.sigmoid(x) * (1 - tf.math.sigmoid(x))
+
+    def cu(self, M):
```
davmre, line 54: as above, we prefer descriptive, readable names: `convex_update` instead of `cu`, etc.

Also, most of these methods should probably be private (`_convex_update`, etc) unless you expect them to be used directly by external code.

davmre (commented):
Thanks for sharing this, it looks great! I've left some initial comments just to give an idea of what we'll expect in a final version.

For a full PR, we'll also eventually need:
- A file `highway_flow_test.py` with unit tests.
- Updates to the `BUILD` and `__init__.py` files to refer to the new code.
- Docstrings on all new classes and methods (can be one-line for short/trivial methods, but generally we'll want to describe the arguments and return values in a similar style to the rest of TFP).

You might want to hold off on, e.g., writing docstrings until we've converged further on the interface, but I'm happy to talk more about what any of these would look like.

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,91 @@
...
+
+        self.activation = activation
+
+        self.masku = tfe.numpy.tril(tf.ones((self.width, self.width)), -1)
+        self.maskl = tfe.numpy.triu(tf.ones((self.width, self.width)), 1)
+
+    def get_l(self):
+        return tf.nn.sigmoid(self.pre_l)
+
+    def get_L(self):
```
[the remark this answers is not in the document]
gisilvs: My initial understanding of `TransformedVariable` was that we specify our variable before the bijector, and what we get is the transformed variable according to the chosen bijector. However, it seems like it works the other way around, i.e. the initial_value should be defined as if it was already transformed by the bijector. Is my understanding correct?

If this is correct, then one way to define the matrix U could be:
```
initial_value = np.triu(np.random.normal(0., 1., (5, 5)), 1) + np.diag(tf.math.softplus(np.random.normal(0, 1, (5,))))
fill_scale_tril_bijector = tfb.FillScaleTriL(diag_bijector=tfb.Softplus(), diag_shift=None)
transpose_bijector = tfp.bijectors.Transpose(rightmost_transposed_ndims=2)
chain_of_bijectors = tfp.bijectors.Chain([transpose_bijector, fill_scale_tril_bijector])

upper_diagonal_weights_matrix = tfp.util.TransformedVariable(initial_value=initial_value,
                                                             bijector=chain_of_bijectors,
                                                             trainable=True)
```
where we obtain `initial_value` by doing the whole transformation ourselves.

Following the same procedure, then the `residual_fraction` (lambda) could be defined as:

```
self.residual_fraction = tfp.util.TransformedVariable(
            initial_value=tf.math.sigmoid(np.random.normal(0, 0.1, (1,))),
            bijector=tfb.Sigmoid(), trainable=True)
```

Is this the correct way to use `TransformedVariable`? And should the definition of the `initial_value` be done differently?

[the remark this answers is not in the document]
emilyfertig: Good questions -- you're right that `initial_value` (somewhat counterintuitively) is the initial value *after* the transformation. Since `FillScaleTriL.inverse` ignores entries above the diagonal, you can define `initial_value` just as `np.random.uniform(0., 1.), (width, width))` and the value of the `TransformedVariable` will be triangular (here, using `uniform` instead of `normal` to avoid negative values).

Instead of the `Transpose` bijector, it might be more efficient to pass `transpose_a=True` into `matvec`, would that work? (Also, it looks like the `matvec` args are reversed -- the matrix should be first, then the vector.)

`residual_fraction` looks good, except it might be simpler to define `initial_value` as a `uniform` sample (or as a fixed value -- in our implementation, this is a parameter we expose to the user, with a default of 0.5). Also, `tf.Variable`s are trainable by default, so you don't need to pass `trainable=True`.

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -1,27 +1,29 @@
...
-tfb = tfp.bijectors
-tfe = tf.experimental
+import tensorflow.compat.v2 as tf
 
+from tensorflow_probability.python import bijectors as tfb
+from tensorflow_probability.python import util
+from tensorflow_probability.python.experimental.bijectors import scalar_function_with_inferred_inverse
 
 def build_highway_flow_layer(width, residual_fraction_initial_value=0.5, activation_fn=None):
     # FIXME: should everything be in float32 or float64?
```
davmre, line 9: We typically try to infer the dtype that the user wants from the dtype of float-valued inputs. In this case you could write at the top:

```
residual_fraction_initial_value = tf.convert_to_tensor(residual_fraction_initial_value,
                                                       dtype_hint=tf.float32,
                                                       name='residual_fraction_initial_value')
dtype = residual_fraction_initial_value.dtype
```

This will use float32 by default, but adapt to float64 / float16 / etc. if the caller passes a Tensor value with explicit dtype (`convert_to_tensor` is a no-op if the input is already a Tensor). Then use `dtype` throughout when defining variables, e.g., 

`bias=tf.Variable(tf.random.normal([width], mean=0., stddev=0.01, dtype=dtype))`

etc.

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -1,27 +1,29 @@
...
     # TODO: add control that residual_fraction_initial_value is between 0 and 1
     return HighwayFlow(
         width=width,
-        residual_fraction=tfp.util.TransformedVariable(
+        residual_fraction=util.TransformedVariable(
             initial_value=np.asarray(residual_fraction_initial_value, dtype='float32'),
             bijector=tfb.Sigmoid()),
         activation_fn=activation_fn,
         bias=tf.Variable(np.random.normal(0., 0.01, (width,)), dtype=tf.float32),
```
davmre, line 17: We've been moving in TFP towards using stateless RNGs, which are required by JAX and now also supported by TF, and tend to be easier to reason about (see https://github.com/tensorflow/probability/blob/master/PRNGS.md if you want to learn more). The idiom is along the lines of:

```
# At top

# Function defn
def build_highway_flow_layer(..., seed=None):
   # Ensure that each part gets a different initialization.
   bias_seed, upper_seed, lower_seed = samplers.split_seed(seed, n=3)
   ...
   # Drawing initial values.
   bias = tf.Variable(samplers.normal([width], mean=0., stddev=0.01, dtype=dtype, seed=bias_seed)),
   upper_diagonal_weights_matrix = TransformedVariable(samplers.normal(..., seed=upper_seed), ...),
   lower_diagonal_weights_matrix = TransformedVariable(samplers.normal(..., seed=lower_seed), ...))
```

davmre (changes requested):

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,125 @@
...
+        lower_diagonal_weights_matrix=util.TransformedVariable(
+            initial_value=samplers.normal((width, width), 0., 1., seed=lower_seed),
+            bijector=tfb.Chain([tfb.TransformDiagonal(diag_bijector=tfb.Shift(1.)),
+                                tfb.Pad(paddings=[(1, 0), (0, 1)]),
+                                tfb.FillTriangular()]),
+            dtype=dtype)
+    )
+
+
+class HighwayFlow(tfb.Bijector):
```
davmre, line 41: I think I figured out how FFJORD manages to cache the fldj/ildj computations from `_augmented_forward` / `_augmented_inverse`. You'll need to add a line to the HighwayFlow class similar to this one from FFJORD:

```
_cache = cache_util.BijectorCacheWithGreedyAttrs(
      forward_name='_augmented_forward',
      inverse_name='_augmented_inverse')
```
(with a comment similar to what's in https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/bijectors/ffjord.py#L246-L252)

Internally, calling `bijector.forward()` invokes `bijector._cache.forward()`, so this line sets up a cache that knows to call `_augmented_forward` and remember both of the results that it returns.

tensorflow_probability/python/experimental/bijectors/highway_flow_test.py
```diff
@@ -0,0 +1,127 @@
...
+
+            bijector = tfp.experimental.bijectors.build_highway_flow_layer(width, activation_fn=tf.nn.softplus)
+            self.evaluate([v.initializer for v in bijector.trainable_variables])
+            self.assertStartsWith(bijector.name, 'highway_flow')
+            self.assertAllClose(x, bijector.inverse(tf.identity(bijector.forward(x))))
+            self.assertAllClose(bijector.forward_log_det_jacobian(x, event_ndims=dim+1),
+                                -bijector.inverse_log_det_jacobian(tf.identity(bijector.forward(x)), event_ndims=dim+1))
+
+    def testJacobianWithActivation(self):
+        #activations = ['sigmoid', 'softplus', 'tanh', 'none']
```
davmre, line 46: as elsewhere, we can delete the commented code until we're ready to actually support this.

tensorflow_probability/python/experimental/bijectors/highway_flow_test.py
```diff
@@ -0,0 +1,127 @@
...
+                width=width,
+                residual_fraction=residual_fraction,
+                activation_fn=activation_fn,
+                bias=tf.Variable(0., dtype=dtype),
+                upper_diagonal_weights_matrix=tf.Variable(tf.eye(width)),
+                lower_diagonal_weights_matrix=tf.Variable(tf.eye(width))
+            )
+
+            self.evaluate([v.initializer for v in bijector.trainable_variables])
+            x = tf.ones((batch_size, width)) # * samplers.uniform((batch_size, width), -10., 10., seed=seed)
```
davmre, line 73: I have a very slight preference for using (seeded) random samples here instead of ones (here and elsewhere).

tensorflow_probability/python/experimental/bijectors/highway_flow_test.py
```diff
@@ -0,0 +1,127 @@
...
+        eps = 1e-6
+
+        bijector.residual_fraction = residual_fraction + h
+        y1 = tf.reduce_mean(target.log_prob(bijector.forward(x)))
+        bijector.residual_fraction = residual_fraction - h
+        y2 = tf.reduce_mean(target.log_prob(bijector.forward(x)))
+
+        manual_grad = (y1 - y2) / (2 * h)
+
+        diff = tf.math.abs(tf_grad - manual_grad) / tf.reduce_max((eps, tf.math.abs(tf_grad) + tf.math.abs(manual_grad)))
```
davmre, line 122: It's probably clearer (and easier) to write just `self.assertAllClose(tf_grad, manual_grad, rtol=1e-4)` in place of this explicit calculation.

tensorflow_probability/python/experimental/bijectors/highway_flow_test.py
```diff
@@ -0,0 +1,127 @@
...
+        batch_size = 3
+        dtype = tf.float32
+        residual_fraction = tf.Variable(0.5)
+        bijector = tfp.experimental.bijectors.HighwayFlow(
+            width=width,
+            residual_fraction=residual_fraction,
+            activation_fn=tf.nn.softplus,
+            bias=tf.Variable(0., dtype=dtype),
+            upper_diagonal_weights_matrix=tf.Variable(tf.eye(width)),
+            lower_diagonal_weights_matrix=tf.Variable(tf.eye(width))
```
davmre, line 103: I don't think any of these parameters (including residual_fraction) need to be Variables, since you're not actually updating them. You can just write `residual_fraction = tf.constant(0.5)`,
`upper_diagonal_weights_matrix=tf.eye(width)`, etc. Variables are special in that they're automatically watched by gradient tapes, but you don't need that since you're already asking the tape to watch `residual_fraction` specifically.

tensorflow_probability/python/experimental/bijectors/__init__.py
```diff
@@ -17,9 +17,13 @@
 from tensorflow_probability.python.bijectors.ldj_ratio import inverse_log_det_jacobian_ratio
 from tensorflow_probability.python.experimental.bijectors.distribution_bijectors import make_distribution_bijector
 from tensorflow_probability.python.experimental.bijectors.scalar_function_with_inferred_inverse import ScalarFunctionWithInferredInverse
+from tensorflow_probability.python.experimental.bijectors.highway_flow import build_highway_flow_layer, HighwayFlow
```
davmre, line 20: Google's Python style prefers only one import per line, even if that means redundant lines. Here, you'll need to write:

```
from tensorflow_probability.python.experimental.bijectors.highway_flow import build_highway_flow_layer 
from tensorflow_probability.python.experimental.bijectors.highway_flow import HighwayFlow
```

davmre (changes requested):
Hi Gianluigi, thanks for the updates; this looks great! Almost all of my comments at this point are nitpicking about stylistic details---unfortunately, as a new contributor you're subject to some up-front overhead in conforming to our (somewhat arbitrary) style rules. Hopefully it's not too bad; I can attest that if you write enough TFP code it eventually becomes mostly automatic . :-)

Other than my specific comments, I think the code and tests look to be in generally good shape, so I think we're really close to being able to pull this in. Let me know if anything I said doesn't make sense (this is very likely), or if there's anything else I can help with.

Dave

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,189 @@
...
+    """Implements an Highway Flow bijector [1], which interpolates the input X with the transformations at each step of the bjiector.
```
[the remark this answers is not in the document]
gisilvs: Should the `backticks` be used also in the Args description in the `__init__()` function?

tensorflow_probability/python/experimental/bijectors/highway_flow_test.py
gisilvs: I have changed this but then to pass the test I needed to increase h to 1e-3. Could you please double-check that the gradient test I implemented is reliable?

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,376 @@
...
+      y = self._inverse_of_softplus(y)
+      ildj -= tf.reduce_sum(tf.math.log(self._derivative_of_softplus(y)),
+                            -1)
+
+    y = y - tf.concat([(1. - self.residual_fraction) * tf.ones(
+      self.gate_first_n), tf.ones(self.width - self.gate_first_n)],
+                      axis=0) * self.bias
+    y = tf.linalg.triangular_solve(tf.transpose(
+      self._convex_update(self.upper_diagonal_weights_matrix)),
+      tf.linalg.matrix_transpose(y),
```
davmre, line 346: Why are we transposing `y` here? Transposes are expensive, since they actually change the data layout in memory---as opposed to other ops like reshape, expand_dims, squeeze, etc., which just change metadata---so it's good to avoid them when possible. (also it just feels weird to be transposing `y` since it's conceptually a vector).

I think you can remove all `transpose` and `matrix_transpose` calls from this method, using a combination of:

1) Work with `y` of shape `[..., width, 1]` throughout, so that you can feed it directly to `triangular_solve` calls. That is, instead of `y = expand_dims(y, 0)` at the beginning, do `y = y[..., tf.newaxis]` to add the extra dimension at the end rather than at the beginning. (as a bonus, you can do this unconditionally so you don't need the `added_batch` logic any more). You might need to sprinkle in a few more `[..., tf.newaxis]` expansions for other Tensors, eg `bias`, to get the dimensions to line up, but this is costless.

2. Use the `adjoint` flag to `triangular_solve`. Since transposes are expensive, most TF matrix ops give you the option to treat a matrix as transposed without explicitly constructing the transpose. For example, `triangular_solve(A, B, adjoint=True)` is equivalent to, and more efficient than, `triangular_solve(tf.linalg.matrix_transpose(A), B)`.

tensorflow_probability/python/experimental/bijectors/highway_flow_test.py
```diff
@@ -0,0 +1,149 @@
...
+
+@test_util.test_all_tf_execution_regimes
+class HighwayFlowTests(test_util.TestCase):
+
+  def testBijector(self):
+    width = 1
+    for dim in range(2):
+      if dim == 0:
+        # Test generic case with scalar input
+        x = tf.ones((width,)) * samplers.uniform((width,), minval=-1.,
```
davmre, line 34: Multiplying the samples by `tf.ones` here doesn't do anything; you can remove it (also in similar lines elsewhere).

tensorflow_probability/python/experimental/bijectors/highway_flow_test.py
```diff
@@ -0,0 +1,149 @@
...
+    with tf.GradientTape() as g:
+      g.watch(bijector.residual_fraction)
+      y = tf.reduce_mean(target.log_prob(bijector.forward(x)))
+    tf_grad = g.gradient(y, bijector.residual_fraction)
+
+    h = 1e-3
+
+    # pylint: disable=protected-access
+    bijector._residual_fraction = residual_fraction + h
+    y1 = tf.reduce_mean(target.log_prob(bijector.forward(x)))
```
davmre, line 138: I think the bijector cache will give a wrong answer here (and two lines below), since it doesn't know that `_residual_fraction` has changed from the previous call. You'll probably need to pass `tf.identity(x)`.

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,376 @@
...
+    gate_first_n=gate_first_n
+  )
+
+
+class HighwayFlow(tfb.Bijector):
+  """Implements an Highway Flow bijector [1].
+
+  HighwayFlow interpolates the input `X` with the transformations at each step
+  of the bjiector. The Highway Flow can be used as building block for a
+  Cascading flow [1 or as a generic normalizing flow.
```
davmre, line 106: nit: [1]

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,376 @@
...
+
+  The transformation consists in a convex update between the input `X` and a
```
davmre, line 108: nit: 'consists in' -> 'consists of'

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,376 @@
...
+
+    # Log determinant term from the upper matrix. Note that the log determinant
+    # of the lower matrix is zero.
+
+    added_batch = False
+    if len(x.shape) <= 1:
+      if len(x.shape) == 0:
+        x = tf.reshape(x, -1)
+      added_batch = True
+      x = tf.expand_dims(x, 0)
```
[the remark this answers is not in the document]
gisilvs: I think I had a problem when using the BrownianMotion from inference gym, where each prior had an empty shape []. But with the tf.newaxis seems to be fixed without this logic.

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,376 @@
...
+  # updating them only when the LDJ methods themselves are called).
+
+  _cache = cache_util.BijectorCacheWithGreedyAttrs(
+    forward_name='_augmented_forward',
+    inverse_name='_augmented_inverse')
+
+  def __init__(self, residual_fraction, activation_fn, bias,
+               upper_diagonal_weights_matrix,
+               lower_diagonal_weights_matrix,
+               gate_first_n,
```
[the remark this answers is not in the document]
davmre: Thanks for doing this! Can you also
1) Make `gate_first_n=None` the default in the signature, and
2) Update the docstring to document the default behavior.
?

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,395 @@
...
+
+    Returns:
+      x after forward flow and a dict containing forward and inverse log
+      determinant of the jacobian.
+    """
+
+    # Log determinant term from the upper matrix. Note that the log determinant
+    # of the lower matrix is zero.
+
+    fldj = tf.zeros(x.shape[:-1], dtype=self.dtype) + tf.reduce_sum(
```
davmre, line 297: `x.shape` should be `ps.shape(x)` (also in the analogous line of `_augmented_inverse`)

tensorflow_probability/python/experimental/bijectors/highway_flow.py
```diff
@@ -0,0 +1,395 @@
...
+      self._convex_update(self.upper_diagonal_weights_matrix),
+      x, transpose_a=True)
+    x += (tf.concat([(1. - self.residual_fraction) * tf.ones(
+      self.gate_first_n, dtype=self.dtype),
+                     tf.ones(self.num_ungated, dtype=self.dtype)],
+                    axis=0) * self.bias)[tf.newaxis, ...]
+
+    if self.activation_fn:
+      fldj += tf.reduce_sum(tf.math.log(self._derivative_of_softplus(x[0])),
+                            -1)
```
davmre, line 320: nit: prefer the keyword argument, `axis=-1`

(similarly elsewhere)

davmre (changes requested):

tensorflow_probability/python/experimental/bijectors/highway_flow.py
gisilvs: For the comments on the default behaviour, I have added a simple line for the HighwayFlow `__init__`, but maybe you meant something more sophisticated?

PR #1276, opened 2021-03-21
Fix broken link to Random Walk Metropolis paper

JulianWgs:
Alternatively the direct link to the pdf: https://projecteuclid.org/journalArticle/Download?urlid=10.1214%2F154957804100000024

I think however the DOI link will be much more future proof.


JulianWgs:
@googlebot I signed it!

PR #1280, opened 2021-03-23
add casting to backend for Owen's T function

j-wilson:
When compiled using `tf.function`, Owen's T function currently fails for `tf.float64` typed inputs. By way of example,
```python
import tensorflow as tf
from tensorflow_probability.python.math.special import owens_t

h = tf.cast(1, tf.float64)
a = tf.cast(1, tf.float64)

@tf.function
def closure():
  return owens_t(h, a)

T = closure()  # fails
```
Fortunately, this looks to be a quick fix. The fallback values used in two `tf.where` statements simply need to be wrapped by the `numpy_dtype` helper.

### DIFF
--- a/tensorflow_probability/python/math/special.py
+++ b/tensorflow_probability/python/math/special.py
@@ -1748,7 +1748,7 @@ def _owens_t_naive_gradient(h, a):
        tf.math.erfc(modified_h / np.sqrt(2.))), result)
 
   # When a = 0, we should return 0.
-  result = tf.where(tf.math.equal(modified_a, 0.), 0., result)
+  result = tf.where(tf.math.equal(modified_a, 0.), numpy_dtype(0.), result)
 
   normh = tf.math.erfc(h / np.sqrt(2.))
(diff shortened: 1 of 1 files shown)


j-wilson:
@googlebot I signed it!

srvasude (approved):

srvasude (commented):
I'm a little confused here. Specifically, the tests use float64 / float32, so should have caught this.

j-wilson:
> I'm a little confused here. Specifically, the tests use float64 / float32, so should have caught this.

@srvasude Do the tests compile Owen's T as a `tf.function`? From what I can tell, they appear to be run in eager mode. When compiled however, `tf.where` convert primitives. As a result,
```python
indicator = lambda x: tf.where(x, 1, tf.cast(0, tf.float64))
indicator(True)  # this works fine
tf.function(indicator)(True)  # this fails
```

davmre:
I went ahead and pulled this in, along with updates to the tests so that they run in both graph and eager mode (which all TFP tests should, so this was also an oversight). Thanks for the fix!

PR #1285, opened 2021-03-29
Fix BatchNormalization docstring

hartikainen:
The current `BatchNormalization` docstring is broken in several ways. There's a syntax error on line https://github.com/tensorflow/probability/blob/43a9d6c3f5992a24ddf7aa3fa826a038d70697c5/tensorflow_probability/python/bijectors/batch_normalization.py#L95

and the initializers both for `tfd.Normal` and `tfd.MultivariateNormalDiag` are currently broken. This should make the example runnable (save imports).


axch (approved):
Thanks for the catch!  Sorry it took us so long to get to this.  I'll pull it in now.

Issue #1273, opened 2021-03-18
Contributing for an absolute beginner?


mikeevmm:
Hello all,

I have recently implemented the [bounded version of the L-BFGS-B][0] in (a local copy of) Tensorflow Probability.

It's not a blazing fast implementation (it is outright hacky in some places --- e.g., how it calculates <math>M</math> for different histories), and I had to reduce scope in order to manage (I assumed the inputs were `float32`, and there was only one batching dimension; this fit my use case well). Despite this, L-BFGS-B is incredibly useful when employed in, e.g., periodic functions, and the code behaves correctly; it correctly converges for a high-dimensional quadratic surface as well as for a Rosenbrock surface (with a sufficiently large "box"), but converges to the same wrong minimum as the [original Fortran implementation][1] (`(5., 5.)`) when considering the Rosenbrock function and bounds `[(5, 20), (-10, 5)]`.

I'd like to contribute this implementation to TensorFlow Probability, but I've read the [`CONTRIBUTING.md`][2] page (and related links) and am still at a lost. My main concerns are the following:

1. Do I have to support arbitrary dtypes and arbitrary batching dimensions?
2. Since I am not contributing a new module but would rather be putting the code into the `optimizer` module, how does this affect the recommendations?

It's also fairly difficult to grapple with all the information in `CONTRIBUTING.md` without being familiar with TensorFlow's internals. Is it possible to request a code review so I have something to go off of?

Thanks in advance

[0]: https://dl.acm.org/doi/10.1145/279232.279236
[1]: https://github.com/scipy/scipy/blob/v1.6.0/scipy/optimize/lbfgsb_src/lbfgsb.f
[2]: https://github.com/tensorflow/probability/blob/master/CONTRIBUTING.md

davmre:
Hi Miguel, this sounds like a contribution we'd be happy to have (and those sound like good unit tests!). We do need to support arbitrary dtypes, and ideally multiple batch dimensions (though now that `tf.vectorized_map` exists it's becoming more feasible to work around this), but typically these are pretty straightforward and we have a fair amount of experience with idioms for making them work, so it's something we could cover in a code review if you're willing to iterate a bit. The best way to request review is to just create a new PR.

mikeevmm:
(Thank you @davmre for the prompt response; I haven't forgotten and will make a PR asap, but I'm fighting some deadlines right now :) )

mikeevmm:
Checking in, the pull-request has been filed ( #1290 ).

[...]

PR #1290, opened 2021-04-04
feat: implementation of bounded LBFGS algorithm

mikeevmm:
This commit implements the bounded version of the Limited-memory Broyden–Fletcher–Goldfarb–Shanno optimization algorithm, per [1], in `lbfgsb.py`. It is lacking unit tests (which have been run outside of the committed changes), because I don't understand how to properly implement them alongside, e.g., the LBFGS unit tests. The following unit tests were ran:

- Converges to a 60-dimensional quadratic bowl
- Converges to the correct minimum of the Rosenbrock surface given sufficiently far-away bounds
- Converges to `[5., 5.]` in the Rosenbrock surface, when given bounds `[(5, 20), (-10, 5)]`, which matches the [original Fortran implementation](https://github.com/scipy/scipy/blob/v1.6.0/scipy/optimize/lbfgsb_src/lbfgsb.f).

I now need help making sure the code is up to the style guide, per issue #1273 .

[2] Richard H. Byrd, Peihuang Lu, Jorge Nocedal, & Ciyou Zhu (1995).
      A Limited Memory Algorithm for Bound Constrained Optimization
      SIAM Journal on Scientific Computing, 16(5), 1190–1208.
      https://doi.org/10.1137/0916069

### DIFF
--- a/tensorflow_probability/python/optimizer/__init__.py
+++ b/tensorflow_probability/python/optimizer/__init__.py
@@ -27,6 +27,7 @@ from tensorflow_probability.python.optimizer.bfgs_utils import converged_any
 from tensorflow_probability.python.optimizer.differential_evolution import minimize as differential_evolution_minimize
 from tensorflow_probability.python.optimizer.differential_evolution import one_step as differential_evolution_one_step
 from tensorflow_probability.python.optimizer.lbfgs import minimize as lbfgs_minimize
+from tensorflow_probability.python.optimizer.lbfgsb import minimize as lbfgsb_minimize
 from tensorflow_probability.python.optimizer.nelder_mead import minimize as nelder_mead_minimize
 from tensorflow_probability.python.optimizer.nelder_mead import nelder_mead_one_step
 from tensorflow_probability.python.optimizer.proximal_hessian_sparse import minimize as proximal_hessian_sparse_minimize
@@ -42,6 +43,7 @@ _allowed_symbols = [
     'differential_evolution_minimize',
     'differential_evolution_one_step',
     'lbfgs_minimize',
+    'lbfgsb_minimize',
     'nelder_mead_minimize',
     'nelder_mead_one_step',
     'proximal_hessian_sparse_minimize',
--- a/tensorflow_probability/python/optimizer/lbfgsb.py
+++ b/tensorflow_probability/python/optimizer/lbfgsb.py
@@ -1,0 +1,1625 @@
+# Copyright 2018 The TensorFlow Probability Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


mikeevmm:
@googlebot I signed it!

davmre:
Hi Miguel, thanks very much for this PR. This is a really nice contribution, and it's great to see such thorough and well-documented code. I made a quick pass with a few minor style contents, but overall this looks to be in quite good shape.

I do have a high-level concern around the use of ragged Tensors. As you note, they can be pretty finicky, they make it tricky to work with batches, and they raise compatibility issues with, e.g., our JAX backend. It's not a dealbreaker, but if you're willing, I'd like to try to help think through whether another approach could work. I haven't fully worked through the code yet---to help me understand, can you explain what the ragged Tensors are being used for, and why they need to be ragged?

Before checking this in, we'll need to add some tests, along with corresponding entries in the BUILD file (you can use the entries for `lbfgs` and `lbfgs_test` as a model). I don't think the tests need to interact with the LBFGS tests, necessarily (although feel free to copy any that seem relevant), but at minimum we need a file `lbfgsb_test.py` that follows the pattern of `lfbgs_test` and other tests:
1. It should define a test class that inherits from tensorflow_probability.python.internal.test_util.TestCase
2. This class should be decorated with `@test_util.test_all_tf_execution_regimes` (i.e., runs tests in graph and eager modes)
3. The bottom of the file should have the the magic lines `if __name__ == '__main__': tf.test.main()`.
4. Unit tests are implemented as methods of the test class whose names start with `test`.

It sounds like you already have some tests, so if you could adapt them to add to this PR that'd be a great start. Is there anything in particular you're not sure about? I'm happy to try to help figure out the best way to set things up.

- Dave

mikeevmm:
Hello,

Thank you for the kind words. I am again a bit swamped, so I will return to this when I can, but I have read all of your comments and have no objections.

Re. ragged tensors: they were more in use (and a greater source of problems) previously, as I was keeping track of `free_vars_idx` and other free-variable-related tensors with ragged arrays. I eventually replaced those with tensors with flag values, so it should no longer be a problem.

Right now, as far as I can tell and remember, ragged tensors are only in use in lines 576-590, to transpose differently between batches, and since these are immediately `to_tensor`ed, I don't expect it to be a problem (but really I wouldn't know).

I have also found that the current approach evaluates the cost function far far more often than the Fortran implementation. Depending on how costly this is, it can be a real bottleneck (this turned out to be the case for me). I am working on implementing the direct primal method to circumvent the double use of `hanger_zhang_line_search`.

srvasude:
Hi!
  I was curious if you had time to investigate addressing / implementing these issues for LBFGS-B. We have some internal use cases for it, so I think we would be excited to have it! I am also happy to help / use this PR as a foundation for checking in an implementation for LBFGS-B.

Thank you!
   Srinivas

mikeevmm:
Hello @srvasude,

I've uploaded the more recent changes I've made to the code.

It remains to write unit tests, and modifying the code to accept more batch dimensions. The modifications of commit 41b2267 should help a lot with this latter point, because the queue no longer has the batch dimensions "in the middle", and are instead leading. This means a lot of the stuff that's using `einsum` and `...` should already work out of the box with many batch dimensions. Regardless, the code still needs some adapting for that to work.

(If I recall correctly, 41b2267 also does away completely with ragged tensors.)

Commit 64d30a4 is meant to be a starting point for writing the unit tests. However, it would maybe be best to implement multiple batching dimensions before writing the tests.

Miguel

PR #1297, opened 2021-04-08
Truncated normal quantile

### LINKED ISSUE #1120: No implementation of quantile method for TruncatedNormal
Would it be possible to implement?

harahu:
- Add a naive implementation of the truncated normal quantile function. Fixes #1120
- Fix some dead links in the style guide.


harahu:
@srvasude I followed your advice in #1120, with some help from [stats.stackexchange](https://stats.stackexchange.com/questions/490877/what-is-the-ppf-of-the-truncated-normal-distribution/491045?noredirect=1#comment907020_491045).

As mentioned in the `#TODO` I also added, this implementation might not be optimal from the perspective of numerical stability, but I feel it adds value over it not being implemented.

STYLE_GUIDE.md
```diff
@@ -6,10 +6,11 @@ number of errors, and promote consistency.
 
 ## TensorFlow Style
 
-Follow the [TensorFlow style
-guide](https://www.tensorflow.org/community/style_guide) and [documentation
-guide](https://www.tensorflow.org/community/documentation), and in particular
-the [TensorFlow code
+Follow the [TensorFlow style
```
srvasude, line 9: Revert changes in this file?

tensorflow_probability/python/distributions/truncated_normal.py
```diff
@@ -420,6 +420,19 @@ def _mode(self):
     bc_loc = tf.broadcast_to(loc, shp)
     return tf.clip_by_value(bc_loc, low, high)
 
+  def _quantile(self, p):
+    # TODO: This implementation is analytically correct, but might not
+    #  perform well in all cases. See
+    #  https://en.wikipedia.org/wiki/Truncated_normal_distribution#Generating_values_from_the_truncated_normal_distribution
+    #  for a discussion on alternatives.
+    loc, scale, low, high = self._loc_scale_low_high()
+    alpha = (low - loc) / scale
```
srvasude, line 429: We generally avoid greek for variable names. How about `std_low` and `std_high`?

STYLE_GUIDE.md
harahu: Are you sure? The old links were dead. These seem to be working, at least the last time I checked.

tensorflow_probability/python/distributions/truncated_normal.py
harahu: I'd love some clearer names for these, but I am unsure what would be appropriate. These are the same alpha and beta that are defined here: https://en.wikipedia.org/wiki/Truncated_normal_distribution

To me they just seem like convenient placeholders allowing for a more compact expression. What is your motivation for `std_low` and `std_high` in particular?

srvasude: `std_low` and `std_high` are used elsewhere in this file for this expression. It is also expressing that these are standard scores for `low` and `high`.

harahu: I renamed the variables given the naming scheme you suggested.

STYLE_GUIDE.md
harahu: Or would you want this in a separate PR? Can do that too.

ColCarroll: I think just neater if this is in a separate PR (thanks for catching this!)

harahu: I dropped these changes from this PR. Will make a separate one for these.

harahu: New PR: https://github.com/tensorflow/probability/pull/1332

harahu:
Looks like the linting failure isn't caused by my changes.

ColCarroll (approved):
Looks good -- thank you!

PR #1300, opened 2021-04-12
Update setup.py

brianwa84:
JAX 0.2.12 removed custom_transforms.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -105,7 +105,7 @@ setup(
     ],
     keywords='tensorflow probability statistics bayesian machine learning',
     extras_require={  # e.g. `pip install tfp-nightly[jax]`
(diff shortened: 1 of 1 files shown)


NeilGirdhar:
Isn't this already fixed on master?

brianwa84:
It is, but we are considering cutting another stable release for jax users
since the latest stable is broken.

NeilGirdhar:
@brianwa84 Yeah, sorry for the noise, I noticed too late that you were merging into an older version and (quickly) deleted my comment.

As a JAX user, I end up reinstalling tensorflow-probability (using `tfp_nightly`) a lot, so I'm looking forward to a new stable release :)

PR #1303, opened 2021-04-13
Upgrade setuptools for OSS testing

ColCarroll:
OSS tests have been failing, and it looks like it is due to an old version of setuptools.


SiegeLordEx:
We merged this internally at some point.

PR #1343, opened 2021-05-26
fix decompose_forecast_by_component on LinearRegression

### LINKED ISSUE #1342: decompose_forecast_by_component on LinearRegression
When executing `decompose_forecast_by_component`, if you include LinearRegression in the sts model, only LinearRegression's counting steps start from zero of a design matrix.
It means forecasting plots start from the first step of training data.

output_3ZD2faZ3ZKq0_0
(This is from https://www.tensorflow.org/probability/examples/Structural_Time_Series_Modeling_Case_Studies_Atmospheric_CO2_and_Electricity_Demand)

So, I want to propose changing `decompose_forecast_by_component` to require training data to extract an offset step to make sure counting starts from the first of the forecast data.
Then the offset step is given to `initial_step` of `model.make_component_state_space_models` .

It would be better if the sts model had the information of training steps inside itself after training is done.

europeanplaice:
Fixes #1342 

When executing "decompose_forecast_by_component", if you include LinearRegression in the sts model, only LinearRegression's counting steps start from zero of a design matrix.
It means forecasting plots start from the first step of training data.

So, I changed decompose_forecast_by_component to require training data to extract an offset step to make sure counting starts from the first of the forecast data.


europeanplaice:
@googlebot I signed it!

tensorflow_probability/python/sts/decomposition.py
```diff
@@ -221,7 +221,7 @@ def decompose_by_component(model, observed_time_series, parameter_samples):
         model, posterior_means, posterior_covs, parameter_samples)
 
 
-def decompose_forecast_by_component(model, forecast_dist, parameter_samples):
+def decompose_forecast_by_component(model, forecast_dist, parameter_samples, observed_time_series):
```
davmre, line 224: Adding a new argument would be a breaking change, and I'm not sure we actually need to do it. Would it work to pass `initial_step=forecast_lgssm.initial_step` below?

tensorflow_probability/python/sts/decomposition.py
```diff
@@ -39,7 +39,7 @@ def _split_covariance_into_marginals(covariance, block_sizes):
 
 
 def _decompose_from_posterior_marginals(
-    model, posterior_means, posterior_covs, parameter_samples):
+    model, posterior_means, posterior_covs, parameter_samples, initial_step=0):
```
davmre, line 42: Would you mind adding an entry for `initial_step` to the docstring as well? For example:

```initial_step: optional `int` specifying the initial timestep of the decomposition.```

davmre (changes requested):
Thanks for catching this, and for the fix! Just a couple of small comments.

tensorflow_probability/python/sts/decomposition.py
europeanplaice: Thank you for you review. I agree with it and changed the code as suggested.

tensorflow_probability/python/sts/decomposition.py
europeanplaice: I added an entry to the docstring.

davmre:
Thanks for the changes! I think we can go ahead and merge this. Would you mind removing the extraneous file `output_3ZD2faZ3ZKq0_0.png` from this PR? I unfortunately don't have a good way to do that on our end (it causes an error with the tool we use to pull in changes from Github).

europeanplaice:
Sorry, I deleted `output_3ZD2faZ3ZKq0_0.png`. Is it OK for you?

davmre:
Looks great; thanks! This should get pulled in soon.

PR #1349, opened 2021-06-03
gaussian binary tree inference_gym collider model

gisilvs:


gisilvs:
@davmre

spinoffs/inference_gym/inference_gym/targets/gaussian_binary_tree.py
```diff
@@ -0,0 +1,103 @@
...
+Root = tfd.JointDistributionCoroutine.Root
+
+# coupling link could be e.g. tf.nn.tanh
+def gaussian_binary_tree_prior_fn(num_layers, initial_scale, nodes_scale,
+                                  coupling_link=None):
+  initial_loc = 0.
+  nodes = []
+  # in the "root" layer (or inverse root, as it is a reversed tree) we have
+  # 2**num_layers nodes (with depth 2 --> 4 nodes, depth 4 --> 16 nodes)
+  for i in range(2 ** num_layers):
```
davmre, line 21: It would be more efficient in general to write each layer as a single distribution with batch shape:

```
layer = yield Root(tfd.Normal(loc=initial_loc * tf.ones([2 ** num_layers]), scale=initial_scale, name='layer_1')
for l in range(num_layers, 1, -1):
  layer = coupling_link(layer) if coupling_link else layer
  layer = yield tfd.Normal(loc=layer[:-1:2] - layer[1::2], scale=nodes_scale, name='layer_{}'.format(l))
```

We'd need to be sure that the CF code does the right thing on batched distributions (which *should* be treated equivalently to a list of independent dists), but we'd need to do that anyway.

Issue #1129, opened 2020-10-12
Add support for multivariate structural time series


bgroenks96:
Are there any plans or existing prototypes for extending the TFP STS API to handle multivariate observations?

I must admit to being somewhat new to STS, but I don't see any obvious theoretical limitations. There seems to be existing literature on the subject.

I would also be interested in possibly contributing to this extension!

brianwa84:
We don't have any near term plans to add support, but acknowledge it could
be very useful. I believe the underlying
tfp.distributions.LinearGaussianStateSpaceModel already supports
multivariate observations, so it would be a matter of adding support
throughout tfp.sts. This could be a little tricky, as the code is likely
assuming scalar observations in many places.

bgroenks96:
Yes, `LinearGaussianStateSpaceModel` already supports it; that's why I figured it wouldn't be (too) hard... but I'm sure there will be problems. The trick would be how to set up the observation matrix in the higher dimensional cases, I guess.

From an API standpoint, do you think it would make more sense to generalize the existing STS types to multiple dimensions or to create a new `MultivariateStructuralTimeSeries` class of types?

brianwa84:
I'd defer to @davmre .

My hunch is that it might be easier to at least make a `Multivariate*` implementation class to correspond with each current scalar class e.g. `tfp.sts.LinearRegression` gets a `MultivariateLinearRegression` look-alike. As to how that's exposed to the user, perhaps we could retain existing naming but defer to the alternative subclass with `__new__` handling a kwarg like `multivariate_observations=True` (`tfp.sts.LinearRegression(..., multivariate_observations=True)`)?

There are likely some performance penalties we would not want to pay in the non-multivariate case (e.g. cholesky can be slower than sqrt, on a 1x1 matrix).

davmre:
I think I agree with Brian that it might be cleaner to separate out multivariate components in the class hierarchy---that'd at least make the shape semantics easier to disambiguate. I think there's enough here that it'd be worth maybe doing some simple prototyping and/or writing up a few paragraphs about potential designs before diving into writing a ton of code.  A couple of questions I'd suggest thinking about:

[...]

PR #1356, opened 2021-06-08
[WIP] Update sts.internal.util.py `empirical_statistics` to accept multivariate time series.

europeanplaice:
Related to https://github.com/tensorflow/probability/issues/1129

Before implementing support for multivariate structural time series, utilities are necessary to be updated.
As a first step, I want to propose an update of `empirical_statistics` to accept multivariate time series.

But I haven't implemented a procedure when mask is not none.
It seems to me that `moments_of_masked_time_series` is simple function to compute `mean` and `std`.
But concerning covariance, I'm not sure what is a suitable way to compute.

I also added `event_axis` to decide the shape of the distribution. Is it redundant?

I welcome your opinions.
Thank you.


davmre:
Thanks for looking into this! Multivariate models are a really common request and supporting them would be a big improvement to STS.

Before making low-level code changes, I think it'd be worth answering some questions about the overall plan for multivariate series. For example:

- How do we expect users to specify multivariate models, and what classes of models can we expect to support? We probably want at least a couple of examples of user-level code that we're trying to enable.

- Can we manually build a couple of these models as a proof of concept? That is: STS is basically a machine to automatically construct instances of `tfp.distributions.LinearGaussianStateSpaceModel` over univariate time series. If we want a machine to automatically construct distributions over multivariate time series, it's helpful to first try writing some examples of the distributions we want, to get a sense of what parts will be needed, before we work too hard at automating the process. This doesn't have to be pretty; it could just be a [colab notebook](https://colab.research.google.com).

- What would the APIs of public methods look like? Can we update the existing STS API backward-compatibly, or is it easier/cleaner to create a new interface? Working with Tensor shapes can get complicated really quickly, so we generally have to be *really* careful to document a consistent semantics for the dimensions of every Tensor we accept or return. (for example, a potential backward-compatibility issue is that `canonicalize_observed_time_series_with_mask` currently pads observed time series with an implicit trailing dimension of size 1, if not already present). You could get a start on this by updating the args/returns section of the `empirical_statistics` docstring to describe the behavior after your proposed change.

See also my comment on the issue thread (https://github.com/tensorflow/probability/issues/1129#issuecomment-707979474).

As an example of the sort of thing we'd like to know: will we ever actually need the empirical covariances? For normalizing the series or initializing parameters, it might be sufficient, or even preferable, to just use the elementwise variances. Then again (depending on how we expect to specify models, and what parameters the components will need), it might not. Having a better sense of the overall plan would make it easier to decide what kind of changes we need to utility methods like this one.

europeanplaice:
There seem many things to consider before implementation.
As we know, the vector autoregression model provides a powerful tool to predict the future or analyze the causes and effects of related things.
I hope we have the overall plan for Multivariate STS someday.

Thank you for the comment with a lot of details!

PR #1357, opened 2021-06-11
DOC: log likelihood fails to sum over observed data points in one place

Leengit:


Leengit:
@googlebot I signed it!

PR #1369, opened 2021-06-24
Fix typo in LaTeX definition of Gaussian mixture

grisaitis:
I'm pretty sure this summation is supposed to happen over the mixtures _k_, not the covariates _i_:

image


SiegeLordEx:
Thanks!

PR #1380, opened 2021-07-13
Fix markdown

europeanplaice:
Hello.
I haven't tested, but I think his PR would solve the error of the document.


ColCarroll (approved):
Thanks for the catch!

Issue #1209, opened 2021-01-03
Error with seaborn in examples/bayesian_neural_network.py


rrklearn2020:
Request for help for bayesian_neural_network.py
I am trying to run the examples/bayesian_neural_network.py

### **_Trial 1: without seaborn_**

1. Program runs, but the graphs are not stored in the folder

### **_Trial 2: with seaborn_** (**_error_**, program stops and the graphs are not stored in the 'bayesian_neural_network' folder)
Version: 
seaborn-0.11.1
tensorflow==2.4.0
tensorflow-estimator==2.4.0
tensorflow-probability==0.12.1

Error: while " Running monte carlo inference" after the 1st 400th step (ie. viz_steps=400) and trying for "plot_heldout_prediction". Error details are shared below

### 
**_Error Details in "plot_heldout_prediction"_**

`saved /..../bayesian_neural_network/epoch0_step00399_weights.png
Traceback (most recent call last):
  File "/home/..../tf_latest/lib/python3.6/site-packages/seaborn/_core.py", line 1472, in categorical_order
    order = vector.cat.categories
AttributeError: 'tensorflow.python.framework.ops.EagerTensor' object has no attribute 'cat'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/..../tf_latest/lib/python3.6/site-packages/seaborn/_core.py", line 1476, in categorical_order
    order = vector.unique()
AttributeError: 'tensorflow.python.framework.ops.EagerTensor' object has no attribute 'unique'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "BNN_trial_1.py", line 366, in <module>
    app.run(main)
  File "/home/..../tf_latest/lib/python3.6/site-packages/absl/app.py", line 303, in run
    _run_main(main, args)
  File "/home/..../tf_latest/lib/python3.6/site-packages/absl/app.py", line 251, in _run_main
    sys.exit(main(argv))
  File "BNN_trial_1.py", line 362, in main
    .format(heldout_log_prob))
  File "BNN_trial_1.py", line 150, in plot_heldout_prediction
    sns.barplot(np.arange(10), prob_sample[i, :], alpha=0.1, ax=ax)
  File "/home/..../tf_latest/lib/python3.6/site-packages/seaborn/_decorators.py", line 46, in inner_f
    return f(**kwargs)
  File "/home/..../tf_latest/lib/python3.6/site-packages/seaborn/categorical.py", line 3182, in barplot
[...]

PR #1399, opened 2021-08-13
issue 1209 correction seaborn errors

gregorystrubel:
Proposition for correcting issue #1209 Error with seaborn in examples/bayesian_neural_network.py 

Convert tensorflow tensors in numpy array just before calling seaborn functions. It seems certain operations on tf tensors that numpy propose is not available yet.


gregorystrubel:
@googlebot I signed it!

gregorystrubel:
Need to change e-mail

Issue #685, opened 2019-12-13
NotImplementedError: mean is not implemented: Categorical


nbro:
I get the following error

> NotImplementedError: mean is not implemented: Categorical

If I execute the following piece of code

```
import tensorflow_probability as tfp

prob_dist = tfp.distributions.Categorical(probs=[1.0])
print(prob_dist.mean())
```

I get a similar error, if I attempt to call `stddev()`. This error 

The parameters mean and standard deviation are not usually defined for a categorical distribution and this error is due to the fact that [`tfp.distributions.Categorical`](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Categorical) inherits from [`tfp.distributions.Distribution`](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Distribution), which defines the `mean` and `stddev` methods. PyCharm even suggests the auto-completion of these two methods for the `tfp.distributions.Categorical`, but this is at least misleading. Maybe `tfp.distributions.Categorical` should not inherit those methods in the first place or `tfp.distributions.Distribution` should not have those methods in the first place, because not all distributions have a well-defined mean or standard deviation.

tomatitito:
I get the same error when calling `mode()` with an instance of the `Multinomial` or `DirichletMultinomial` distribution.

rajpratyush:
@srvasude is this issue still open for work?

PR #1411, opened 2021-08-23
Implemention of Mean Method in distributions.Categorical

PavanKishore21:
## Proposition for correcting issue #685 Error with Implementation of Mean Method in distributions.Categorical

> NotImplementedError: mean is not implemented: Categorical

```py

prob_dist = tfp.distributions.Categorical(probs=[1.0])
print(prob_dist.mean())

```

Defined a `_mean` method for implementing mean

### DIFF
--- a/tensorflow_probability/python/distributions/categorical.py
+++ b/tensorflow_probability/python/distributions/categorical.py
@@ -321,6 +321,12 @@ class Categorical(distribution.AutoCompositeTensorDistribution):
         tf.math.multiply_no_nan(masked_logits, tf.math.exp(logits)),
         axis=-1) / tf.math.exp(lse_logits)
 
+  def _mean(self):
+    x = self._probs if self._logits is None else self._logits
+    mean = tf.cast(tf.reduce_mean(x, axis=-1), self.dtype)
+    tensorshape_util.set_shape(mean, x.shape[:-1])
+    return mean      
+
   def _mode(self):
     x = self._probs if self._logits is None else self._logits
     mode = tf.cast(tf.argmax(x, axis=-1), self.dtype)
--- a/tensorflow_probability/python/distributions/categorical.py
+++ b/tensorflow_probability/python/distributions/categorical.py
@@ -322,10 +322,22 @@ class Categorical(distribution.AutoCompositeTensorDistribution):
         axis=-1) / tf.math.exp(lse_logits)
 
   def _mean(self):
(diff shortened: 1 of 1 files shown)


PavanKishore21:
@googlebot I signed it!

SiegeLordEx:
This implementation is not correct. The mean of a categorical is defined as `sum(i * prob(x=i) for i in range(num_categories)`. You should be able to modify the implementation of the `_entropy` to implement this.

This will also need tests to verify the implementation is correct (entropy tests, again, should provide a good set of test cases).

PavanKishore21:
Hello @SiegeLordEx, Thanks for your response.
I have updated the formula for Mean

`tf.reduce_sum(tf.math.multiply_no_nan(i , self._prob(x=i)) for i in range(self._num_categories), axis=-1`

Check this out

fotisdr:
The previous implementation was incorrect. I've written an example implementation of how the Categorical mean should be like, the code below works for me and produces the desired result. The mean should also account for unnormalised probabilities between the classes, since there are cases where the probabilities don't sum up to 1. 
```python
  def _mean(self):
    probs = self.probs_parameter()
    num_categories = self._num_categories(probs)
    # Initialise the mean with zeros
    categorical_mean = tf.zeros(tf.shape(probs[...,0]))
    # Compute the normalisation factors such that all probabilities sum up to 1
    normalisation = tf.reduce_sum(probs, axis=-1)
    # Sum up all the normalised probabilities
    # sum(i * prob(X=i) for i in range(num_categories) )
    for i in range(num_categories):
    	categorical_mean = categorical_mean + tf.cast(i,probs.dtype) * probs[...,i] / normalisation
    
    return categorical_mean
``` 
Shall I make a new commit for this?

fotisdr:
> I don't think there is currently an implementation. The following might be simpler: tf.reduce_sum(tf.range(self._num_categories(probs)) * probs, axis=-1) / tf.reduce_sum(probs, axis=-1) You could send a PR, sure.

Hi, 
Indeed this looks much simpler, thanks! We want to ensure that the multiplication is applied across the last axis of probs, i.e. each slice of probs across the last dimension (probs[...,i]) gets multiplied by each number i in tf.range(self._num_categories(probs)). The multiplication should be doing this by default, but I will test the code and submit a new PR.

SiegeLordEx:
This got implemented in b14ae1d79de4a52d834a3ba2dc88f7e5d849e6c7.

Issue #1421, opened 2021-09-04
non-static shapes prevent cholesky update to be used in JAX substrate mode.


AdrienCorenflos:
For the below, simply changing `tf.shape` to `ps.shape` does the trick.

https://github.com/tensorflow/probability/blob/84ea7be50ae1680a0e2cf851a95c55732c0ab5c7/tensorflow_probability/python/math/linalg.py#L153

https://github.com/tensorflow/probability/blob/84ea7be50ae1680a0e2cf851a95c55732c0ab5c7/tensorflow_probability/python/math/linalg.py#L199

Is there any reason why this would have undesired effects?

PR #1423, opened 2021-09-06
BUG FIX: prefer static shapes in cholesky update

AdrienCorenflos:
This is to fix https://github.com/tensorflow/probability/issues/1421

I am not sure the way I tested it is the preferred way, but this substrates testing is a bit fuzzy to me, so feel free to edit it (I left edits open for maintainers).

### DIFF
--- a/tensorflow_probability/python/math/linalg_test.py
+++ b/tensorflow_probability/python/math/linalg_test.py
@@ -142,8 +142,16 @@ class PushApartTest(test_util.TestCase):
 
 
 class _CholeskyUpdate(test_util.TestCase):
+  def testCholeskyUpdateXLA(self):
+    self.skip_if_no_xla()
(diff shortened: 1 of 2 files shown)


tensorflow_probability/python/math/linalg_test.py
```diff
@@ -142,8 +142,17 @@ def testPreservesSortOrder(self, data):
...
 
 class _CholeskyUpdate(test_util.TestCase):
+  def testCholeskyUpdateXLA(self):
+    self.skip_if_no_xla()
+    cholesky_update_fun = tf.function(tfp.math.cholesky_update,
+                                      jit_compile=True)
+    self._testCholeskyUpdate(cholesky_update_fun)
 
   def testCholeskyUpdate(self):
+    self.skip_if_no_xla()
```
brianwa84, line 152: let's omit this line

brianwa84:
Actually how about I do that fix while I pull it in.

PR #1463, opened 2021-11-12
:memo: Fix href-markdown problems

MiWeiss:
The current documentation was wrongly parsed as markdown links, making the changes within the `()` invisible in the deployed docs, and instead showing broken links.

Changes: I removed unneeded brackets which provides a (subjectively) smoother read and also fixes the above mentioned markdown problems. I also renamed the reference to John D. Cook's Blog.


SiegeLordEx (approved):
Thanks!

PR #1476, opened 2021-12-10
Autoregressive moving average cla

jakee417:
@davmre clean PR with correct cla! `log_prob` issue is resolved, just needed to kick off the recursion in a similar manner as `autoregressive_test.py`. Lmk if there is anything else we need before we merge, enjoy the long weekend!


davmre:
This just went in. Thanks again for putting this together, it's a really nice contribution!

Issue #1081, opened 2020-09-16
[Feature Request] tensorflow_probability.distributions.GeneralizedGamma


kmdalton:
In X-ray crystallography, the most important prior distributions include two special cases of the [generalized gamma distribtion](https://en.wikipedia.org/wiki/Generalized_gamma_distribution). I am very keen to try this parameterization of the variational distritribution in my [research project](https://github.com/Hekstra-Lab/careless). How hard would it be for the TFP devs to implement this distribution? What is the likelihood of it being available in the near future?

Unrelated: TFP is a great package. Nobody else has so many useful reparameterized distributions. Thanks!

brianwa84:
It looks like there is a rejection sampling approach from 1979, but I haven't read it closely to make sure it's the same/compatible pdf.https://www.researchgate.net/profile/Pandu_Tadikamalla/publication/227309949_Random_sampling_from_the_generalized_gamma_distribution/links/55de04bc08aeaa26af0f20a2.pdf

This could be a nice "first contribution" sized project, good way to get some exposure to TFP's rejection sampling lib.

SanjayMarreddi:
@kmdalton @brianwa84 

I would like to work on it. Could you please assign this to me !

Thanks and regards,
Sanjay Marreddi.

brianwa84:
Sure, done. In terms of sequencing, I'd suggest to first get the rejection sampler written & a PR committed, then build a Distribution class around that.

You could look at some example rejection samplers in e.g. https://cs.opensource.google/tensorflow/probability/+/master:tensorflow_probability/python/distributions/poisson.py;l=351
https://cs.opensource.google/tensorflow/probability/+/master:tensorflow_probability/python/distributions/gamma.py;l=654

The corresponding test files will be likely be using the statistical_testing library to verify the empirical cdf of a number of samples against a true cdf.

kmdalton:
I found some info on the Amaroso distribution, which adds a fourth parameter to the generalized gamma parameterization linked in my original post. 
<img src="https://user-images.githubusercontent.com/2790777/93474777-cd9bb180-f8c5-11ea-964b-178e14626cb1.png" width="75%">

According to [this](https://arxiv.org/pdf/1005.3274.pdf), Amaroso random samples can be generated from gamma distributed random samples by a simple transformation. 

<img src="https://user-images.githubusercontent.com/2790777/93474619-9927f580-f8c5-11ea-9524-1e93955aaa41.png" width=75%>
[...]

PR #1489, opened 2022-01-07
Generalized gamma distribution

joshchang:
In response to #1081 , here is a GeneralizedGamma distribution based on the parameterization here: https://en.wikipedia.org/wiki/Generalized_gamma_distribution


joshchang:
forgot to update branch before pr

PR #1501, opened 2022-01-18
modified SplineParams example for improved clarity

gisilvs:
When using SplineParams as in the example, it was not clear to me how to set the interval width and height B as defined in the Neural Splines Flow paper. I have modified the example by adding parameters that the user can specify: the interval width, the smallest knot position, the minimum width/height of the bins (assuming that they will be equal) and the minimum slope. The default values for `min_bin_width` and `min_slope` are set as they are in the official pytorch implementation https://github.com/bayesiains/nflows/blob/master/nflows/transforms/splines/rational_quadratic.py

I believe that some comments might be needed to clarify what `nunits` is and why it is given as argument to `__call__`, as it was quite confusing to me and it's specific to using the splines with RealNVP. Otherwise, an additional example using the splines with MaskedAutoregressiveFlow might also be helpful.

### DIFF
--- a/tensorflow_probability/python/bijectors/rational_quadratic_spline.py
+++ b/tensorflow_probability/python/bijectors/rational_quadratic_spline.py
@@ -73,8 +73,13 @@ class RationalQuadraticSpline(bijector.AutoCompositeTensorBijector):
 
   class SplineParams(tf.Module):
 
-    def __init__(self, nbins=32):
+    def __init__(self, nbins=32, interval_width=2, range_min=-1,
+                 min_bin_width=1e-3, min_slope=1e-3):
       self._nbins = nbins
+      self._interval_width = interval_width
+      self._range_min = range_min
+      self._min_bin_width = min_bin_width
+      self._min_slope = min_slope
       self._built = False
       self._bin_widths = None
       self._bin_heights = None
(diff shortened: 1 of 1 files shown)


tensorflow_probability/python/bijectors/rational_quadratic_spline.py
```diff
@@ -73,8 +73,13 @@ class RationalQuadraticSpline(bijector.AutoCompositeTensorBijector):
 
   class SplineParams(tf.Module):
 
-    def __init__(self, nbins=32):
+    def __init__(self, nbins=32, interval_width=2, range_min=-1,
+                 min_bin_width=1e-3, min_slope=1e-3):
       self._nbins = nbins
+      self._interval_width = interval_width
```
emilyfertig, line 79: Could you add some inline comments to briefly define these, like
```
      self._interval_width = interval_width   # Sum of bin widths.
      self._min_bin_width = min_bin_width  # Bin width lower bound.
```
etc.?

tensorflow_probability/python/bijectors/rational_quadratic_spline.py
```diff
@@ -85,26 +90,29 @@ def __call__(self, x, nunits):
         def _bin_positions(x):
           out_shape = tf.concat((tf.shape(x)[:-1], (nunits, self._nbins)), 0)
           x = tf.reshape(x, out_shape)
-          return tf.math.softmax(x, axis=-1) * (2 - self._nbins * 1e-2) + 1e-2
+          return tf.math.softmax(x, axis=-1) * (
+                self._interval_width - self._nbins * self._min_bin_width) + \
+                 self._min_bin_width
```
emilyfertig, line 95: Nit: Google's style guide prefers implicit line joining inside parentheses instead of backslash line continuation: https://google.github.io/styleguide/pyguide.html#32-line-length (i.e. the last line of this block could be `) + self._min_bin_width`

emilyfertig (changes requested):
Thanks, this is a great clarification!

emilyfertig (approved):

PR #1506, opened 2022-01-29
Update differential_evolution.py

Cheril311:
Fixes https://github.com/tensorflow/tensorflow/issues/54187

### DIFF
--- a/tensorflow_probability/python/optimizer/differential_evolution.py
+++ b/tensorflow_probability/python/optimizer/differential_evolution.py
@@ -122,7 +122,7 @@ def one_step(
     population,
     population_values=None,
     differential_weight=0.5,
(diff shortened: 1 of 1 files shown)


ColCarroll:
Sure! You would want to put a method in [this file](https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/optimizer/differential_evolution_test.py) that will fail without your change, and not fail with your change. Ideally, this would be a small, fast bit of code to run.

Cheril311:
@ColCarroll I am closing this PR, as the issue could not be reproduced in tf-nightly today, so I think something has been taken care of! Thanks for you time and guidance

ColCarroll:
Thank you for following up!

PR #1507, opened 2022-02-04
Fix Multivariate Student's t-distribution docstring

jvishnuvardhan:
Python example within the docstring was not closed and hence the TF page was not rendered correctly


axch (approved):
Thanks for the fix!

The lints look like pre-existing problems, so I will pull this in.

PR #1514, opened 2022-02-14
Avoid custom_gradient warnings

tk2lab:
python.interlnal.custom_gradient.custom_gradient generate warnings: "@custom_gradient grad_fn has 'variables' in signature, but no ResourceVariables were used on the forward pass.".

TF will specify not only arg 'variabels' but also **kwargs.
In 'f_wrapped', vjp_bwd can use kwargs even if vjp_bwd_wrapped does not have kwargs.


tk2lab:
signed

PR #1518, opened 2022-02-18
Correctly rendering iterated_filter by correcting backticks

jvishnuvardhan:
Backtics are not formatted well which is resulting in rendering issues. This PR updates those backticks.


axch:
Looks like the lint is failing on pre-existing problems.  These changes look good -- thanks! -- I will pull them in.

axch (approved):

PR #1520, opened 2022-02-21
Fixed typo in kernel_bias.py

mohantym:
Updated "Recomendations" to "Recommendations"


axch (approved):
Thanks for the catch!

Issue #399, opened 2019-05-08
Skew-Normal distribution


brianwa84:
PyMC folks would like a skew-normal (ideally multivariate) distribution. A start was made here: https://github.com/pymc-devs/pymc4/pull/85

brianwa84:
Also see https://github.com/pymc-devs/pymc4/issues/44

mihailyanchev:
Hi, would you know if there is a univariate version of the skewed normal distribution already implemented anywhere?

leandrolcampos:
Hi @brianwa84,

Can I try to solve this issue implementing the univariate and multivariate skew-normal and skew-t distributions? I'd like to use the method proposed in [1] to construct a skew density from a symmetric one. For the univariate cases, the closed-form expressions for mean, variance, PDF, CDF, and inverse CDF are shown in [2] (see the distributions SN2 and ST3). A multivariate extension is outlined in [3], which also describes a general method for sampling from these distributions.

[1] Fernández, C., & Steel, M. F. (1998). On Bayesian modeling of fat tails and skewness. Journal of the American Statistical Association, 93(441), 359-371.

[2] Rigby, R. A., Stasinopoulos, M. D., Heller, G. Z., & De Bastiani, F. (2019). Distributions for modeling location, scale, and shape: Using GAMLSS in R. Chapman and Hall/CRC.

[3] Bauwens, L., & Laurent, S. (2005). A new class of multivariate skew densities, with application to generalized autoregressive conditional heteroscedasticity models. Journal of Business & Economic Statistics, 23(3), 346-354.

brianwa84:
I think the closest thing we have right now is this:
https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/GeneralizedNormal

As a general rule we'd welcome contributions of new distributions w/
tractable sampling & pdf. All the other methods are icing on the cake. It's
worth noting as you start that we have a bit higher of a bar than
many opensource projects in terms of testing, code style, etc. (probably
comparable to TensorFlow or Numpy)
https://github.com/tensorflow/probability/blob/main/CONTRIBUTING.md and
https://github.com/tensorflow/probability/blob/main/discussion/tfp_distributions_contract.md
would be good to browse.

Brian Patton | Software Engineer | ***@***.***

leandrolcampos:
Hi @brianwa84,

I created a pull request, [#1524](https://github.com/tensorflow/probability/pull/1524), that provides an implementation for the Skew Normal distribution.

[...]

PR #1524, opened 2022-02-23
Add Skew Normal distribution

leandrolcampos:
This PR provides an implementation for the Skew Normal distribution proposed by Fernández and Steel (1998)[1]. It addresses the issue [#399](https://github.com/tensorflow/probability/issues/399).

The Skew Normal generalizes the Normal distribution with an additional shape parameter. It is parameterized by location `loc`, scale `scale`, and shape `skewness`. If the skewness is above one, the distribution becomes positively skewed (or right-skewed). If the skewness is greater than zero and less than one, the distribution becomes negatively skewed (or left-skewed). A skewness equal to one results in a Normal distribution.

I tried to follow all the style decisions provided in the [style guide](https://github.com/tensorflow/probability/blob/main/STYLE_GUIDE.md). But in some cases I couldn't follow the TensorFlow Probability-specific conventions (8), (16) or (24) because one or more tests failed.

This PR, including the test suite, was inspired by the available implementation of the Generalized Normal distribution.

Any feedback is appreciated.

[1] Carmen Fernández and Mark F. J. Steel. On Bayesian modeling of fat tails and skewness. _Journal of the American Statistical Association_, 93(441), 359-371, 1998.

### DIFF
--- a/tensorflow_probability/python/distributions/skew_normal.py
+++ b/tensorflow_probability/python/distributions/skew_normal.py
@@ -53,7 +53,7 @@ def standardize(value, loc, scale, skewness):
     A tensor with shape broadcast according to the arguments.
   """
   return (value - loc) / tf.math.abs(scale) * tf.math.abs(
-      tf.where(value < loc, x=skewness, y=tf.math.reciprocal(skewness)))
+      tf.where(value < loc, skewness, tf.math.reciprocal(skewness)))
 
 
 def cdf(value, loc, scale, skewness):
@@ -79,8 +79,8 @@ def cdf(value, loc, scale, skewness):
   squared_skewness = tf.math.square(skewness)
   return tf.math.reciprocal(one + squared_skewness) * tf.where(
       z < 0.,
-      x=two * normal_cdf,
-      y=one - squared_skewness + two * squared_skewness * normal_cdf)
+      two * normal_cdf,
+      one - squared_skewness + two * squared_skewness * normal_cdf)
 
 
 def quantile(value, loc, scale, skewness):
@@ -107,12 +107,12 @@ def quantile(value, loc, scale, skewness):
   # Here we use the following fact:
   # X ~ Normal(loc=0, scale=1) => 2 * X**2 ~ Gamma(alpha=0.5, beta=1)
   probs = (one - value * (one + squared_skewness)) * tf.where(
-      cond, x=one, y=-tf.math.reciprocal(squared_skewness))
+      cond, one, -tf.math.reciprocal(squared_skewness))
   gamma_quantile = tfp_math.igammainv(half, p=probs)
 
   abs_skewness = tf.math.abs(skewness)
   adj_scale = tf.math.abs(scale) * tf.where(
-      cond, x=-tf.math.reciprocal(abs_skewness), y=abs_skewness)
+      cond, -tf.math.reciprocal(abs_skewness), abs_skewness)
 
   return loc + adj_scale * tf.cast(
       tf.math.sqrt(two * gamma_quantile), dtype=loc.dtype)
@@ -297,7 +297,6 @@ class SkewNormal(distribution.AutoCompositeTensorDistribution):
 
   @classmethod
   def _parameter_properties(cls, dtype, num_classes=None):
-    # pylint: disable=g-long-lambda
     return dict(
         loc=parameter_properties.ParameterProperties(),
         scale=parameter_properties.ParameterProperties(
@@ -306,7 +305,6 @@ class SkewNormal(distribution.AutoCompositeTensorDistribution):
         skewness=parameter_properties.ParameterProperties(
             default_constraining_bijector_fn=(
                 lambda: softplus_bijector.Softplus(low=dtype_util.eps(dtype)))))
-    # pylint: enable=g-long-lambda
 
   @property
   def loc(self):
@@ -346,8 +344,8 @@ class SkewNormal(distribution.AutoCompositeTensorDistribution):
 
     sample = tf.abs(normal_sample) * tf.where(
         uniform_sample < tf.math.reciprocal(1. + skewness**2),
-        x=-tf.math.reciprocal(skewness),
-        y=skewness)
+        -tf.math.reciprocal(skewness),
+        skewness)
 
     return loc + scale * sample
 
--- a/tensorflow_probability/python/distributions/skew_normal.py
+++ b/tensorflow_probability/python/distributions/skew_normal.py
@@ -1,0 +1,475 @@
+# Copyright 2022 The TensorFlow Probability Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ============================================================================
+"""The Skew Normal distribution class."""
+
+# Dependency imports
+import numpy as np
+
+import tensorflow.compat.v2 as tf
+
+from tensorflow_probability.python import math as tfp_math
+from tensorflow_probability.python.bijectors import identity as identity_bijector
+from tensorflow_probability.python.bijectors import softplus as softplus_bijector
+from tensorflow_probability.python.distributions import distribution
+from tensorflow_probability.python.internal import assert_util
+from tensorflow_probability.python.internal import dtype_util
+from tensorflow_probability.python.internal import parameter_properties
+from tensorflow_probability.python.internal import prefer_static as ps
+from tensorflow_probability.python.internal import reparameterization
+from tensorflow_probability.python.internal import samplers
+from tensorflow_probability.python.internal import special_math
+from tensorflow_probability.python.internal import tensor_util
+from tensorflow_probability.python.math.numeric import log1psquare
+
+__all__ = [
+    'SkewNormal',
+]
+
+
+def standardize(value, loc, scale, skewness):
+  """Apply mean-variance-skewness standardization to input `value`.
+
+  Note that scale and skewness can be negative.
+
+  Args:
+    value: Floating-point tensor; the value(s) to be standardized.
+    loc: Floating-point tensor; the location(s) of the distribution(s).
+    scale: Floating-point tensor; the scale(s) of the distribution(s).
+    skewness: Floating-point tensor; the skewness(es) of the distribution(s).
+
+  Returns:
+    A tensor with shape broadcast according to the arguments.
+  """
+  return (value - loc) / tf.math.abs(scale) * tf.math.abs(
+      tf.where(value < loc, x=skewness, y=tf.math.reciprocal(skewness)))
+
+
+def cdf(value, loc, scale, skewness):
+  """Compute cumulative distribution function of Skew Normal distribution.
+
+  Note that scale and skewness can be negative.
+
+  Args:
+    value: Floating-point tensor; where to compute the cdf.
+    loc: Floating-point tensor; the location(s) of the distribution(s).
+    scale: Floating-point tensor; the scale(s) of the distribution(s).
+    skewness: Floating-point tensor; the skewness(es) of the distribution(s).
+
+  Returns:
+    A tensor with shape broadcast according to the arguments.
+  """
+  one = tf.constant(1., dtype=loc.dtype)
+  two = tf.constant(2., dtype=loc.dtype)
+
+  z = standardize(value, loc=loc, scale=scale, skewness=skewness)
+  normal_cdf = tf.cast(special_math.ndtr(z), dtype=loc.dtype)
+
+  squared_skewness = tf.math.square(skewness)
+  return tf.math.reciprocal(one + squared_skewness) * tf.where(
+      z < 0.,
+      x=two * normal_cdf,
+      y=one - squared_skewness + two * squared_skewness * normal_cdf)
+
+
+def quantile(value, loc, scale, skewness):
+  """Compute quantile function (inverse cdf) of Skew Normal distribution.
+
+  Note that scale and skewness can be negative.
+
+  Args:
+    value: Floating-point tensor; where to compute the quantile function.
+    loc: Floating-point tensor; the location(s) of the distribution(s).
+    scale: Floating-point tensor; the scale(s) of the distribution(s).
+    skewness: Floating-point tensor; the skewness(es) of the distribution(s).
+
+  Returns:
+    A tensor with shape broadcast according to the arguments.
+  """
+  half = tf.constant(0.5, dtype=loc.dtype)
+  one = tf.constant(1., dtype=loc.dtype)
+  two = tf.constant(2., dtype=loc.dtype)
+
+  squared_skewness = tf.math.square(skewness)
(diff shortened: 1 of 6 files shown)


tensorflow_probability/python/distributions/skew_normal.py
```diff
@@ -0,0 +1,473 @@
...
+  abs_skewness = tf.math.abs(skewness)
+  adj_scale = tf.math.abs(scale) * tf.where(
+      cond, x=-tf.math.reciprocal(abs_skewness), y=abs_skewness)
+
+  return loc + adj_scale * tf.cast(
+      tf.math.sqrt(two * gamma_quantile), dtype=loc.dtype)
+
+
+class SkewNormal(distribution.AutoCompositeTensorDistribution):
+  """The Skew Normal distribution.
```
SiegeLordEx, line 122: I'm a little concerned that we're calling this 'The' skew normal distribution, when there's at least two common variants of this. This one appears to be by the 'type 2' from gamlss, while when people search for skew normal, e.g., on wikipedia they get the 'type 1' (https://en.wikipedia.org/wiki/Skew_normal_distribution). Wikipedia also has https://en.wikipedia.org/wiki/Split_normal_distribution which looks similar to this formulation, but not quite the same. I note you cite `Fernández-Steel distribution` below (I couldn't get a copy of that manuscript, so I don't know the details).

Do you have any insight about what best to call it?

tensorflow_probability/python/distributions/skew_normal.py
```diff
@@ -0,0 +1,473 @@
...
+  # X ~ Normal(loc=0, scale=1) => 2 * X**2 ~ Gamma(alpha=0.5, beta=1)
+  probs = (one - value * (one + squared_skewness)) * tf.where(
+      cond, x=one, y=-tf.math.reciprocal(squared_skewness))
+  gamma_quantile = tfp_math.igammainv(half, p=probs)
+
+  abs_skewness = tf.math.abs(skewness)
+  adj_scale = tf.math.abs(scale) * tf.where(
+      cond, x=-tf.math.reciprocal(abs_skewness), y=abs_skewness)
+
+  return loc + adj_scale * tf.cast(
```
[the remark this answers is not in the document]
leandrolcampos: When I try to remove this cast, the test fails with the following message:

> (...)
INFO: Analyzed 3 targets (0 packages loaded, 0 targets configured).
INFO: Found 3 test targets...
FAIL: //tensorflow_probability/python/distributions:skew_normal_test.numpy (see /home/leandro/.cache/bazel/_bazel_leandro/b8274678e6a17bc831bbd9e0d37a2c28/execroot/tensorflow_probability/bazel-out/k8-opt/testlogs/tensorflow_probability/python/distributions/skew_normal_test.numpy/test.log)
(...)
FAIL: testSampleLikeArgsGetDistDType (__main__.SkewNormalTestDynamicShapeFloat32)
SkewNormalTestDynamicShapeFloat32.testSampleLikeArgsGetDistDType
  File "/home/leandro/.cache/bazel/_bazel_leandro/b8274678e6a17bc831bbd9e0d37a2c28/sandbox/linux-sandbox/874/execroot/tensorflow_probability/bazel-out/k8-opt/bin/tensorflow_probability/python/distributions/skew_normal_test.numpy.runfiles/tensorflow_probability/tensorflow_probability/python/distributions/_numpy/_generated_skew_normal_test.py", line 110, in testSampleLikeArgsGetDistDType
    self.assertEqual(self.dtype, getattr(dist, method)(1).dtype)
  File "/home/leandro/.cache/bazel/_bazel_leandro/b8274678e6a17bc831bbd9e0d37a2c28/sandbox/linux-sandbox/874/execroot/tensorflow_probability/bazel-out/k8-opt/bin/tensorflow_probability/python/distributions/skew_normal_test.numpy.runfiles/tensorflow_probability/tensorflow_probability/python/internal/backend/numpy/test_lib.py", line 85, in assertEqual
    return super(TestCase, self).assertEqual(first, second, msg)
AssertionError: <class 'numpy.float32'> != dtype('float64')
FAIL: testSampleLikeArgsGetDistDType (__main__.SkewNormalTestStaticShapeFloat32)
SkewNormalTestStaticShapeFloat32.testSampleLikeArgsGetDistDType
  File "/home/leandro/.cache/bazel/_bazel_leandro/b8274678e6a17bc831bbd9e0d37a2c28/sandbox/linux-sandbox/874/execroot/tensorflow_probability/bazel-out/k8-opt/bin/tensorflow_probability/python/distributions/skew_normal_test.numpy.runfiles/tensorflow_probability/tensorflow_probability/python/distributions/_numpy/_generated_skew_normal_test.py", line 110, in testSampleLikeArgsGetDistDType
    self.assertEqual(self.dtype, getattr(dist, method)(1).dtype)
  File "/home/leandro/.cache/bazel/_bazel_leandro/b8274678e6a17bc831bbd9e0d37a2c28/sandbox/linux-sandbox/874/execroot/tensorflow_probability/bazel-out/k8-opt/bin/tensorflow_probability/python/distributions/skew_normal_test.numpy.runfiles/tensorflow_probability/tensorflow_probability/python/internal/backend/numpy/test_lib.py", line 85, in assertEqual
    return super(TestCase, self).assertEqual(first, second, msg)
AssertionError: <class 'numpy.float32'> != dtype('float64')

tensorflow_probability/python/distributions/skew_normal.py
```diff
@@ -0,0 +1,473 @@
...
+
+    half = tf.constant(0.5, dtype=self.dtype)
+    two = tf.constant(2., dtype=self.dtype)
+    pi = tf.constant(np.pi, dtype=self.dtype)
+
+    z = standardize(value, loc=loc, scale=scale, skewness=skewness)
+
+    log_unnormalized = -half * tf.math.square(z)
+    log_normalization = (
+        tf.cast(log1psquare(skewness), dtype=self.dtype) -
```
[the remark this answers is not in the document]
leandrolcampos: When I try to remove this tf.cast, the test also fails with the same message:

> (...)
INFO: Analyzed 3 targets (0 packages loaded, 0 targets configured).
INFO: Found 3 test targets...
FAIL: //tensorflow_probability/python/distributions:skew_normal_test.numpy (see /home/leandro/.cache/bazel/_bazel_leandro/b8274678e6a17bc831bbd9e0d37a2c28/execroot/tensorflow_probability/bazel-out/k8-opt/testlogs/tensorflow_probability/python/distributions/skew_normal_test.numpy/test.log)
(...)
FAIL: testSampleLikeArgsGetDistDType (__main__.SkewNormalTestDynamicShapeFloat32)
SkewNormalTestDynamicShapeFloat32.testSampleLikeArgsGetDistDType
  File "/home/leandro/.cache/bazel/_bazel_leandro/b8274678e6a17bc831bbd9e0d37a2c28/sandbox/linux-sandbox/883/execroot/tensorflow_probability/bazel-out/k8-opt/bin/tensorflow_probability/python/distributions/skew_normal_test.numpy.runfiles/tensorflow_probability/tensorflow_probability/python/distributions/_numpy/_generated_skew_normal_test.py", line 110, in testSampleLikeArgsGetDistDType
    self.assertEqual(self.dtype, getattr(dist, method)(1).dtype)
  File "/home/leandro/.cache/bazel/_bazel_leandro/b8274678e6a17bc831bbd9e0d37a2c28/sandbox/linux-sandbox/883/execroot/tensorflow_probability/bazel-out/k8-opt/bin/tensorflow_probability/python/distributions/skew_normal_test.numpy.runfiles/tensorflow_probability/tensorflow_probability/python/internal/backend/numpy/test_lib.py", line 85, in assertEqual
    return super(TestCase, self).assertEqual(first, second, msg)
AssertionError: <class 'numpy.float32'> != dtype('float64')
FAIL: testSampleLikeArgsGetDistDType (__main__.SkewNormalTestStaticShapeFloat32)
SkewNormalTestStaticShapeFloat32.testSampleLikeArgsGetDistDType
  File "/home/leandro/.cache/bazel/_bazel_leandro/b8274678e6a17bc831bbd9e0d37a2c28/sandbox/linux-sandbox/883/execroot/tensorflow_probability/bazel-out/k8-opt/bin/tensorflow_probability/python/distributions/skew_normal_test.numpy.runfiles/tensorflow_probability/tensorflow_probability/python/distributions/_numpy/_generated_skew_normal_test.py", line 110, in testSampleLikeArgsGetDistDType
    self.assertEqual(self.dtype, getattr(dist, method)(1).dtype)
  File "/home/leandro/.cache/bazel/_bazel_leandro/b8274678e6a17bc831bbd9e0d37a2c28/sandbox/linux-sandbox/883/execroot/tensorflow_probability/bazel-out/k8-opt/bin/tensorflow_probability/python/distributions/skew_normal_test.numpy.runfiles/tensorflow_probability/tensorflow_probability/python/internal/backend/numpy/test_lib.py", line 85, in assertEqual
    return super(TestCase, self).assertEqual(first, second, msg)
AssertionError: <class 'numpy.float32'> != dtype('float64')

tensorflow_probability/python/distributions/skew_normal.py
leandrolcampos: Thanks for your feedback!

We can call this “The Two-Piece Normal distribution” [(Robert A. Rigby et al., 2019, Section 18.3.5, p380)][3].

In fact, according to [Rubio and Steel (2020, Section “Who discovered them?”)][4], the two-piece normal distribution “was introduced in a general framework by Fernández and Steel under an alternative parameterisation”. They refer to the formulation proposed by [Fernández and Steel (1998)][2]. So, to be more specific, we can call this distribution “The Two-Piece Normal distribution under the parameterization proposed by [Fernández and Steel (1998)][2]”. And we can name the class `TwoPieceNormal`. What do you think? If you agree, I can make the necessary changes.

We can see the mathematical equivalence between the parameterization proposed by [Fernández and Steel (1998)][2] and the main parameterization discussed in https://en.wikipedia.org/wiki/Split_normal_distribution. Consider that the two-piece normal distribution is built joining at the mode the corresponding halves of two normal distributions with the same mode but different standard deviation. Under the former parameterization, the left (or right) hand side standard deviation is given by `sigma / skewness` (or `sigma * skewness`), where `sigma` and `skewness` are, respectively, the scale and shape parameters of the two-piece normal distribution. Under the latter parameterization, the left (or right) hand side standard deviation is given directly by `sigma_1` (or `sigma_2`). If we use the relationships `sigma_1 = sigma / skewness` and `sigma_2 = sigma * skewness`, then we can go from one parameterization to the other.

The other option consists of maintaining the class name `SkewNormal` but presenting the distribution as "The Fernández-Steel Skew Normal distribution", following [Castillo et al. (2011)][1]. In fact, there is no official version of the skew normal distribution. And the class name `SkewNormal` is better at informing the purpose and main features of this distribution than the class name `TwoPieceNormal` (although the latter is more specific).

[1] Nabor O. Castillo et al. On the Fernández-Steel distribution: Inference and application. _Computational Statistics & Data Analysis_, 55(11), 2951-2961, 2011.

[2] Carmen Fernández and Mark F. J. Steel. On Bayesian modeling of fat tails and skewness. _Journal of the American Statistical Association_, 93(441), 359-371, 1998.

[3] Robert A. Rigby et al. _Distributions for modeling location, scale, and shape: Using GAMLSS in R_. Chapman and Hall/CRC, 2019.

[4] Francisco J. Rubio and Mark F. J. Steel. The family of two‐piece distributions. _Significance_, 17(1), 12-13, 2020.
https://rss.onlinelibrary.wiley.com/doi/full/10.1111/j.1740-9713.2020.01352.x

SiegeLordEx: Thanks for looking into it. Let's do this:

1. Rename it to TwoPieceNormal.
2. Add the sentence you proposed (The Two-Piece Normal distribution under the parameterization proposed by [Fernández and Steel (1998)][2])
3. Somewhere in the docstring we should have a section saying that this distribution has also been called the SkewNormal and SplitNormal and cite wikipedia and GAMLSS.

The docstring change should mitigate the somewhat indirect name, so if people seach for skewnormal they might still be able to find this distribution.

leandrolcampos: I made the necessary changes as we agreed.

SiegeLordEx (approved):

SiegeLordEx:
Thanks again. I'll send this along to our internal review system and once that's done, we'll merge this.

SiegeLordEx:
In internal review we observed that the samples of this distributions are not actually reparameterizable, due to the discrete bernoulli used to pick which piece to sample from. We removed/changed the code that implied otherwise. If you're interested in fixing that, it's still possible to implement reparameterizable gradients for this via implicit methods. A good example of that is the https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/distributions/truncated_normal.py code. The overall approach is described here: https://arxiv.org/abs/1805.08498

mihailyanchev:
@leandrolcampos Thank you for your contribution. I had a quick question. I noticed that the textbook rule on the relative positions of mean and median does not apply here - for right/positive skewness the mean is greater than the median and vice versa. Is this expected given this specific implementation? I know this textbook rule is not reliable (https://www.tandfonline.com/doi/pdf/10.1080/10691898.2005.11910556), but just wondered if you have looked into this by any chance?

`test_dist = SkewNormal(loc=0, scale=1, skewness=1.25)
print(np.array(test_dist.quantile(0.5)))
0.28443122
`

leandrolcampos:
Hi @mihailyanchev! I hope this message finds you well. And thanks for your feedback!

In the case of Two-Piece Normal distribution, the location parameter is different from the mean. In fact, it defines the mode. Please consider the following code block:

```python
tfd = tfp.distributions

# If skewness is greater than zero and less than one, dist becomes left-skewed 
# (or negatively skewed).
dist = tfd.TwoPieceNormal(loc=0., scale=1., skewness=0.75)

mode = dist.mode() # == loc
median = dist.quantile(0.5)
mean = dist.mean()

print(mean < median < mode) # True
print(mean, median, mode) # -0.46543267, -0.3702539, 0.0

# If skewness is above one, the distribution becomes right-skewed
# (or positively skewed).
dist = tfd.TwoPieceNormal(loc=0., scale=1., skewness=1.33)

mode = dist.mode() # == loc
median = dist.quantile(0.5)
mean = dist.mean()

print(mode < median < mean) # True
print(mode, median, mean) # 0.0, 0.3668836, 0.46127328
```

Another reference: https://en.wikipedia.org/wiki/Skewness

> @leandrolcampos Thank you for your contribution. I had a quick question. I noticed that the textbook rule on the relative positions of mean and median does not apply here - for right/positive skewness the mean is greater than the median and vice versa. Is this expected given this specific implementation? I know this textbook rule is not reliable (https://www.tandfonline.com/doi/pdf/10.1080/10691898.2005.11910556), but just wondered if you have looked into this by any chance?
> 
> `test_dist = SkewNormal(loc=0, scale=1, skewness=1.25)` `print(np.array(test_dist.quantile(0.5)))` `0.28443122`

mihailyanchev:
@leandrolcampos thanks for the quick reply. This is very helpful. I did not realize loc defines the mode and this makes it clear.

leandrolcampos:
> In internal review we observed that the samples of this distributions are not actually reparameterizable, due to the discrete bernoulli used to pick which piece to sample from. We removed/changed the code that implied otherwise. If you're interested in fixing that, it's still possible to implement reparameterizable gradients for this via implicit methods. A good example of that is the https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/distributions/truncated_normal.py code. The overall approach is described here: https://arxiv.org/abs/1805.08498

Hi @SiegeLordEx, I tried to make the Two-Piece Normal distribution differentiable w.r.t. all parameters in this PR: https://github.com/tensorflow/probability/pull/1541

PR #1541, opened 2022-04-02
Add implicit reparameterization gradient to tfd.TwoPieceNormal

leandrolcampos:
Add implicit reparameterization gradient w.r.t. `skewness`, making the distribution `tfd.TwoPieceNormal` fully reparameterized. The applied approach to compute the implicit gradient was proposed by Figurnov, Mohamed, and Mnih (2018)[1].

In addition to implementing the implicit gradient described above, this PR makes the following changes:
* In the file `tensorflow_probability/python/distributions/BUILD`:
    * Add a new dependency
* In the file `tensorflow_probability/python/distributions/two_piece_normal.py`:
    * Reorganize the file, putting all functions below the class definition
    * Run all Tensor args through `tf.convert_to_tensor` in the functions `standardize`, `cdf`, and `quantile`, making them consistent with `random_two_piece_normal`
* In the file `tensorflow_probability/python/distributions/two_piece_normal_test.py`:
    * Improve `testSample`, testing different sample shapes
    * Add `testDifferentiableSampleNumerically` and `testDifferentiableSampleAnalytically`, testing the gradients of the samples w.r.t. `loc`, `scale`, and `skewness`

#### Reference

[1] Michael Figurnov, Shakir Mohamed, and Andriy Mnih. Implicit Reparameterization Gradients. In _Advances in Neural Information Processing Systems_, 31, 2018.
https://arxiv.org/abs/1805.08498

### DIFF
--- a/tensorflow_probability/python/distributions/two_piece_normal.py
+++ b/tensorflow_probability/python/distributions/two_piece_normal.py
@@ -221,9 +221,7 @@ class TwoPieceNormal(distribution.AutoCompositeTensorDistribution):
           skewness, dtype=dtype, name='skewness')
       super().__init__(
           dtype=dtype,
-          # skewness contributes to a discrete choice. The other two variables
-          # are fine.
-          reparameterization_type=reparameterization.NOT_REPARAMETERIZED,
+          reparameterization_type=reparameterization.FULLY_REPARAMETERIZED,
           validate_args=validate_args,
           allow_nan_stats=allow_nan_stats,
           parameters=parameters,
--- a/tensorflow_probability/python/distributions/two_piece_normal.py
+++ b/tensorflow_probability/python/distributions/two_piece_normal.py
@@ -40,93 +40,6 @@ __all__ = [
 NUMPY_MODE = False
 
 
-def _numpy_cast(x, dtype):
-  # TODO(b/223684173): Many special math routines don't respect the input dtype.
-  if NUMPY_MODE:
-    return tf.cast(x, dtype)
-  else:
-    return x
-
-
-def standardize(value, loc, scale, skewness):
-  """Apply mean-variance-skewness standardization to input `value`.
-
-  Note that scale and skewness can be negative.
-
-  Args:
-    value: Floating-point tensor; the value(s) to be standardized.
-    loc: Floating-point tensor; the location(s) of the distribution(s).
-    scale: Floating-point tensor; the scale(s) of the distribution(s).
-    skewness: Floating-point tensor; the skewness(es) of the distribution(s).
-
-  Returns:
-    A tensor with shape broadcast according to the arguments.
-  """
-  return (value - loc) / tf.math.abs(scale) * tf.math.abs(
-      tf.where(value < loc, skewness, tf.math.reciprocal(skewness)))
-
-
-def cdf(value, loc, scale, skewness):
-  """Compute cumulative distribution function of Two-Piece Normal distribution.
-
-  Note that scale and skewness can be negative.
-
-  Args:
-    value: Floating-point tensor; where to compute the cdf.
-    loc: Floating-point tensor; the location(s) of the distribution(s).
-    scale: Floating-point tensor; the scale(s) of the distribution(s).
-    skewness: Floating-point tensor; the skewness(es) of the distribution(s).
-
-  Returns:
-    A tensor with shape broadcast according to the arguments.
-  """
-  one = tf.constant(1., dtype=loc.dtype)
-  two = tf.constant(2., dtype=loc.dtype)
-
-  z = standardize(value, loc=loc, scale=scale, skewness=skewness)
-  normal_cdf = _numpy_cast(special_math.ndtr(z), loc.dtype)
-
-  squared_skewness = tf.math.square(skewness)
-  return tf.math.reciprocal(one + squared_skewness) * tf.where(
-      z < 0.,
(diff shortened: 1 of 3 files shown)


SiegeLordEx:
Thanks for this. I'll take a look at this by EOW.

tensorflow_probability/python/distributions/two_piece_normal.py
```diff
@@ -490,3 +393,301 @@ def _parameter_control_dependencies(self, is_init):
...
+
+def _two_piece_normal_sample_bwd(_, aux, dy):
+  """The gradients of Two Piece Normal samples w.r.t. skewness."""
+  skewness, samples = aux
+  broadcast_skewness = tf.broadcast_to(skewness, ps.shape(samples))
+
+  grad = dy * _two_piece_normal_sample_gradient(broadcast_skewness, samples)
+  # Sum over the sample dimensions. Assume that they are always the first
+  # ones.
+  num_sample_dimensions = (tf.rank(broadcast_skewness) -
```
SiegeLordEx, line 604: In this function, you should use `ps.rank` and `ps.range`, as you're dealing with shapes and we want those to be statically propagated if possible.

tensorflow_probability/python/distributions/two_piece_normal_test.py
```diff
@@ -348,6 +350,55 @@ def sampler(loc, scale, skewness):
...
+    """Test the gradients of the samples w.r.t. skewness."""
+    sample_shape = [int(2e5)]
+    seed = test_util.test_seed()
+
+    def sampler(skewness):
+      loc = tf.constant(0., self.dtype)
+      scale = tf.constant(1., self.dtype)
+      dist = tfd.TwoPieceNormal(
+          loc, scale=scale, skewness=skewness, validate_args=True)
+      return tf.reduce_mean(dist.sample(sample_shape, seed=seed))
```
SiegeLordEx, line 364: Can you add some non-linearity after the mean computation (tanh or something), to make sure this is checking the that the backwards pass gradients are correctly incorporated.

tensorflow_probability/python/distributions/two_piece_normal_test.py
```diff
@@ -348,6 +350,55 @@ def sampler(loc, scale, skewness):
...
+        np.linspace(0.1, stop=10., num=n_params), dtype=self.dtype)
+    skewness = tf.constant(
+        np.linspace(0.75, stop=1.33, num=n_params), dtype=self.dtype)
+
+    seed = test_util.test_seed()
+
+    def sampler(loc, scale):
+      dist = tfd.TwoPieceNormal(
+          loc=loc, scale=scale, skewness=skewness, validate_args=True)
+      return dist.sample(sample_shape, seed=seed)
```
SiegeLordEx, line 391: Same non-linearity suggestion here as well (although here you want something that's easy to analytically handle).

SiegeLordEx (changes requested):
This looks excellent. Just a few minor suggestions.

tensorflow_probability/python/distributions/two_piece_normal.py
leandrolcampos: Done.

tensorflow_probability/python/distributions/two_piece_normal_test.py
leandrolcampos: As in `truncated_normal_test.testReparametrizable`, I used the nonlinear function `tf.abs` inside `tf.reduce_mean`. Is this OK?

tensorflow_probability/python/distributions/two_piece_normal_test.py
leandrolcampos: Here I used the nonlinear function `tf.math.exp`.

tensorflow_probability/python/distributions/two_piece_normal_test.py
SiegeLordEx: I think it should be okay. The goal is to get `dy` to not always be 1 and I think abs flips it to -1 sometimes?

leandrolcampos:
Thanks for your suggestions, @SiegeLordEx! It's a pleasure to learn from you!

SiegeLordEx (approved):

SiegeLordEx:
Thanks for the contribution! I'll send it along to our internal system now, as before.

SiegeLordEx:
In internal review we focused on the numerical stability of the gradient implementation, and we made the following changes:

- After a bit of expansion and simplification, we reduced the left and right gradient expressions to be expressed in terms of the Mills ratio, which we could compute accurately via `erfcx` special function.
- We applied the double-where trick to avoid NaNs in the second derivative (see https://github.com/tensorflow/probability/blob/main/discussion/where-nan.pdf)

leandrolcampos:
> In internal review we focused on the numerical stability of the gradient implementation, and we made the following changes:
> 
> * After a bit of expansion and simplification, we reduced the left and right gradient expressions to be expressed in terms of the Mills ratio, which we could compute accurately via `erfcx` special function.
> * We applied the double-where trick to avoid NaNs in the second derivative (see https://github.com/tensorflow/probability/blob/main/discussion/where-nan.pdf)

Hi @SiegeLordEx,

Thank you so much for the improvements in the gradient implementation. I'll study the changes you made and the material you shared about the double-where trick. As I said before, it's a pleasure to learn from you and contribute to TensorFlow Probability!

PR #1550, opened 2022-04-12
Empirical distributions quantiles

### LINKED ISSUE #1549: quantiles not implemented for Empirical distribution
Quantiles are not currently implemented for `tensorflow_probability.distributions.Empirical`
To reproduce: 

```python
samples = [1.,2.,3.,2.,1.]
dist = tfp.distributions.Empirical(samples=samples)
dist.quantile(0.5)

# raises: NotImplementedError
```

This should be an easy fix, using the already available in tensorflow_probability functions from the `tensorflow_probability.stats` module.

ltsaprounis:
Fixes #1549

Implemented the quantile method for the Empirical distribution class by interfacing the `tfp.stats.percentile` method and scaling the input appropriately. I think that the tests pass locally. I've tested using the command below: 
`bazel test -c opt tensorflow_probability/python/distributions:normal_test \
    --test_arg=--test_regex=EmpiricalScalarTest`

Let me know if I need to add more tests in the `EmpiricalNdTest` and the `EmpiricalVectorTest` classes. 

Thanks!

### DIFF
--- a/tensorflow_probability/python/distributions/empirical.py
+++ b/tensorflow_probability/python/distributions/empirical.py
@@ -219,12 +219,6 @@ class Empirical(distribution.AutoCompositeTensorDistribution):
     return tf.sqrt(var)
   
   def _quantile(self, value, samples=None, **kwargs):
-    if value > 1 or value < 0:
-      raise ValueError(
-        "Quantile values in tensorflow_probability."
-        "distributions.Empirical.quantile must be between 0 and 1."
-      )
-
     if samples is None:
       samples = tf.convert_to_tensor(self._samples)
 
--- a/tensorflow_probability/python/distributions/empirical.py
+++ b/tensorflow_probability/python/distributions/empirical.py
@@ -25,7 +25,7 @@ from tensorflow_probability.python.internal import reparameterization
 from tensorflow_probability.python.internal import samplers
 from tensorflow_probability.python.internal import tensor_util
 from tensorflow_probability.python.internal import tensorshape_util
-from tensorflow_probability.python.stats import percentile
+from tensorflow_probability.python import stats
 
 __all__ = [
     'Empirical'
@@ -228,7 +228,7 @@ class Empirical(distribution.AutoCompositeTensorDistribution):
     if samples is None:
       samples = tf.convert_to_tensor(self._samples)
 
-    return percentile(x=samples, q=value * 100, axis=self._samples_axis, **kwargs)
+    return stats.percentile(x=samples, q=value * 100, axis=self._samples_axis, **kwargs)
 
   def _sample_n(self, n, seed=None):
     samples = tf.convert_to_tensor(self._samples)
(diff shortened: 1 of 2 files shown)


tensorflow_probability/python/distributions/empirical.py
```diff
@@ -25,7 +25,7 @@
 from tensorflow_probability.python.internal import samplers
 from tensorflow_probability.python.internal import tensor_util
 from tensorflow_probability.python.internal import tensorshape_util
-
+from tensorflow_probability.python.stats import percentile
```
SiegeLordEx, line 28: Our style guide forbids us from importing functions directly, so this should just import stats and then later on the call should be `stats.percentile`

tensorflow_probability/python/distributions/empirical.py
```diff
@@ -217,6 +217,18 @@ def _stddev(self):
     r = samples - tf.expand_dims(self._mean(samples), axis=axis)
     var = tf.reduce_mean(tf.square(r), axis=axis)
     return tf.sqrt(var)
+  
+  def _quantile(self, value, samples=None, **kwargs):
+    if value > 1 or value < 0:
```
SiegeLordEx, line 222: You don't need to do this check here, it's already handled by the base Distribution class.

SiegeLordEx (changes requested):
Thanks for the change. Sorry for the delay, some weird issue with the CLA bot not telling us that the CLA was signed. I left a few comments.

ltsaprounis:
> Thanks for the change. Sorry for the delay, some weird issue with the CLA bot not telling us that the CLA was signed. I left a few comments.

Thanks for the review! I addressed all the requested changes and the [linting checks](https://github.com/tensorflow/probability/blob/main/CONTRIBUTING.md#helper-scripts)  from the CONTRIBUTING.md now pass locally.

SiegeLordEx (approved):

SiegeLordEx:
Thanks! I'll send this along to our internal review system and we'll get this merged once that is done.

ltsaprounis:
According to the [CI logs](https://github.com/tensorflow/probability/runs/6359160764?check_suite_focus=true#step:4:1860) the test I wrote for the quantiles `test_empirical_quantiles` is failing and the following error is raised: 
`AttributeError: module 'tensorflow_probability.python.internal.backend.numpy.v2' has no attribute 'bfloat16'`

@SiegeLordEx - Do you have any idea what this means? I'm a bit stuck on this...

SiegeLordEx:
The percentile function is erroneously exposed by the numpy backend. What I'll do is just disable the test and the quantile implementation under numpy.

ltsaprounis:
@SiegeLordEx - thanks for merging this! 

> The percentile function is erroneously exposed by the numpy backend. What I'll do is just disable the test and the quantile implementation under numpy.

Any idea why this happens? 
Apologies if this sounds naive, I'm not that familiar with the backend of tensorflow probability, but numpy has a percentile function of it's own, is there a way to resolve this by replacing the percentile function with the native numpy function?

SiegeLordEx:
TFP has its own [percentile function implementation](https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/stats/quantiles.py#L428) which supports batching and (outside of numpy, were it to work) gradients.

Currently the module where the TFP's percentile is defined is erroneously exposed under the numpy backend. This is because of a misconfiguration of the [bazel target](https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/stats/BUILD#L157), which should ideally indicate that it doesn't work under numpy. Unfortunately it's tricky to fix, since there's a bit of a cascading set of changes that would need to be made to change just that one target. We haven't had the time to do that.

PR #1572, opened 2022-06-06
Update vae.py link to an active link

LukeWood:
Update vae.py link to an active link.  the current link points to a missing file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ for end-to-end examples. It includes tutorial notebooks such as:
 
 It also includes example scripts such as:
 
-* [Variational Autoencoders](https://github.com/tensorflow/probability/tree/main/tensorflow_probability/examples/vae.py).
+* [Variational Autoencoders](https://github.com/tensorflow/probability/tree/main/tensorflow_probability/examples/jupyter_notebooks/Probabilistic_Layers_VAE.ipynb).
   Representation learning with a latent code and variational inference.
 * [Vector-Quantized Autoencoder](https://github.com/tensorflow/probability/tree/main/tensorflow_probability/examples/vq_vae.py).
   Discrete representation learning with vector quantization.


ColCarroll:
Thanks for this -- we should definitely do _something_ with it. vae.py used to exist, but the keras API changed enough that it no longer worked. In the same section, the LDA link is also broken (for the same reason). I think I'd prefer to just remove both links from the README for now -- do you have time to update the PR to do that?

LukeWood:
> Thanks for this -- we should definitely do _something_ with it. vae.py used to exist, but the keras API changed enough that it no longer worked. In the same section, the LDA link is also broken (for the same reason). I think I'd prefer to just remove both links from the README for now -- do you have time to update the PR to do that?

Sure, give me a few minutes to get to my desk.

LukeWood:
> Thanks for this -- we should definitely do _something_ with it. vae.py used to exist, but the keras API changed enough that it no longer worked. In the same section, the LDA link is also broken (for the same reason). I think I'd prefer to just remove both links from the README for now -- do you have time to update the PR to do that?

Curious what the API change was that broke it.  We typically hold backwards compatibility in the highest regard. (Keras team member here for reference)

ColCarroll:
Sorry, pointed the (non-blaming!) finger too soon: [this was the commit](https://github.com/tensorflow/probability/commit/5ade5823b72cd0c4bcb82d3f490e1eb43e5a3921) which was cleanup after getting rid of tensorflow Estimator module.

ColCarroll (approved):

LukeWood:
> Sorry, pointed the (non-blaming!) finger too soon: [this was the commit](https://github.com/tensorflow/probability/commit/5ade5823b72cd0c4bcb82d3f490e1eb43e5a3921) which was cleanup after getting rid of tensorflow Estimator module.

Absolutely no worries, I was genuinely curious as we do occasionally put out breaking changes.  If it was due to one of those I was going to offer to fix the scripts, as I am extremely familiar with the workarounds to get around our breaking changes.

Estimator on the other hand, I cannot offer up fixing those scripts.

Cheers!

PR #1577, opened 2022-06-18
Fix: typo in docstring

RonsenbergVI:
The docstring contained a typo:

An exception is raised if `self.allow_nan_stats=False` when df <=1


emilyfertig (approved):
Thank you @RonsenbergVI !

PR #1580, opened 2022-06-25
Improve the partial derivatives of tfp_math.betainc

leandrolcampos:
This PR tries to improve the partial derivatives of ` tfp_math.betainc(a, b, x)` with respect to `a` and `b`.

Some of the improvement opportunities I identified in the current implementation of these partial derivatives are:
- When dtype is `float64`:
  - Moderate loss of accuracy when `a` and `b` are uniformly distributed in the interval (0, 10).
  - When `a` and `b` are very small, gradients contain a very high proportion of NaN.
- When dtype is `float32`:
  - Heavy loss of accuracy in all domains evaluated.
  - Gradients contain NaN (sometimes infinite as well) in all domains evaluated.
  - When `a` and `b` are very small, gradients contain only NaN.

Please see the following notebook:
https://gist.github.com/leandrolcampos/c5a77b0f24ee8d2a38c9ea2f8ddc3c77

The alternative implementation I propose addresses all these improvement opportunities. Please see the following notebook:
https://gist.github.com/leandrolcampos/776048810f9c4ae18fc9598f54543057

### DIFF
--- a/tensorflow_probability/python/math/special.py
+++ b/tensorflow_probability/python/math/special.py
@@ -107,7 +107,7 @@ def _betainc_naive(a, b, x):
 def _betainc_even_partial_numerator(iteration, a, b, x, dtype):
   """Even partial numerator used in the continued fraction for betainc."""
   # This function computes the partial numerator d_{2m} that is specified
-  # here: https://dlmf.nist.gov/8.17#E23
+  # here: https://dlmf.nist.gov/8.17.E23
   one = tf.constant(1., dtype=dtype)
   two = tf.constant(2., dtype=dtype)
 
@@ -126,7 +126,7 @@ def _betainc_even_partial_numerator(iteration, a, b, x, dtype):
 def _betainc_odd_partial_numerator(iteration, a, b, x, dtype):
   """Odd partial numerator used in the continued fraction for betainc."""
   # This function computes the partial numerator d_{2m + 1} that is specified
-  # here: https://dlmf.nist.gov/8.17#E23
+  # here: https://dlmf.nist.gov/8.17.E23
   one = tf.constant(1., dtype=dtype)
   two = tf.constant(2., dtype=dtype)
 
@@ -283,7 +283,7 @@ def _betainc_der_continued_fraction(a, b, x, dtype, use_continued_fraction):
   # This continued fraction expansion of betainc converges rapidly
   # for x < (a - 1) / (a + b - 2). For x >= (a - 1) / (a + b - 2),
   # we can obtain an equivalent computation by using the symmetry
-  # relation given here: https://dlmf.nist.gov/8.17#E4
+  # relation given here: https://dlmf.nist.gov/8.17.E4
   #   betainc(a, b, x) = 1 - betainc(b, a, 1 - x)
   use_symmetry_relation = (x >= (a - one) / (a + b - two))
   a_orig = a
@@ -317,7 +317,7 @@ def _betainc_der_power_series(a, b, x, dtype, use_power_series):
   """Returns the partial derivatives of betainc with respect to a and b."""
   # This function evaluates betainc(a, b, x) by its series representation:
   #   x ** a * 2F1(a, 1 - b; a + 1; x) / (a * B(a, b)) ,
-  # where 2F1 is defined here: http://dlmf.nist.gov/15.2.i
+  # where 2F1 is the Gauss series as defined here: http://dlmf.nist.gov/15.2.i
   # We apply this function when the input (a, b, x) satisfies at least one
   # of the following conditions:
   #   C1: (x < a / (a + b)) & (b * x <= 1) & (x <= 0.95)
@@ -334,7 +334,7 @@ def _betainc_der_power_series(a, b, x, dtype, use_power_series):
   safe_x = tf.where(use_power_series, x, half)
 
   # When the condition C1 is false, we apply the symmetry relation given
-  # here: https://dlmf.nist.gov/8.17#E4
+  # here: https://dlmf.nist.gov/8.17.E4
   #   betainc(a, b, x) = 1 - betainc(b, a, 1 - x)
   use_symmetry_relation = (safe_x >= safe_a / (safe_a + safe_b))
   safe_a_orig = safe_a
--- a/tensorflow_probability/python/math/special.py
+++ b/tensorflow_probability/python/math/special.py
@@ -240,11 +240,11 @@ def _betainc_der_continued_fraction(a, b, x, dtype, where):
   # for x < (a - 1) / (a + b - 2). For x >= (a - 1) / (a + b - 2),
   # we can obtain an equivalent computation by using the symmetry
   # relation given here: https://dlmf.nist.gov/8.17#E4
-  apply_symmetry = (x >= (a - one) / (a + b - two))
+  use_symmetry_relation = (x >= (a - one) / (a + b - two))
   a_orig = a
-  a = tf.where(apply_symmetry, b, a)
-  b = tf.where(apply_symmetry, a_orig, b)
-  x = tf.where(apply_symmetry, one - x, x)
+  a = tf.where(use_symmetry_relation, b, a)
+  b = tf.where(use_symmetry_relation, a_orig, b)
+  x = tf.where(use_symmetry_relation, one - x, x)
 
   cf, cf_grad_a, cf_grad_b = _betainc_modified_lentz_method(
       a, b, x, dtype, where)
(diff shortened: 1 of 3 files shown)


tensorflow_probability/python/math/special.py
```diff
@@ -88,280 +88,378 @@ def atan_difference(x, y, name=None):
...
+    max_iterations = 200
+    tolerance = eps
+  else:
+    max_iterations = 600
+    tolerance = tf.constant(3., dtype=dtype) * eps
+
+  small = tf.sqrt(tiny)
+
+  def continued_fraction_evaluation(should_stop, iteration, values, gradients):
+    c, d, f = values
```
srvasude, line 168: Can these be renamed to be more descriptive?

tensorflow_probability/python/math/special.py
```diff
@@ -88,280 +88,378 @@ def atan_difference(x, y, name=None):
...
+  apply_symmetry = (x >= (a - one) / (a + b - two))
+  a_orig = a
+  a = tf.where(apply_symmetry, b, a)
+  b = tf.where(apply_symmetry, a_orig, b)
+  x = tf.where(apply_symmetry, one - x, x)
+
+  cf, cf_grad_a, cf_grad_b = _betainc_modified_lentz_method(
+      a, b, x, dtype, where)
+
+  more_terms = tf.math.exp(
```
srvasude, line 252: Rename? Maybe normalization? Here and elsewhere.

tensorflow_probability/python/math/special.py
```diff
@@ -88,280 +88,378 @@ def atan_difference(x, y, name=None):
...
+    dpdb, da, db = gradients
+
+    x_div_n = safe_x / n
+    factor = (n - safe_b) * x_div_n
+    apn = safe_a + n
+
+    new_product = product * factor
+    term = new_product / apn
+    new_dpdb = factor * dpdb - product * x_div_n
+    new_da = da - new_product / (apn * apn)
```
srvasude, line 317: Use `tf.math.square`

tensorflow_probability/python/math/special.py
```diff
@@ -88,280 +88,378 @@ def atan_difference(x, y, name=None):
...
+  one = tf.constant(1., dtype=dtype)
+
+  # Avoid returning NaN or infinity when the input does not satisfy either
+  # C1 or C2.
+  safe_a = tf.where(where, a, half)
+  safe_b = tf.where(where, b, half)
+  safe_x = tf.where(where, x, half)
+
+  # When the condition C1 is false, we apply the symmetry relation given
+  # here: http://dlmf.nist.gov/8.17.E4
```
srvasude, line 292: It would be nice (here and elsewhere) to inline the relationship, that way it is easier to see at a glance what the variable / code is referring to.

srvasude (commented):
Hi, thanks for this awesome PR! Overall looks good to me, and most of my comments have to do with naming (basically preferring more verbose names over short names so as to keep the code more maintainable).

tensorflow_probability/python/math/special.py
```diff
@@ -88,280 +88,378 @@ def atan_difference(x, y, name=None):
...
+
+    db = -a_plus_n * x / denominator
+    value =  db * a_plus_b_plus_n
+    da = -value * ((a_plus_2n + a_plus_2n_plus_one) / denominator) - x * (
+        two * a_plus_n + b) / denominator
+    gradient = tf.concat([da, db], axis=-1)
+    return value, gradient
+
+  # Assume iteration is always greater than 1.
+  return tf.cond(
```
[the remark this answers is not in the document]
leandrolcampos: If we use `tf.where` instead of `tf.cond`, I understand that the tuple `(values, gradients)` will be computed twice for each call of `_betainc_nth_partial_numerator`, regardless of whether `iteration` is odd or even. Is my understanding correct? If so, why should we prefer `tf.where` over `tf.cond`?

tensorflow_probability/python/math/special.py
leandrolcampos: Sorry for the non-suggestive names. Here I follow the notation used on the Wikipedia page about the [Lentz algorithm](https://en.wikipedia.org/wiki/Lentz%27s_algorithm). On this page, `f_n` is the convergent `f_n = A_n / B_n = f_{n-1} * C_n * D_n` and the quantities `C_n` and `D_n` are the ratios `C_n = A_n / A_{n-1}` and `D_n = B_{n-1} / B_n`. The partial numerator `A_n` and denominator `B_n` are given by the [fundamental recurrence formulas](https://en.wikipedia.org/wiki/Generalized_continued_fraction#Fundamental_recurrence_formulas). Do you have any suggestions?

Maybe:
`convergent` for `f`
`ratio_numerators` for `c`
`ratio_denominators` for `d`
`dconvergent` for `f_grad`
`dratio_numerators` for `c_grad`
`dratio_denominators` for `d_grad`

tensorflow_probability/python/math/special.py
leandrolcampos: I'm not sure if I understood this comment. Is it about making the symmetry relationship explicit in the code comment, instead of using a link?

leandrolcampos:
Hi, @srvasude! Many thanks for your time reviewing this PR. I have a few questions about your comments above. Can you take a look at them?

tensorflow_probability/python/math/special.py
srvasude: Sorry, yes I mean making it explicit in the code comment (and leaving the link in case someone wants to read more).

tensorflow_probability/python/math/special.py
srvasude: Those names sound good to me! I think having a comment which specifies that map (in case someone is familiar with Lentz's algorithm) would also be welcome (so using the more descriptive names, but saying that ratio_numerators = C_n in a comment.

tensorflow_probability/python/math/special.py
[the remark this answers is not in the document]
srvasude: In general we prefer using `tf.where`, especially in cases like this that are relatively cheap. `tf.cond` creates a runtime branch to decide which expression to compute. This can be expensive, and often `tf.where` (which is branchless) will be more performant on accelators like GPU or Cloud TPU.

leandrolcampos:
Hi, @srvasude! I addressed all of your comments. Can you please review the PR again?

After refactoring the continued fraction evaluation procedures, `tf.cond` is no longer needed.

Thank you very much for your guidance!

tensorflow_probability/python/math/special.py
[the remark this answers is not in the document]
leandrolcampos: After refactoring the continued fraction evaluation procedures, `tf.cond` is no longer needed.

tensorflow_probability/python/math/special_test.py
```diff
@@ -194,27 +196,300 @@ def testBetaincBroadcast(self):
...
+
+    self._testBetaincDerivative(
+        a, b, x, rtol_a=0.05, rtol_b=0.05, rtol_x=1e-5)
+
+  @test_util.numpy_disable_gradient_test
+  def testBetaincDerivativeSmallFloat(self):
+    strm = test_util.test_seed_stream()
+    a = tfp.distributions.Uniform(
+        high=np.float32(5.)).sample(100, strm())
+    a = self.evaluate(a)
```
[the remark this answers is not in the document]
leandrolcampos: Done.

tensorflow_probability/python/math/special.py
```diff
@@ -88,280 +88,378 @@ def atan_difference(x, y, name=None):
...
-  return tf.math.exp(result)
+  # The partial derivative of betainc with respect to x can be obtained
+  # directly using the expression given here:
+  # http://functions.wolfram.com/06.21.20.0001.01
+  grad_x = tf.math.exp(
+      tf.math.xlogy(a - one, x) + tf.math.xlog1py(b - one, -x) - lbeta(a, b))
+
+  # The partial derivatives of betainc with respect to a and b are computed
+  # using forward mode.
+  region_ps = ((x < a / (a + b)) & (b * x <= 1.) & (x <= 0.95) |
```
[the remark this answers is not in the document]
leandrolcampos: Done. I replaced `region_ps` by `use_power_series`.

tensorflow_probability/python/math/special.py
leandrolcampos: Done.

tensorflow_probability/python/math/special.py
leandrolcampos: Done (here and elsewhere).

tensorflow_probability/python/math/special.py
leandrolcampos: Done.

leandrolcampos:
Hi, @srvasude! If I still have a chance, I'd like to ask you two more questions.

1. I noticed that I should have used 4 spaces for indentation on lines 218 and 222. If I fix this now, GitHub will run all the checks just for that. Can you fix this during the internal review? Or do you prefer that I do it myself?

2. Following the current implementation, I call `tf.convert_to_tensor` on `_betainc_naive`, `_betainc_partials`, and on `betainc`. Is this necessary? Or is it enough to just call `tf.convert_to_tensor` in `betainc`, immediately after name_scope?

All the best,

srvasude (approved):
Thanks so much for these changes, and the thought and care in your analysis!

I'll be merging these in (and looking forward to any future work on this and related functions :) ).

Issue #1587, opened 2022-07-15
Anomaly threshold does not affect is_anomaly, lower_limit and upper_limit in python.sts.anomaly_detection.detect_anomalies


tarstars:
In the following function call we are to pass anomaly_threshold in order to override default value 0.01

https://github.com/tensorflow/probability/blob/42cb6741cccc19b5af3f060c1cce139d29158149/tensorflow_probability/python/sts/anomaly_detection/anomaly_detection_lib.py#L133

PR #1588, opened 2022-07-15
issue 1587 pass anomaly_threshold parameter

tarstars:
In order to override the default value of anomaly_threshold variable, we are to pass this value as a parameter

Fixes https://github.com/tensorflow/probability/issues/1587


SiegeLordEx (approved):
Thanks!

Issue #1591, opened 2022-07-25
Copula primer mentions Uniform marginals, but example does not have uniform marginals


gmjw:
[TFP is awesome, thanks so much for all the amazing functionality!]

I've been reading the copulas primer at https://www.tensorflow.org/probability/examples/Gaussian_Copula

Twice in the first couple of paragraphs it mentions that a copula is a distribution on `[0, 1]^n` with _uniform marginals_.

> a copula is a multivariate distribution C(U_1, U_2, ...) such that marginalizing gives U_i ~ Uniform(0, 1).

> the Gaussian Copula is a distribution over the unit hypercube `[0, 1]^n` with uniform marginals.

However, in the example just below, the copula pdf `pdf = GaussianCopulaTriL(...)` **doesn't have uniform marginals**. The contour plot [here](https://www.tensorflow.org/static/probability/examples/Gaussian_Copula_files/output_JbLhzyye9c49_0.png) is clearly non-uniform in the y direction (the marginal is Normal-like, not Uniform).

Would anyone be able to explain why this is?

gmjw:
Ah, I think the `scale_tril` argument to the GaussianCopulaTriL is wrong. 

I think 
```
 scale_tril=[[1., 0.8], [0., 0.6]],
```

should be 
```
scale_tril=[[1., 0], [0.8, 0.6]],
```

This seems to fix the example to have uniform marginals. I'll stick up a PR.

PR #1593, opened 2022-07-25
Fix `scale_tril` argument in Gaussian_Copula example notebook

gmjw:
Fix `scale_tril` argument, so marginals are actually uniform, as stated in the primer text.

See https://github.com/tensorflow/probability/issues/1591

### DIFF
--- a/tensorflow_probability/examples/jupyter_notebooks/Gaussian_Copula.ipynb
+++ b/tensorflow_probability/examples/jupyter_notebooks/Gaussian_Copula.ipynb
@@ -166,7 +166,7 @@
         "\n",
         "pdf = GaussianCopulaTriL(\n",
         "    loc=[0., 0.],\n",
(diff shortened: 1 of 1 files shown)


gmjw:
I wanted to mention that I'm not sure how to recreate the image that follows on from the corrected code snipped in the example. Perhaps it will be done automatically? I'm not too familiar with the repo/release setup, sorry!

The image in question is https://www.tensorflow.org/static/probability/examples/Gaussian_Copula_files/output_JbLhzyye9c49_0.png

shown on the Copulas Primer page at https://www.tensorflow.org/probability/examples/Gaussian_Copula

With the changes in this PR, the image should look quite different, something more like this:
image

SiegeLordEx:
I think you'd need to re-run the colab and then download and add the altered colab file into your pull request.

Issue #1581, opened 2022-06-26
The inverse of the regularized incomplete beta function


leandrolcampos:
Hello TFP community,

I’d like to work on the implementation of the inverse of the regularized incomplete beta function (`special.betaincinv` in SciPy).

To do this, I plan to study the following references:

- Majumder, Kantilal L., and G. P. Bhattacharjee. “Algorithm AS 64: Inverse of the incomplete beta function ratio.” Journal of the Royal Statistical Society. Series C (Applied Statistics) 22.3 (1973): 411-414. Link: https://www.jstor.org/stable/2346798.

- Cran, G. W., K. J. Martin, and G. E. Thomas. “Remark as r19 and algorithm as 109: A remark on algorithms: As 63: The incomplete beta integral as 64: Inverse of the incomplete beta function ratio.” Journal of the Royal Statistical Society. Series C (Applied Statistics) 26.1 (1977): 111-114. Link: https://www.jstor.org/stable/2346887.

- Berry, K. J., and P. W. Mielke. “A remark on algorithm AS 109: inverse of the incomplete beta function ratio.” Applied statistics 39.2 (1990): 309-310. Link: https://www.jstor.org/stable/2347779.

- YouTube video “Newton Bisection Hybrid (Newt-Safe)”.  Link: https://youtu.be/FD3BPTMGJds.

For the partial derivatives, I can take advantage of the fact that this function and `tfp_math.betainc` are inverses of each other (the same strategy used for computing the gradients of `tfp_math.igammainv`). By the way, I worked on some improvements for evaluating the partial derivatives of `tfp_math.betainc`: see the PR https://github.com/tensorflow/probability/pull/1580.

Can I work on this task?

All the best,

Leandro Campos

leandrolcampos:
Hi, @srvasude! 

I hope this message finds you well.

I already have an implementation of `tfp_math.betaincinv`. In my opinion, its numerical accuracy and stability are good enough. Because it depends on the PR #1580, which hasn't been merged yet, I don't know how to proceed to create a new PR with this implementation. Should I wait for the PR #1580 to be merged? Should I create a new PR based on a branch created from the branch of the PR #1580? Another alternative?

All the best,

brianwa84:
I'd suggest waiting until #1580 gets merged. I think it's close.


Brian Patton | Software Engineer | ***@***.***

srvasude:
Sorry for the delay, the PR has been merged.

leandrolcampos:
[...]

PR #1597, opened 2022-08-03
Add `tfp.math.betaincinv`

leandrolcampos:
This PR provides an implementation for `tfp.math.betaincinv`, the inverse of the regularized incomplete beta function (`tfp.math.betainc`) with respect to `x`. It addresses the issue #1581.

This implementation is based on ideas and equations available in the following references:
[1] Milton Abramowitz and Irene A. Stegun. Handbook of Mathematical Functions with Formulas, Graphs, and Mathematical Tables. US Government Printing Office, 1964 (reprinted 1972). [[Link](https://archive.org/details/AandS-mono600)]
[2] William Press, Saul Teukolsky, William Vetterling and Brian Flannery. Numerical Recipes: The Art of Scientific Computing. Cambridge University Press, 2007 (Third Edition). [[Link](http://numerical.recipes/book/book.html)]
[3] John Maddock, Paul A. Bristow, et al. The Incomplete Beta Function Inverses. [[Link](https://www.boost.org/doc/libs/1_79_0/libs/math/doc/html/special.html)]
[4] Stephen L. Moshier. Cephes Mathematical Library. [[Link](https://netlib.org/cephes/)]

The expression of these ideas and equations as source code is my own.

In the following notebook, I measure and present the numerical accuracy of this implementation:
- [My implementation of `betaincinv`](https://gist.github.com/leandrolcampos/767818366dd8cbdffb6189aa00847d34)

As references (especially for single-precision results), I also measure and present the numerical accuracy of the following methods:
- [Cephes implementation of `betaincinv` for `float32` [my port]](https://gist.github.com/leandrolcampos/5dfbc66753310598ef531f59901a343b)
- [TFP implementation of `igammainv`](https://gist.github.com/leandrolcampos/b19177cac0e4f3f0b962edca265ec8ab)

In addition to implementing the function `tfp.math.betaincinv`, this PR also fixes a bug in the implementation of the partial derivative of `tfp.math.betainc` with respect to `x`: it's not true that this partial is equal to `0` whenever the parameter `x` is equal to `0` or `1`.

### DIFF
--- a/tensorflow_probability/python/math/special.py
+++ b/tensorflow_probability/python/math/special.py
@@ -620,6 +620,136 @@ def _betaincinv_initial_approx(a, b, y, dtype):
   return tf.clip_by_value(result, tiny, one - eps)
 
 
+def _betaincinv_computation(a, b, y):
+  """Returns the inverse of `betainc(a, b, x)` with respect to `x`."""
+  dtype = dtype_util.common_dtype([a, b, y], tf.float32)
+  numpy_dtype = dtype_util.as_numpy_dtype(dtype)
+  zero = tf.constant(0., dtype=dtype)
+  tiny = tf.constant(np.finfo(numpy_dtype).tiny, dtype=dtype)
+  eps = tf.constant(np.finfo(numpy_dtype).eps, dtype=dtype)
+  half = tf.constant(0.5, dtype=dtype)
+  one = tf.constant(1., dtype=dtype)
+  two = tf.constant(2., dtype=dtype)
+  halley_correction_min = tf.constant(0.5, dtype=dtype)
+  halley_correction_max = tf.constant(1.5, dtype=dtype)
+
+  a = tf.convert_to_tensor(a, dtype=dtype)
+  b = tf.convert_to_tensor(b, dtype=dtype)
+  y = tf.convert_to_tensor(y, dtype=dtype)
+
+  broadcast_shape = functools.reduce(
+      ps.broadcast_shape, [ps.shape(a), ps.shape(b), ps.shape(y)])
+
+  a = tf.broadcast_to(a, broadcast_shape)
+  b = tf.broadcast_to(b, broadcast_shape)
+  y = tf.broadcast_to(y, broadcast_shape)
+
+  # When tfp_math.betainc(a, b, 0.5) < y, we apply the symmetry relation given
+  # here: https://dlmf.nist.gov/8.17.E4
+  #   betainc(a, b, x) = 1 - betainc(b, a, 1 - x) .
+  # If dtype != float64, we have additional conditions to apply this relation:
+  #   (a < 1) & (b < 1) & (tfp_math.betainc(a, b, a / (a + b)) < y) .
+  error_at_half = betainc(a, b, half) - y
+  if numpy_dtype == np.float64:
+    use_symmetry_relation = (error_at_half < zero)
+  else:
+    a_and_b_are_small = (a < one) & (b < one)
+    error_at_mean = betainc(a, b, a / (a + b)) - y
+    use_symmetry_relation = (error_at_half < zero) & a_and_b_are_small & (
+        error_at_mean < zero)
+
+  a_orig = a
+  a = tf.where(use_symmetry_relation, b, a)
+  b = tf.where(use_symmetry_relation, a_orig, b)
+  y = tf.where(use_symmetry_relation, one - y, y)
+
+  a_minus_1 = a - one
+  b_minus_1 = b - one
+  one_minus_eps = one - eps
+
+  # tolerance was set by experimentation and max_iterations was taken from [4].
+  if numpy_dtype == np.float64:
+    tolerance = 1e-12
+    max_iterations = 8
+  else:
+    tolerance = 1e-6
+    max_iterations = 10
+
+  def root_finding_iteration(should_stop, low, high, candidate):
+    error = betainc(a, b, candidate) - y
+    error_over_der = error / tf.math.exp(
+        tf.math.xlogy(a_minus_1, candidate) +
+        tf.math.xlog1py(b_minus_1, -candidate) -
+        lbeta(a, b))
+    second_der_over_der = a_minus_1 / candidate - b_minus_1 / (one - candidate)
+    # Following [2, section 9.4.2, page 463], we limit the influence of the
+    # Halley's correction to the Newton's method, since this correction can
+    # reduce the Newton's region of convergence. We set minimum and maximum
+    # values for this correction by experimentation.
+    halley_correction = tf.clip_by_value(
+        one - half * error_over_der * second_der_over_der,
+        halley_correction_min,
+        halley_correction_max)
+    halley_delta = error_over_der / halley_correction
+    halley_candidate = tf.where(
+        should_stop, candidate, candidate - halley_delta)
+
+    # Fall back to bisection if the current step would take the new candidate
+    # out of bounds.
+    new_candidate = tf.where(
+        tf.less_equal(halley_candidate, low),
+        half * (candidate + low),
+        tf.where(
+            tf.greater_equal(halley_candidate, high),
+            half * (candidate + high),
+            halley_candidate))
+
+    new_delta = candidate - new_candidate
+    new_delta_is_negative = (new_delta < zero)
+    new_low = tf.where(new_delta_is_negative, candidate, low)
+    new_high = tf.where(new_delta_is_negative, high, candidate)
+
+    adjusted_tolerance = tf.maximum(tolerance * new_candidate, two * tiny)
+    should_stop = (should_stop | (tf.math.abs(new_delta) < adjusted_tolerance) |
+        tf.math.equal(new_low, new_high))
+
+    return should_stop, new_low, new_high, new_candidate
+
+  initial_candidate = _betaincinv_initial_approx(a, b, y, dtype)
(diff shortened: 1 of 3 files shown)


leandrolcampos:
I'm worried about dtype support in the implementations for `tfp.math._betainc_partials` and `tfp.math.betaincinv`. They only support `float32` and `float64` by design, like the TF C++ and XLA implementations for `betainc`. But there is an important difference between these last two implementations: the former throws an error when the input dtype is `float16` or `bfloat16`, while the latter converts a `float16` or `bfloat16` input to `float32` and then, after the computations, converts the result back to the original dtype [[link](https://github.com/tensorflow/tensorflow/blob/af2b6266a53b8bfa8535c84860313b11651b363a/tensorflow/compiler/xla/client/lib/math.cc#L1737-L1862)].

Thinking about consistency, should we cast a `float16` or `bfloat16` input to `float32` and then, after the computations, convert the result back to the original dtype?

leandrolcampos:
Hi, @srvasude!

I hope this message finds you very well.

When you have time, could you review this PR, please?

All the best,

tensorflow_probability/python/math/special.py
```diff
@@ -521,6 +545,332 @@ def betainc(a, b, x, name=None):
...
+  integral_approx_part_a = tf.math.exp(tf.math.xlogy(a, mean) - tf.math.log(a))
+  integral_approx_part_b = tf.math.exp(tf.math.xlogy(b, mean_complement) -
+      tf.math.log(b))
+  integral_approx = integral_approx_part_a + integral_approx_part_b
+
+  # Solve Equation 6.4.8 [2, page 271] for x in the respective regimes.
+  inv_a = tf.math.reciprocal(a)
+  inv_b = tf.math.reciprocal(b)
+  result_for_small_a_or_b = tf.where(
+      tf.math.less_equal(y, (integral_approx_part_a / integral_approx)),
```
srvasude, line 619: y <= integral_approx_part_a / integral_approx. Can use `<=` and `>=` overrides here and elsewhere

tensorflow_probability/python/math/special.py
```diff
@@ -521,6 +545,332 @@ def betainc(a, b, x, name=None):
...
+      tf.math.exp(tf.math.xlogy(inv_a, y) + tf.math.xlogy(inv_a, a) +
+          tf.math.xlogy(inv_a, integral_approx)),
+      one - tf.math.exp(tf.math.xlog1py(inv_b, -y) + tf.math.xlogy(inv_b, b) +
```
srvasude, line 622: Can use `-tf.math.expm1` instead of `one - exp(...)`

tensorflow_probability/python/math/special.py
```diff
@@ -521,6 +545,332 @@ def betainc(a, b, x, name=None):
...
+#     Cephes Mathematical Library
+#     https://netlib.org/cephes/
+
+
+def _betaincinv_initial_approx(a, b, y, dtype):
+  """Computes an initial approximation for `betaincinv(a, b, y)`."""
+  numpy_dtype = dtype_util.as_numpy_dtype(dtype)
+  tiny = tf.constant(np.finfo(numpy_dtype).tiny, dtype=dtype)
+  eps = tf.constant(np.finfo(numpy_dtype).eps, dtype=dtype)
+  one = tf.constant(1., dtype=dtype)
```
srvasude, line 572: FWIW can use `numpy_dtype` instead (just my own style preference, feel free to keep as is):

`one = tf.constant(1., dtype=dtype)` vs
`one = numpy_dtype(1.)`

tensorflow_probability/python/math/special.py
```diff
@@ -521,6 +545,332 @@ def betainc(a, b, x, name=None):
...
+  two = tf.constant(2., dtype=dtype)
+  three = tf.constant(3., dtype=dtype)
+  five = tf.constant(5., dtype=dtype)
+  six = tf.constant(6., dtype=dtype)
+  min_log = tf.math.log(
```
srvasude, line 577: `min_log = numpy_dtype(np.finfo(numpy_dtype).minexp * np.log(2.))`
`max_log = numpy_dtype((np.info(numpy_dtype).maxexp - 1.) * np.log(2.))`

tensorflow_probability/python/math/special_test.py
```diff
@@ -194,6 +194,37 @@ def testBetaincBroadcast(self):
...
+    expected_result = tfp_math.betainc(
+        *[tf.cast(z, tf.float32) for z in [a, b, x]])
+    expected_result = tf.cast(expected_result, a.dtype)
+
+    self.assertAllEqual(*self.evaluate([expected_result, result]))
+
+  @test_util.disable_test_for_backend(
+      disable_numpy=True, disable_jax=False,
+      reason="Numpy does not support brain floating point.")
+  def testBetaincBrain(self):
```
srvasude, line 214: `testBetaincBFloat16`, and probably change the reason to `bfloat16`

tensorflow_probability/python/math/special.py
```diff
@@ -521,6 +545,332 @@ def betainc(a, b, x, name=None):
...
+          tf.math.xlogy(inv_b, integral_approx)))
+
+  # And when max(a, b) < 1, we use the approximation proposed by [3] for the
+  # same domain:
+  #   betaincinv(a, b, y) ~= xg / (1 + xg) ,
+  # where:
+  #   xg = (a * y * Beta(a, b)) ** (1 / a) .
+  log_xg = tf.math.xlogy(inv_a, a) + tf.math.xlogy(inv_a, y) + (
+      inv_a * lbeta(a, b))
+  xg = tf.math.exp(tf.clip_by_value(log_xg, min_log, max_log))
```
srvasude, line 632: There seems to be quite a bit of clipping in this code (here and a few places below). Are these necessary / can they be removed in favor of `tf.where` guarding code earlier up?

tensorflow_probability/python/math/special_test.py
```diff
@@ -263,15 +294,16 @@ def betainc_partials(a, b, x):
     x = np.array([0.5, 0.5, 0.5, 0.5, -1., 2.], dtype=dtype)
 
     for partial in self.evaluate(betainc_partials(a, b, x)):
-      self.assertAllClose(np.full_like(a, np.nan), partial)
+      self.assertAllEqual(np.full_like(a, np.nan), partial)
```
srvasude, line 297: can use `self.assertAllNan`

srvasude (commented):
Sorry for the delay, and thanks for this contribution! Overall looks great to me, with a few style comments and one comment (about clipping).

tensorflow_probability/python/math/special.py
leandrolcampos: Done (here and elsewhere).

tensorflow_probability/python/math/special.py
leandrolcampos: Done.

tensorflow_probability/python/math/special.py
leandrolcampos: Done (here and elsewhere). Thanks for sharing this!

tensorflow_probability/python/math/special.py
leandrolcampos: Done.

tensorflow_probability/python/math/special_test.py
leandrolcampos: Done (here and elsewhere).

tensorflow_probability/python/math/special_test.py
leandrolcampos: Done (here and elsewhere).

tensorflow_probability/python/math/special.py
leandrolcampos: There were four uses of `tf.clip_by_value` in the code (I replaced two of them). They handled cases that depend on the combination of values for the parameters `a`, `b`, and `y`. I don't know if it's possible (and convenient) to list and describe all these problematic combinations of values and then use `tf.where` to guard the code from the respective cases.

The first use of `tf.clip_by_value` (line 629) avoided overflow in `tf.math.exp(log_xg)`: for example, when `a` and `b` are very small and `x` is close enough to `1`. But in this case, I replaced `tf.clip_by_value` with `tf.math.minimum`.

The second use of `tf.clip_by_value` (line 641) ensures that the result of `_betaincinv_initial_approx` is always greater than `0` and less than `1` and, consequently, that the condition `initial_low < initial_candidate < initial_high` is always satisfied. For example, if

```
a, b, y, dtype = (3.5565727, 0.011408293, 0.130523313, float32)
```

then `initial_candidate == 1.` and `initial_candidate == initial_high`.

The reason for the third use of `tf.clip_by_value` (line 714) is commented out in the code.

The last use of `tf.clip_by_value` (line 764) prevented `one - result` from being rounded up to `1` when `use_symmetry_relation` and `result < eps`. But in this case, I replaced `tf.clip_by_value` with `tf.math.maximum`.

leandrolcampos:
Hi @srvasude,

Thank you very much for your time reviewing this PR. It's always a pleasure to learn from you!

Please let me know if I've properly addressed all your comments.

All the best,

tensorflow_probability/python/math/special.py
srvasude: Thanks for the explanation, those uses seem pretty reasonable to me.

srvasude (approved):

PR #1600, opened 2022-08-08
Enable batch support for `windowed_mean|variance`

nicolaspi:
This PR makes functions `windowed_mean` and `windowed_variance` to accept indices with batch dimensions.

Example:

```python
x = np.array([[1, 2, 3], [1, 2, 3], [1, 2, 3]], dtype=np.float32)
low_indices = [[0, 0, 0], [1, 0, 0], [2, 2, 0]]
high_indices = [[3, 3, 3], [1, 2, 3], [3, 2, 1]]
tfp.stats.windowed_mean(x, low_indices=low_indices, high_indices=high_indices, axis=1)
```
Now gives:
```
<tf.Tensor: shape=(3, 3), dtype=float32, numpy=
array([[2. , 2. , 2. ],
       [0. , 1.5, 2. ],
       [3. , 0. , 1. ]], dtype=float32)>
```
Was previously failing with:
```
tensorflow.python.framework.errors_impl.InvalidArgumentError: required broadcastable shapes [Op:SelectV2]
```


nicolaspi:
@axch I made changes in the code you authored, could you kindly have a look at this PR?
Thanks

axch:
@nicolaspi thanks for the contribution!  I am no longer an active maintainer of TFP, so I'm not really in a position to review your PR in detail (@jburnim please suggest someone?).  On a quick look, though, I see a couple potential code style issues:
- Do we need the dependency on tf.experimental.numpy?
- Do we need the special case for rank-1 indices?  Could we define a more uniform behavior instead?
- I'm guessing some of the shape munging already has relevant helpers defined elsewhere in TFP, but I don't remember off-hand
- TFP generally tries to handle static and dynamic TF shapes uniformly using [prefer_static](https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/internal/prefer_static.py) and [tensorshape_util](https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/internal/tensorshape_util.py).

nicolaspi:
Thanks for your feedback!

> * Do we need the dependency on tf.experimental.numpy?

We need specifically the `take_along_axis` function that allow to gather the slices along each batch dimensions. I replaced the 'experimental' import path with `from tensorflow.python.ops import numpy_ops`.

> * Do we need the special case for rank-1 indices?  Could we define a more uniform behavior instead?

There is two motivations for this case. First, for backward compatibility, it is equivalent to the legacy non batched usage. Second, it is the only case I can think of where the broadcast is unambiguous when `rank(indices) < rank(x)`.

> * I'm guessing some of the shape munging already has relevant helpers defined elsewhere in TFP, but I don't remember off-hand

In any case, I modified the unit tests to test against non static shapes.

> * TFP generally tries to handle static and dynamic TF shapes uniformly using [prefer_static](https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/internal/prefer_static.py) and [tensorshape_util](https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/internal/tensorshape_util.py).

I made usage of `prefer_static` whenever possible.

nicolaspi:
@jburnim can you please suggest a reviewer?
CC @axch

SiegeLordEx:
I'll take a look at this.

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -906,30 +925,28 @@ def _prepare_window_args(x, low_indices=None, high_indices=None, axis=0):
...
-  return low_indices, high_indices, low_counts, high_counts
-
-
-def _safe_average(totals, counts):
-  # This tf.where protects `totals` from getting a gradient signal
-  # when `counts` is 0.
-  safe_totals = tf.where(~tf.equal(counts, 0), totals, 0)
-  return tf.where(~tf.equal(counts, 0), safe_totals / counts, 0)
+  indices_shape = ps.shape(low_indices)
+  if ps.rank(low_indices) < ps.rank(x):
```
SiegeLordEx, line 929: We can't write this like this. prefer_static will return Tensors if the inputs are not statically known, which cannot appear in a conditional like this. This code needs to be written without the conditional, or, as a last resort, using `tf.cond`.

Alternatively, we could require that the ranks are known, but then you need to do something like:

```
indices_rank = tf.get_static_value(ps.rank(low_indices))
if indices_rank is None:
  raise ValueError(...)
```

tensorflow_probability/python/stats/sample_stats_test.py
```diff
@@ -683,6 +683,129 @@ def test_windowed_mean_corner_cases(self):
...
+      check_fn((64, 4, 8), (4, 1, 4), axis=2)
+
+    with self.assertRaises(Exception):
+      # Non broadcastable shapes
+      check_fn((64, 4, 8), (2, 4), axis=2)
+
+  def test_windowed_mean(self):
+    self.check_windowed(func=tfp.stats.windowed_mean, numpy_func=np.mean)
+
+  def test_windowed_mean_graph(self):
```
SiegeLordEx, line 797: There's no need to explicitly test this, `test_all_tf_execution_regimes` decorator already runs each test case in graph mode.

tensorflow_probability/python/stats/sample_stats_test.py
```diff
@@ -683,6 +683,129 @@ def test_windowed_mean_corner_cases(self):
...
+    check_fn((64, 4, 8), (64, 4, 64), axis=2)
+    check_fn((64, 4, 8), (1, 1, 64), axis=2)
+    check_fn((64, 4, 8), (64, 4, 4), axis=2)
+    check_fn((64, 4, 8), (1, 1, 4), axis=2)
+    check_fn((64, 4, 8), (64, 4, 8), axis=2)
+    check_fn((64, 4, 8), (16,), axis=2)
+    check_fn((64, 4, 8), (1, 4), axis=2)
+    check_fn((64, 4, 8), (64, 4), axis=2)
+
+    with self.assertRaises(Exception):
```
SiegeLordEx, line 786: This should be assertRaisesRegexp with a more specific error class and a snippet of the error message.

tensorflow_probability/python/stats/sample_stats_test.py
```diff
@@ -683,6 +683,129 @@ def test_windowed_mean_corner_cases(self):
...
+    b[np.isnan(b)] = 0  # We treat stats computed on empty sets as zeros
+    self.assertAllClose(a, b)
+
+  def _make_dynamic_shape(self, x):
+    return tf1.placeholder_with_default(x, shape=(None,)*len(x.shape))
+
+  def check_windowed(self, func, numpy_func):
+    check_fn = functools.partial(self.check_gaussian_windowed,
+                                 window_func=func, np_func=numpy_func)
+    check_fn((64, 4, 8), (128, 1, 1), axis=0)
```
SiegeLordEx, line 761: This needs to be written as parameterized tests, see this [example](https://github.com/tensorflow/probability/blob/f7da474f215ec33c53265e09882c89db84eb7bbf/tensorflow_probability/python/bijectors/masked_autoregressive_test.py#L993).

tensorflow_probability/python/stats/sample_stats_test.py
```diff
@@ -683,6 +683,129 @@ def test_windowed_mean_corner_cases(self):
         [0., 1., 1.5, 2.5], self.evaluate(tfp.stats.windowed_mean(y)))
 
 
+@test_util.test_all_tf_execution_regimes
+class WindowedStatsTest(test_util.TestCase):
+
+  def _maybe_expand_dims_to_make_broadcastable(self, x, shape, axis):
```
SiegeLordEx, line 689: These two functions are as complex as the thing we're testing. Is there any way we can write this via np.vectorize?

SiegeLordEx (changes requested):

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -14,10 +14,20 @@
...
 """Functions for computing statistics of samples."""
 
+JAX_MODE = False
+NUMPY_MODE = False
+
 # Dependency imports
 import numpy as np
 import tensorflow.compat.v2 as tf
 
+if NUMPY_MODE:
```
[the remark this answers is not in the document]
nicolaspi: Thanks for the review!

1. I don't feel comfortable rewriting `take_along_axis` as it would duplicate already existing logics, I feel like it would produce unnecessary maintenance burden.
2. What about [mapping](https://github.com/tensorflow/probability/pull/1600/commits/26f4f121a2ac4b2b14f15af38ebce944ba3f0fa8) `tensorflow.experimental.numpy` to `numpy` and `jax.numpy` backends?

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -726,11 +736,17 @@ def windowed_variance(
...
 
-  Suppose `x` has shape `Bx + [N] + E`, where the `Bx` component has
-  rank `axis`, and `low_indices` and `high_indices` broadcast to shape
-  `[M]`.  Then each element of `low_indices` and `high_indices`
-  must be between 0 and N+1, and the shape of the output will be
-  `Bx + [M] + E`.  Batch shape in the indices is not currently supported.
+  Suppose `x` has shape `Bx + [N] + E`, `low_indices` and `high_indices`
+  have shape `Bi + [M] + F`, such that `rank(Bx) = rank(Bi) = axis`.
+  Then each element of `low_indices` and `high_indices` must be
+  between 0 and N+1, and the shape of the output will be `Bx + [M] + E`.
```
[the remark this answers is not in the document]
nicolaspi: I updated to `broadcast(Bx, Bi) + [M] + broadcast(E, F)`

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -726,11 +736,17 @@ def windowed_variance(
   trailing-window estimators from some iterative process, such as the
   last half of an MCMC chain.
 
-  Suppose `x` has shape `Bx + [N] + E`, where the `Bx` component has
-  rank `axis`, and `low_indices` and `high_indices` broadcast to shape
-  `[M]`.  Then each element of `low_indices` and `high_indices`
-  must be between 0 and N+1, and the shape of the output will be
-  `Bx + [M] + E`.  Batch shape in the indices is not currently supported.
+  Suppose `x` has shape `Bx + [N] + E`, `low_indices` and `high_indices`
+  have shape `Bi + [M] + F`, such that `rank(Bx) = rank(Bi) = axis`.
```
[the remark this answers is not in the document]
nicolaspi: Please check my comment below.

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -726,11 +736,17 @@ def windowed_variance(
...
+  Then each element of `low_indices` and `high_indices` must be
+  between 0 and N+1, and the shape of the output will be `Bx + [M] + E`.
+
+  The shape `Bi + [1] + F` must be broadcastable with the shape of `x`.
+
+  If `rank(Bi + [M] + F) < rank(x)`, then the indices are expanded
```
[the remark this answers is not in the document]
nicolaspi: We specify the implicit rules we uses for broadcasting. I updated the formulation.

tensorflow_probability/python/stats/sample_stats.py
```diff
@@ -726,11 +736,17 @@ def windowed_variance(
...
-  rank `axis`, and `low_indices` and `high_indices` broadcast to shape
-  `[M]`.  Then each element of `low_indices` and `high_indices`
-  must be between 0 and N+1, and the shape of the output will be
-  `Bx + [M] + E`.  Batch shape in the indices is not currently supported.
+  Suppose `x` has shape `Bx + [N] + E`, `low_indices` and `high_indices`
+  have shape `Bi + [M] + F`, such that `rank(Bx) = rank(Bi) = axis`.
+  Then each element of `low_indices` and `high_indices` must be
+  between 0 and N+1, and the shape of the output will be `Bx + [M] + E`.
+
+  The shape `Bi + [1] + F` must be broadcastable with the shape of `x`.
```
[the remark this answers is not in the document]
nicolaspi: > This contradicts the next paragraph, no?

Yes, I reformulated.

> The batching would introduce a batch dimension on the left of those shapes: 
> Thus, the only broadcasting requirements are that Bx and Bi broadcast. I don't know where F came from.

Maybe the term 'batch' is not proper. This contribution adds the possibility to have the more general case where 
idx shape is `Bi + [M] + F`. F could be seen as 'inner batch dimensions', but here 'batch' carries a different semantic than the standard machine learning one where it is represented by outer dims.

tensorflow_probability/python/stats/sample_stats.py
nicolaspi: I added the rank check.

tensorflow_probability/python/stats/sample_stats_test.py
nicolaspi: I removed those tests.

tensorflow_probability/python/stats/sample_stats_test.py
nicolaspi: Done

tensorflow_probability/python/stats/sample_stats_test.py
nicolaspi: Done

tensorflow_probability/python/stats/sample_stats_test.py
nicolaspi: I refactored using np.vectorize, but I am not sure it is easier to read.

PR #1602, opened 2022-08-15
Update the R-hat paper reference

avehtari:
There was a reference to R-hat paper arxiv version, and I just updated it with the journal publication info.


ColCarroll (approved):
thanks! will get this merged and updated.

PR #1622, opened 2022-09-22
Fix missing args in `tfe.stats.RunningVariance.from_shape`

chrism0dwk:
Both the `event_ndims` and `name` arguments of the class method `tfp.experimental.stats.RunningVariance.from_shape` were missing, causing `TypeError: from_example() got an unexpected keyword argument 'event_ndims'`.


chrism0dwk:
N.B. this fix sorts out the same `TypeError` when using `RunningVariance.from_example()`.

SiegeLordEx:
How did you get the original error? RunningVariance is expected to have event_ndims=0, because it treats each element as an independent (co)variance estimation problem.

chrism0dwk:
@SiegeLordEx The original problem arose with
```python
tfp.experimental.stats.RunningVariance.from_example(np.array([1., 1., 1., 1.]))
```

The fundamental issue is with the Python MRO with  `RunningVariance` derived from `RunningCovariance`:

1.  `RunningVariance.from_example` calls `RunningCovariance.from_example`;
2.  `RunningCovariance.from_example` calls `RunningVariance.from_shape`;

The issue is caused because inside `RunningCovariance.from_example`, `cls` is a reference to the _derived_ class, not (as I think the author assumed) the base class.  Thus `RunningCovariance.from_example` assumes that `RunningVariance.from_shape` takes the same arguments as its own `RunningCovariance.from_shape` method.

Took me a while to revise my Python MRO semantics ;-)

PR #1639, opened 2022-10-20
Improve cdf and quantile functions of Student T distribution

leandrolcampos:
This PR tries to improve the numerical accuracy and stability of the cdf and quantile functions of the Student T distribution, especially when the degree of freedom `df` is large or when the square of the sample `t` is too small compared to `df`. It addresses both the values and the gradients of these functions.

When `df` is large, both functions and their partial derivatives suffer moderate to heavy loss of accuracy. And when `t**2` is too small compared to `df`, the partial derivatives return `NaN`.

The following notebooks present results that allow us to compare current and proposed implementations:

- [Test results for `float64` and any value of `t`](https://gist.github.com/leandrolcampos/124c971f426ec030facabcf3abbbc1e8)
- [Test results for `float64` and tiny values of `t`](https://gist.github.com/leandrolcampos/66c213b41821bf8ec1560e3dfb43db9e)
- [Test results for `float32` and any value of `t`](https://gist.github.com/leandrolcampos/93a33013746e42faa67b8462ec6f0d0f)
- [Test results for `float32` and tiny values of `t`](https://gist.github.com/leandrolcampos/118e7b3e129d4a4f32f366d366ecee7d)

All the best,


leandrolcampos:
Hi, @srvasude!

When you have time, could you please review this PR or suggest a reviewer?

All the best,

SiegeLordEx:
Thanks, this looks great. Sorry for the delay in review, we've... got no excuse, haha.

I'm convinced by and happy with the numerical precision improvements but I'm wonder what performance impact these changes have. Do you have a sense how much slower these new implementations are (I don't need precise numbers, just a ballpark estimate from your notebooks).

SiegeLordEx (approved):

leandrolcampos:
Hi @SiegeLordEx,

Thank you so much for reviewing this PR!

I compared the performance of current and proposed implementations. As expected, the former were generally faster than the latter. See the notebooks below:
- [student_t_f64_performance.ipynb](https://gist.github.com/leandrolcampos/4fd9650cbfbfd037aadb053c34f89b26)
- [student_t_f32_performance.ipynb](https://gist.github.com/leandrolcampos/a10308a8ae89cb733975b90d7f9d6ac7)

Current implementations achieved the best comparative results when I evaluated `quantile` for small values of `df` and `dtype == np.float64`: the performance difference was about 2.3x.

But the performance difference became smaller (or reversed) as the value of `df` increased, especially when I also computed the gradients. Proposed implementations achieved the best comparative results when I evaluated `cdf` and its partial derivatives for large values of `df` and `dtype == np.float32`: in this case, the proposed `cdf` was about 14.4x faster than the current `cdf`.

All the best,

SiegeLordEx:
Thanks for the performance measurements!

Issue #1497, opened 2022-01-14
Optimize `_kl_matrix_normal_matrix_normal` for `tfd.MatrixNormalLinearOperator`


fdtomasi:
Hello! I was glad to notice the introduction of the `tfd.MatrixNormal` in 0.13 (a bit late though). However, the implementation of the KL divergence still relies on the general KL implementation for `mvn_linear_operator`'s, hence there is a dense operation on the scale parameter (a Kronecker product), which can be very expensive in high dimensions. I did implement a MatrixNormal distribution before the introduction of this, and I implemented the explicit KL divergence when the LinearOperatorKronecker has only two matrices (ie, the case of MatrixNormal). The implementation is the following (based on the `_kl_brute_force` of `MultivariateNormalLinearOperator`):

```python
def _kl_matrix_normal_matrix_normal_optimized(a, b, name=None):
    """Batched KL divergence `KL(a || b)` for Matrix Normals.

    Args:
      a: Instance of `MatrixNormalLinearOperator`.
      b: Instance of `MatrixNormalLinearOperator`.
      name: (optional) name to use for created ops. Default "kl_mvn".

    Returns:
      Batchwise `KL(a || b)`.
    """

    def squared_frobenius_norm(x):
        """Helper to make KL calculation slightly more readable."""
        # http://mathworld.wolfram.com/FrobeniusNorm.html
        # The gradient of KL[p,q] is not defined when p==q. The culprit is
        # tf.norm, i.e., we cannot use the commented out code.
        # return tf.square(tf.norm(x, ord="fro", axis=[-2, -1]))
        return tf.reduce_sum(tf.square(x), axis=[-2, -1])

    with tf.name_scope(name or "kl_mn"):
        # Calculation is based on:
        # http://stats.stackexchange.com/questions/60680/kl-divergence-between-two-multivariate-gaussians
        # and,
        # https://en.wikipedia.org/wiki/Matrix_norm#Frobenius_norm
        # i.e.,
        #   If Ca = AA', Cb = BB', then
        #   tr[inv(Cb) Ca] = tr[inv(B)' inv(B) A A']
        #                  = tr[inv(B) A A' inv(B)']
        #                  = tr[(inv(B) A) (inv(B) A)']
        #                  = sum_{ij} (inv(B) A)_{ij}**2
        #                  = ||inv(B) A||_F**2
        # where ||.||_F is the Frobenius norm and the second equality follows from
        # the cyclic permutation property.
[...]

PR #1655, opened 2022-11-24
Optimize matrix_normal_linear_operator KL divergence computation.

fdtomasi:
Sorry it took me so long to open this PR!
As per the original issue #1497 , I open this PR to optimize the computation of the KL divergence between two `MatrixNormalLinearOperator`s. As in #1497, the computation is valid and gives a visible speed up.

### DIFF
--- a/tensorflow_probability/python/distributions/matrix_normal_linear_operator.py
+++ b/tensorflow_probability/python/distributions/matrix_normal_linear_operator.py
@@ -17,6 +17,7 @@
 import tensorflow.compat.v2 as tf
 
 from tensorflow_probability.python.bijectors import identity as identity_bijector
+from tensorflow_probability.python.bijectors import transpose as transpose_bijector
 from tensorflow_probability.python.distributions import distribution
 from tensorflow_probability.python.distributions import kullback_leibler
 from tensorflow_probability.python.distributions import mvn_linear_operator
@@ -25,6 +26,7 @@ from tensorflow_probability.python.internal import parameter_properties
 from tensorflow_probability.python.internal import prefer_static
 from tensorflow_probability.python.internal import reparameterization
 from tensorflow_probability.python.internal import tensor_util
+from tensorflow_probability.python.internal import tensorshape_util
 
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


tensorflow_probability/python/distributions/matrix_normal_linear_operator.py
```diff
@@ -259,6 +261,86 @@ def _parameter_control_dependencies(self, is_init):
...
+    #                  = tr[(inv(B) A) (inv(B) A)']
+    #                  = sum_{ij} (inv(B) A)_{ij}**2
+    #                  = ||inv(B) A||_F**2
+    # where ||.||_F is the Frobenius norm and the second equality follows from
+    # the cyclic permutation property.
+    k_inv_s_h = b.scale_row.solve(a.scale_row.to_dense())
+    k_inv_s_x = b.scale_column.solve(a.scale_column.to_dense())
+
+    mt = b.mean() - a.mean()
+    transpose = transpose_bijector.Transpose(rightmost_transposed_ndims=2)
```
SiegeLordEx, line 313: Why not use tf.linalg.matrix_transpose? It should just transpose the two rightmost dimensions, just like this.

tensorflow_probability/python/distributions/matrix_normal_linear_operator.py
```diff
@@ -259,6 +261,86 @@ def _parameter_control_dependencies(self, is_init):
...
+    b: Instance of `MatrixNormalLinearOperator`.
+    name: (optional) name to use for created ops. Default "kl_mn".
+
+  Returns:
+    Batchwise `KL(a || b)`.
+  """
+  def squared_frobenius_norm(x):
+    """Helper to make KL calculation slightly more readable."""
+    # http://mathworld.wolfram.com/FrobeniusNorm.html
+    # The gradient of KL[p,q] is not defined when p==q. The culprit is
```
SiegeLordEx, line 290: I wouldn't bother commenting this, since you would never want to do tf.square(tf.norm(...)) anyway.

tensorflow_probability/python/distributions/matrix_normal_linear_operator.py
```diff
@@ -259,6 +261,86 @@ def _parameter_control_dependencies(self, is_init):
...
+    # tf.norm, i.e., we cannot use the commented out code.
+    # return tf.square(tf.norm(x, ord="fro", axis=[-2, -1]))
+    return tf.reduce_sum(tf.square(x), axis=[-2, -1])
+
+  with tf.name_scope(name or "kl_mn"):
```
SiegeLordEx, line 295: Let's write this out as kl_matrix_normal_matrix_normal (following various other patterns like this)

SiegeLordEx (changes requested):
This looks good, thanks! Just a few small comments.

tensorflow_probability/python/distributions/matrix_normal_linear_operator.py
fdtomasi: Good point, updated!

tensorflow_probability/python/distributions/matrix_normal_linear_operator.py
fdtomasi: Yes it was a leftover from the code I copied from other parts, gone now

SiegeLordEx (approved):

PR #1658, opened 2022-11-30
Add Two-Piece Student's t-distribution

leandrolcampos:
This PR provides an implementation for the Two-Piece Student's t-distribution, also called the Skewed Student distribution [Fernández and Steel (1998)][1] and the Skew T Type 3 (ST3) distribution [(Rigby et al., 2019, Section 18.4.14, p414)][2].

This distribution generalizes the Student's t-distribution with an additional shape parameter. Under the general formulation proposed by [Fernández and Steel (1998)][1], it is parameterized by degree of freedom `df`, location `loc`, scale `scale`, and skewness `skewness`. If `skewness` is above one, the distribution becomes right-skewed (or positively skewed). If `skewness` is greater than zero and less than one, the distribution becomes left-skewed (or negatively skewed). The Student's t-distribution is retrieved when `skewness` is equal to one.

This PR, including the test suite, was inspired by the available implementation of the Two-Piece Normal distribution. It also includes commits that:
- Preserves dtype in `tfp.math.log1psquare`, `tfp.math.lbeta`, `tfp.math.log_gamma_correction`, and `tfp.math.log_gamma_difference`.
- Improves docstrings and parameter name in `TwoPieceNormal`.
- Adds survival function to `StudentT`.

[1] Carmen Fernández and Mark F. J. Steel. On Bayesian modeling of fat tails and skewness. _Journal of the American Statistical Association_, 93(441), 359-371, 1998.

[2] Robert A. Rigby et al. _Distributions for modeling location, scale, and shape: Using GAMLSS in R_. Chapman and Hall/CRC, 2019.

**PS:** Please note that lint errors are related to old code in `hypothesis_testlib.py`. If you want, I can try to fix them.


SiegeLordEx:
Looks great, thanks! Let me pass this by our internal reviews.

leandrolcampos:
Hi, @SiegeLordEx! Many thanks for your time reviewing this PR.

tensorflow_probability/python/distributions/two_piece_student_t.py
```diff
@@ -0,0 +1,847 @@
...
+         Implicit Reparameterization Gradients. In _Advances in Neural
+         Information Processing Systems_, 31, 2018.
+         https://arxiv.org/abs/1805.08498
+  """
+  sample_shape = ps.convert_to_shape_tensor(sample_shape, dtype_hint=tf.int32)
+  df = tf.convert_to_tensor(df)
+  skewness = tf.convert_to_tensor(skewness)
+  seed = samplers.sanitize_seed(seed, salt='two_piece_student_t')
+
+  return _two_piece_student_t_sample_with_gradient(
```
SiegeLordEx, line 846: This came up in the internal review: could we instead sample from a two-piece normal and divide by a gamma sample? The motivation is to avoid the custom gradient implementation.

leandrolcampos: Hi @SiegeLordEx,

Excellent question! And it seems you're right. In the following notebooks, I implemented this new sampler and compared the theoretical and empirical means, standard deviations, and densities:
- [Right-Skewed Two-Piece Student's t-distribution](https://gist.github.com/leandrolcampos/450e664e365f96486bc815f62d310ed4)
- [Left-Skewed Two-Piece Student's t-distribution](https://gist.github.com/leandrolcampos/02293c9b3e0054fcda658026e78f8b15)
- [Symmetric Two-Piece Student's t-distribution](https://gist.github.com/leandrolcampos/1f815119d87bbd1b673311af19bdf40a)

For now, I have no idea about which strategy to use to mathematically prove the equivalence between the two samplers. Some advice? And would such proof be necessary?

Intuitively, this equivalence makes sense. Indeed, the Two-Piece Student's t-distribution results from joining, at the location, the corresponding halves of two Student's t-distributions with same degree of freedom and location but different scales. Each of these Student's t-distributions corresponds to a Normal distribution that can be used to generate samples from the former. The two corresponding Normal distributions would create the Two-Piece Normal distribution that we can use to generate samples from the original Two-Piece Student's t-distribution.

I inspected four R packages that contain samplers from this distribution (or from one of its equivalent alternative parameterizations): [gamlss.dist](https://rdrr.io/cran/gamlss.dist/src/R/ST3.R#sym-rST3), [skewt](https://rdrr.io/cran/skewt/src/R/skewt.R#sym-rskt), [fGarch](https://github.com/cran/fGarch/blob/7d85dd50aab0ed4654f7b39c4b810f75df4f684a/R/dist-sstd.R#L298), and [rugarch](https://github.com/cran/rugarch/blob/c9474fc9defb3854233230b55f6a9b8f3a860e3a/src/distributions.c#L420). None of them implement the idea of this new sampler. The first two packages generate samples using the quantile function. The other two generate samples using the Student's t-distribution sampler.

**Should I update this PR, implementing this new sampler?**

SiegeLordEx: I don't think a proof is necessary, you can see it's right by inspecting the implementation of two piece normal and two piece student t. And yeah, I think if you could implement it that'd be good, and simplify the code significantly since we won't need to maintain the custom gradient.

leandrolcampos:
Hi @SiegeLordEx,

Please let me know if there is anything else I need to do.

All the best,

SiegeLordEx:
There was a bit of an issue trying to get the reparameterization gradient tests passing, we opted to simplify it and really crank up the number of samples taken. I suspect the issue is with the heavy tails of the distribution which makes the variance of the results rather high.

leandrolcampos:
Hi @SiegeLordEx and @srvasude,

Many thanks to you and the TFP team for reviewing this and other PRs throughout 2022. It is a pleasure to contribute to TFP and learn from you!

Happy holidays and happy new year! All the best,

PR #1718, opened 2023-05-08
Implementation of mean in categorical

fotisdr:
> NotImplementedError: mean is not implemented: Categorical

Defined a _mean method for implementing the mean of the Categorical distribution, following a previous PR (https://github.com/tensorflow/probability/pull/1411).

### DIFF
--- a/tensorflow_probability/python/distributions/categorical.py
+++ b/tensorflow_probability/python/distributions/categorical.py
@@ -30,6 +30,7 @@ from tensorflow_probability.python.internal import reparameterization
 from tensorflow_probability.python.internal import samplers
 from tensorflow_probability.python.internal import tensor_util
 from tensorflow_probability.python.internal import tensorshape_util
+from tensorflow_probability.python.math import reduce_weighted_logsumexp
 
 
 def _broadcast_cat_event_and_params(event, params, base_dtype):
--- a/tensorflow_probability/python/distributions/categorical.py
+++ b/tensorflow_probability/python/distributions/categorical.py
@@ -333,6 +333,10 @@ class Categorical(
         _mul_exp(log_probs, log_probs),
         axis=-1)
 
+  def _mean(self):
+    probs = self.probs_parameter()
+    return tf.reduce_sum(tf.range(self._num_categories(probs),dtype=probs.dtype) * probs, axis=-1) / tf.reduce_sum(probs, axis=-1)
+  
   def _mode(self):
     x = self._probs if self._logits is None else self._logits
     mode = tf.cast(tf.argmax(x, axis=-1), self.dtype)
--- a/tensorflow_probability/python/distributions/categorical.py
+++ b/tensorflow_probability/python/distributions/categorical.py
@@ -334,8 +334,11 @@ class Categorical(
         axis=-1)
 
   def _mean(self):
-    probs = self.probs_parameter()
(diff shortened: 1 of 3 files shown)


tensorflow_probability/python/distributions/categorical.py
```diff
@@ -333,6 +333,10 @@ def _entropy(self):
         _mul_exp(log_probs, log_probs),
         axis=-1)
 
+  def _mean(self):
+    probs = self.probs_parameter()
+    return tf.reduce_sum(tf.range(self._num_categories(probs),dtype=probs.dtype) * probs, axis=-1) / tf.reduce_sum(probs, axis=-1)
```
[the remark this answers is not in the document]
csuter: sorry, meant [weighted_logsumexp](https://www.tensorflow.org/probability/api_docs/python/tfp/math/reduce_weighted_logsumexp)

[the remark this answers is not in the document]
fotisdr: Should we always expect that the provided probs sum up to 1 for this method, and thus add an assertion that this is the case before computing the mean? Because other methods in the Categorical work without the sum of the probs necessarily being 1 (with validate_args=False).

tensorflow_probability/python/distributions/categorical.py
```diff
@@ -30,6 +30,7 @@
 from tensorflow_probability.python.internal import samplers
 from tensorflow_probability.python.internal import tensor_util
 from tensorflow_probability.python.internal import tensorshape_util
+from tensorflow_probability.python.math import reduce_weighted_logsumexp
```
csuter, line 33: Please import as

```
from tensorflow_probability.python.math.generic import reduce_weighted_logsumexp
```

and add this to the "categorical" `deps` list in the adjacent BUILD file:

"//tensorflow_probability/python/math:generic"

tensorflow_probability/python/distributions/categorical.py
```diff
@@ -333,6 +334,13 @@ def _entropy(self):
         _mul_exp(log_probs, log_probs),
         axis=-1)
 
+  def _mean(self):
+    #probs = self.probs_parameter()
```
csuter, line 338: new impl looks good, thanks! please

1) remove the commented out lines
2) add a test in categorical_test.py. you can look at other categorical tests and maybe normal_test.py for a hint of how these should look. think of edge cases, like some zero prob categories, etc.

tensorflow_probability/python/distributions/categorical.py
```diff
@@ -333,6 +334,13 @@ def _entropy(self):
...
+    #return tf.reduce_sum(tf.range(self._num_categories(probs),dtype=probs.dtype) * probs, axis=-1) / tf.reduce_sum(probs, axis=-1)
+    # Implement with logits to improve numerical stability
+    logits = self.logits_parameter()
+    return tf.math.exp(reduce_weighted_logsumexp(logits,w=tf.range(self._num_categories(logits),dtype=logits.dtype),axis=-1))
```
csuter, line 342: please format as

```
return tf.math.exp(
    reduce_weighted_logsumexp(
        logits,
        w=tf.range(self._num_categories(logits), dtype=logits.dtype),
        axis=-1))
```

csuter:
A few more comments

csuter (changes requested):

SiegeLordEx:
This got implemented in b14ae1d79de4a52d834a3ba2dc88f7e5d849e6c7.

PR #1724, opened 2023-05-16
Add requires-python metadata

### LINKED ISSUE #1723: Lack of requires-python causes tfp v0.20.0 to break all Python 3.7 and older
Hi. As TensorFlow Probability doesn't have `requires-python` metadata (c.f. https://peps.python.org/pep-0621/#requires-python) the recent [`v0.20.0` release](https://github.com/tensorflow/probability/releases/tag/v0.20.0) has broken installation for Python 3.7 and older for everyone (c.f. https://github.com/tensorflow/probability/commit/e81e7e8af77280aabb0db4e9f5826673fe756346). This is because [TensorFlow `v2.12.0`](https://pypi.org/project/tensorflow/2.12.0/) is the first TensorFlow release to require Python 3.8+. So requiring 

https://github.com/tensorflow/probability/blob/e81e7e8af77280aabb0db4e9f5826673fe756346/tensorflow_probability/python/__init__.py#L54

without properly communicating that Python 3.8+ is now explicitly required by TensorFlow probability means that any Python 3.7 (or older) that was able to install and use TensorFlow `v2.11.0` and previous versions of TensorFlow Probability now gets a **broken** install of `tensorflow` `v2.11.0` and `tensorflow-probability` `v0.20.0` which produce a runtime `ImportError`

```
This version of TensorFlow Probability requires TensorFlow version >= 2.12; Detected an installation of version 2.11.0. Please upgrade TensorFlow to proceed.
```

(Example: https://github.com/scikit-hep/pyhf/pull/2202)

Including `requires-python` metadata would have properly guarded against this behavior. TensorFlow Probability `v0.20.0` should be yanked from PyPI and a `v0.20.1` should be cut that includes PR #1725 (backport of PR #1724).

These patterns have been discussed in many places previously and in great detail in @henryiii's blog post "[Should You Use Upper Bound Version Constraints?](https://iscinumpy.dev/post/bound-version-constraints/)" (c.f. https://iscinumpy.dev/post/bound-version-constraints/#pinning-the-python-version-is-special for somewhat similar ideas).

cc @jburnim who made the release.

matthewfeickert:
Resolves #1723

* Add requires-python metadata through the addition of setuptools's python_requires in setup.py.
   - c.f. https://peps.python.org/pep-0621/#requires-python

* The addition of requires-python is to provide guards to keep older CPython versions from installing releases that could contain unrunnable code.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -70,6 +70,7 @@ setup(
     url='http://github.com/tensorflow/probability',
     license='Apache 2.0',
     packages=find_packages(),
+    python_requires = ">=3.8",
     install_requires=REQUIRED_PACKAGES,
     # Add in any packaged data.
     include_package_data=True,


matthewfeickert:
This is ready for review.

PR #1726, opened 2023-06-01
Correct a misleading comment in linear_gaussian_ssm.py

chenhaotian:
The equation in updating the posterior variance is called "Joseph form" instead of "Jordan form". Jordan form has nothing to do with this equation, reading the comment causes quite a bit of confusion.


SiegeLordEx:
Thanks for spotting this!

Issue #1754, opened 2023-10-02
BFGS initial Hessian scaling makes it impossible to call minimization iteratively.


gmjw:
As suggested by @m0nzderr in his comment on [this commit](https://github.com/tensorflow/probability/commit/1468a8e92edd7def98dc49f8faa7c27f9078052c) by @srvasude, the current setup for scaling the initial Hessian estimate when running BFGS makes it impossible to iteratively call `bfgs_minimize` in a loop, passing in a previous inverse Hessian estimate to the next iteration. This is useful for approximate filtering algorithms where estimates from one iteration are passed in as inputs to the next.

If the initial Hessian scaling could be made optional, this issue would be fixed.

gmjw:
I have a "solution" which involves adding an extra attribute to `BfgsOptimizerResults` and an extra argument to `bfgs.minimize`. 

I'm not sure if that is ideal/desirable but at least it's a starting point for discussion.

I'll aim to put up a PR fairly soon.

PR #1755, opened 2023-10-02
Allow for optional initial inverse Hessian scaling in bfgs.minimize

gmjw:
This aims to address issue #1754.

As stated in that issue, previous code did not allow for clean iterative use of the BFGS algorithm. I would hope that this setup does.

I've tried to show in the added test that the option is having its intended effect.


gmjw:
Hi there - just to say this is my first PR for this repo, I'm not a regular TF contibutor, please be gentle!

PR #1757, opened 2023-10-02
Allow optional initial inverse Hessian scaling in bfgs.minimize

gmjw:
This aims to address issue #1754.

As stated in that issue, previous code did not allow for clean iterative use of the BFGS algorithm. I believe this setup does, by making the scaling on the first iteration optional.


gmjw:
This is my first PR to Python files in this repo. Any advice on style/setup would be much appreciated.

I've tried to show in the added test that the option is having its intended effect - allowing us to choose whether to apply initial inverse Hessian scaling or not. And the test also shows that the scaling is working as intended: adjusting the early behaviour of the fit without changing the eventual maximum identified.

brianwa84 (approved):
i'll merge shortly

PR #1766, opened 2023-11-04
Quick fix.

sophiagu:
typo

### DIFF
--- a/tensorflow_probability/python/bijectors/cumsum.py
+++ b/tensorflow_probability/python/bijectors/cumsum.py
@@ -117,5 +117,5 @@ class Cumsum(bijector.AutoCompositeTensorBijector):
     return tf.constant(0., x.dtype)
 
   @property
-  def _compposite_tensor_shape_params(self):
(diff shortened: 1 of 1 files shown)


mihaimaruseac (changes requested):
Thank you for the PR.

You will need to also change all places that call this property. Otherwise, code breaks.

sophiagu:
Let me know if I missed something specifically? From my console, I can only see that all 7 tests have passed. I also did a search within the repo but didn't see other occurrences of this property; it is a private property.

emilyfertig:
Thanks for the fix! With the typo, `tfb.Cumsum._composite_tensor_shape_params` would have called the base class's method, which just returns an empty tuple. It makes sense that the overriding method with the typo wouldn't have been called elsewhere. (Ideally our bijector test harness would catch this, since it's a bug in the JAX backend/less efficient in TF to treat `axis` as a dynamic value, but it currently doesn't). I confirmed that tests pass internally so I'll merge this.

PR #1775, opened 2023-12-10
SMC2

aleslamitz:
Added SMC2 from https://arxiv.org/abs/1101.1528


tensorflow_probability/python/experimental/mcmc/particle_filter.py
```diff
@@ -670,6 +1122,8 @@ def _compute_observation_log_weights(step,
     observation = tf.nest.map_structure(
         lambda x, step=step: tf.gather(x, observation_idx), observations)
 
+    if particles_dim == 1:
+      observation = tf.expand_dims(observation, axis=0)
```
jburnim, line 1126: I don't think this added code should be necessary.  The lines right below this should already handle any necessary `expand_dims` for `particles_dim`.

tensorflow_probability/python/experimental/mcmc/particle_filter.py
```diff
@@ -435,13 +472,410 @@ def seeded_one_step(seed_state_results, _):
...
+        inner_initial_state_proposal,
+        num_transitions_per_observation,
+        inner_resample_fn,
+        inner_resample_criterion_fn,
+        outer_rejuvenation_criterion_fn,
+        unbiased_gradients,
+        parameter_proposal_kernel,
+        num_inner_particles,
+        num_outer_particles,
+        extra_fn
```
jburnim, line 633: I think we can remove this `extra_fn` argument, too.

For now, SMC^2 can just use the `extra` field of the outer `WeightedParticles` for storing for storing the random seed of the inner filter results.

jburnim (commented):
This is looking great, thanks!

I'm still reviewing, but I left some initial comments.

tensorflow_probability/python/experimental/mcmc/particle_filter.py
```diff
@@ -435,13 +472,410 @@ def seeded_one_step(seed_state_results, _):
...
+        )
+
+        inner_weighted_particles = smc_kernel.WeightedParticles(
+            particles=inner_weighted_particles_particles,
+            log_weights=inner_weighted_particles_log_weights,
+            extra=inner_weighted_particles.extra
+        )
+
+        filter_results = tf.nest.map_structure(
+            lambda a, b: where_fn(accept, a, b,
```
jburnim, line 834: We should be able to use `mcmc_util.choose` instead of this custom `where_fn`.

I think the errors you were seeing (which led to defining `where_fn`) were because the `steps` and `seed` fields in `SequentialMonteCarloResults` do not have the same batch shape as the other fields.  It should work to do something like:
```
(parent_indices,
 incremental_log_marginal_likelihood,
 accumulated_log_marginal_likelihood) = mcmc_util.choose(..., ...old fields..., new fields...)
```

I think it is not currently possible to set the seed correctly here, so taking the seed from either `filter_results` or `rej_filter_results` is okay.  (In a later cleanup, we could consider adding batch shape to `seed` to fix this.)

tensorflow_probability/python/experimental/mcmc/particle_filter.py
aleslamitz: I think it is necessary due to the dimensions mismatch

jburnim: With these two lines included, `ParticleFilterTestFloat32.test_batch_of_filters_particles_dim_1` fails when I run it.  If I delete these two lines, then that tests passes.

tensorflow_probability/python/experimental/mcmc/particle_filter_test.py
```diff
@@ -734,6 +735,207 @@ def marginal_log_likelihood(level_scale, noise_scale):
...
+        num_outer_particles=num_outer_particles,
+        num_inner_particles=num_inner_particles,
+        outer_rejuvenation_criterion_fn=rejuvenation_criterion,
+        inner_transition_fn=lambda params: (
+            lambda _, state: independent.Independent(
+                particle_dynamics(params, _, state), 1)
+        ),
+        inner_observation_fn=lambda params: (
+            lambda _, state: independent.Independent(
+                normal.Normal(state, 2.), 1)
```
jburnim, line 780: It seems like the shapes in this test might not all line up.  In particular, it looks like the observation distribution has event_shape [2], but the observation at a single time step is scalar (because observations has shape `[30]`).

It looks like the plan is for the inner particle filter to have a latent space of shape [2], and for both dimensions of the latent space to be observed at a time step (with some noise).  And for there to be a single scalar outer parameter. 
 In that case, I would expect some of the other shapes to look like (given there are 3 outer particles and 7 inner particles):

 - inner particles: shape [3, 7, 2]

 - outer particles/params: shape [3]

 - observations: shape [30, 2]  (for 30 steps)

 - observation_fn: (array[3], array[3, 7, 2]) -> distribution with batch_shape [3, 7] and event_shape [2]

 - inner_transition_fn: (array[3], array[3, 7, 2]) -> distribution with batch shape [3, 7] and event_shape [2]
    * (In the current code, looks like this distribution currently has event_shape [].)

 -  inner_initial_state_prior: array[3] -> distribution with batch shape [3] and event_shape [2] -- and the SMC2 code should take care of drawing enough samples and reshaping to get an initial state of shape [3, 7, 2].
    * (In the current code, it looks like this function returns a distribution with batch shape [7].)


(This should make it possible for this test to pass without the "if particles_dim == 1:" block in `_compute_observation_log_weights`.)

tensorflow_probability/python/experimental/mcmc/particle_filter_test.py
```diff
@@ -734,6 +735,210 @@ def marginal_log_likelihood(level_scale, noise_scale):
...
 
+  def test_smc_squared_rejuvenation_parameters(self):
+    def particle_dynamics(params, _, previous_state):
+      reshaped_params = tf.reshape(params,
+                                   [params.shape[0]] +
+                                   [1] * (previous_state.shape.rank - 1))
+      broadcasted_params = tf.broadcast_to(reshaped_params,
+                                           previous_state.shape)
+      reshaped_dist = independent.Independent(
+          normal.Normal(previous_state + broadcasted_params + 1, 0.1),
```
jburnim, line 746: `params` is has shape `[num_outer_particles]` and `previous_state` has shape `[num_outer_particles, num_inner_particles, 2]`, so here it should be sufficient to just do:
```
previous_state + params[..., tf.newaxis, tf.newaxis] + 1
```

tensorflow_probability/python/experimental/mcmc/particle_filter_test.py
```diff
@@ -734,6 +735,210 @@ def marginal_log_likelihood(level_scale, noise_scale):
...
+      )
+      return reshaped_dist
+
+    def rejuvenation_criterion(step, state):
+      # Rejuvenation every 2 steps
+      cond = tf.logical_and(
+          tf.equal(tf.math.mod(step, tf.constant(2)), tf.constant(0)),
+          tf.not_equal(state.extra[0], tf.constant(0))
+      )
+      return tf.cond(cond, lambda: tf.constant(True),
```
jburnim, line 757: It should work to just return `cond`, instead of using `tf.cond(cond, ...)`.

tensorflow_probability/python/experimental/mcmc/particle_filter_test.py
```diff
@@ -238,23 +236,14 @@ def observation_fn(_, state):
...
+                         num_particles,
+                         batch_shape[1]])
     self.assertAllEqual(incremental_log_marginal_likelihoods.shape,
                         [num_timesteps] + batch_shape)
 
-    self.assertAllClose(
-        self.evaluate(
-            tf.reduce_sum(tf.exp(log_weights) *
-                          particles['position'], axis=2)),
-        observed_positions,
```
jburnim, line 248: I would hope this assertion would pass if we changed this line to `observed_positions[..., tf.newaxis, tf.newaxis]`.

But if that didn't work, we could also do: `tf.broadcast_to(observation_positions, [num_timesteps] + batch_shape)`.

PR #1828, opened 2024-07-30
Field name checks for StructTuple

chrism0dwk:
This PR introduces a new check for field names in a StructTuple, preventing any that over-ride existing members of the tuple base class.  It is intended to trap a gotcha in JointDistributionCoroutine if a user names a variable "index" or "count".

### DIFF
--- a/tensorflow_probability/python/internal/structural_tuple.py
+++ b/tensorflow_probability/python/internal/structural_tuple.py
@@ -51,6 +51,9 @@ def _validate_field_names(field_names):
     if name.startswith('_'):
       raise ValueError(
           'Field names cannot start with an underscore: {}'.format(name))
+    if name in dir(tuple):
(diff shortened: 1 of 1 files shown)


tensorflow_probability/python/internal/structural_tuple.py
```diff
@@ -51,6 +51,9 @@ def _validate_field_names(field_names):
     if name.startswith('_'):
```
ColCarroll, line 51: i guess this will prevent `name="__gt__"`

ColCarroll (approved):
thanks, chris! i'll try to pull this in soon.

ColCarroll:
thanks again, @chrism0dwk -- ended up using the fix @SiegeLordEx suggested, and used your particular case as a unit test.

Issue #1998, opened 2025-04-21
Adding Dirichlet to inference-gym


reubenharry:
I'm planning to submit a PR to add a Dirichlet distribution (variable dimension, variable concentration) to inference-gym, but first wanted to open an issue to see if the developers would be happy to have this distribution in inference-gym. In the readme, possible reasons to add a model are given; from these, the reasons for adding a Dirichlet are:

- high dimensionality (yes, a nice distribution to test how statistical efficiency of a sampler scales with dimension)
- poor or pathological conditioning: (yes, for low concentration number)

SiegeLordEx:
That sounds good to me!

PR #1999, opened 2025-04-22
Add dirichlet distribution

reubenharry:
Implements issue #1998.

### DIFF
--- a/spinoffs/inference_gym/inference_gym/targets/dirichlet.py
+++ b/spinoffs/inference_gym/inference_gym/targets/dirichlet.py
@@ -114,13 +114,3 @@ class Dirichlet(model.Model):
       samples: a `Tensor` with prepended dimensions `sample_shape`.
     """
     return self.bij.inverse(self._dirichlet.sample(sample_shape, seed=seed, name=name))
-  
-if __name__ == '__main__':
-  # Debug test for Dirichlet model
-    import tensorflow_probability.python.internal.test_util as tfp_test_util
-    dims = 3
-    model = Dirichlet(
-        concentration_vector=tf.ones(dims),
-    )
-    # set seed to 1
-    print(model.sample(seed=1))
--- a/spinoffs/inference_gym/inference_gym/targets/dirichlet.py
+++ b/spinoffs/inference_gym/inference_gym/targets/dirichlet.py
@@ -45,8 +45,7 @@ class Dirichlet(model.Model):
 
   def __init__(
       self,
-      ndims=100,
-      concentration_vector=np.ones(100),
+      concentration_vector=tf.ones(100)+100,
       dtype=tf.float32,
       name='dirichlet',
       pretty_name='Dirichlet',
@@ -54,7 +53,6 @@ class Dirichlet(model.Model):
     """Construct the Dirichlet.
 
     Args:
-      ndims: Python `int`. Dimensionality of the Dirichlet (in unconstrained space, so e.g. a Beta distribution is a 2-dimensional Dirichlet).
       dtype: Dtype to use for floating point quantities.
       name: Python `str` name prefixed to Ops created by this class.
       pretty_name: A Python `str`. The pretty name of this model.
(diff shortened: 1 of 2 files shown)


reubenharry:
I've marked this as a draft because there are currently no tests, but suggestions are welcome.

spinoffs/inference_gym/inference_gym/targets/dirichlet.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Args:
+      sample_shape: 0D or 1D `int32` `Tensor`. Shape of the generated samples.
+      seed: Python integer or `tfp.util.SeedStream` instance, for seeding PRNG.
+      name: Name to give to the prefix the generated ops.
+
+    Returns:
+      samples: a `Tensor` with prepended dimensions `sample_shape`.
+    """
+    return self.bij.inverse(self._dirichlet.sample(sample_shape, seed=seed, name=name))
```
SiegeLordEx, line 116: This sampling statement doesn't correspond to the log-prob. You don't need to call `self.bij.inverse` (and therefore, there's no need to have a self.bij in the first place)

spinoffs/inference_gym/inference_gym/targets/dirichlet.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Args:
+      dtype: Dtype to use for floating point quantities.
+      name: Python `str` name prefixed to Ops created by this class.
+      pretty_name: A Python `str`. The pretty name of this model.
+    """
+    
+    dirichlet = tfp.distributions.Dirichlet(
+        concentration_vector,
+        validate_args=False,
```
SiegeLordEx, line 63: I'd leave these three args at their defaults.

spinoffs/inference_gym/inference_gym/targets/dirichlet.py
```diff
@@ -0,0 +1,116 @@
...
+
+  
+
+  Note that this function produces reproducible targets, i.e. the constructor
+  `seed` argument always needs to be non-`None`.
+  """
+
+  def __init__(
+      self,
+      concentration_vector=tf.ones(100)+100,
```
SiegeLordEx, line 48: This should be `np.ones`. Both TensorFlow and JAX really don't like when you call them in the global context.

spinoffs/inference_gym/inference_gym/targets/dirichlet.py
```diff
@@ -0,0 +1,116 @@
...
+    'Dirichlet',
+]
+
+
+class Dirichlet(model.Model):
+  """Creates a Dirichlet.
+
+  
+
+  Note that this function produces reproducible targets, i.e. the constructor
```
SiegeLordEx, line 42: This is a leftover from, I assume, ill-conditioned gaussian.

Instead, this could discuss why this target is interesting (adding the sentence that you put in the issue about it having interesting conditioning.

spinoffs/inference_gym/inference_gym/targets/dirichlet.py
SiegeLordEx, line 48: This arg should be documented in the "Args" section below, also could you add a sentence of so discussing the choice of this default? Like does it produce interesting conditioning etc.

SiegeLordEx (changes requested):
Thanks! It looks mostly good (I added comments); there's many formatting issues, but I'll take care of them on our end when we import change.

spinoffs/inference_gym/inference_gym/targets/dirichlet.py
```diff
@@ -0,0 +1,116 @@
+# Copyright 2020 The TensorFlow Probability Authors.
```
SiegeLordEx, line 1: This should become 2025

reubenharry:
Thanks for the comments - I think all addressed!

A few points:

- I didn't add a test running HMC on the distribution (although I have done that using my own code), mainly because I'm not very familiar with the tensorflow probability sampling code, and I struggled to debug it.
- I'd like to add a sample transformation for `lambda x: x*2`, but haven't yet found the closed form fourth moment of the Dirichlet (although I presume it exists).
- I'm happy to run autoformatting stuff for future PRs of this nature, but wasn't sure what tools you used

SiegeLordEx:
I merged it with some changes.

> * I didn't add a test running HMC on the distribution (although I have done that using my own code), mainly because I'm not very familiar with the tensorflow probability sampling code, and I struggled to debug it.

There was a bug with the HMC testing harness that I had to fix, but otherwise I just followed examples from the other targets.

> * I'm happy to run autoformatting stuff for future PRs of this nature, but wasn't sure what tools you used

Sadly it is not easy to replicate externally, I wouldn't worry about it.

SiegeLordEx:
Oh, I just noticed we forgot to re-export it in `__init__.py`, whoops.

reubenharry:
Oops! Should I add this, or are you planning to?

PR #2025, opened 2026-05-25
Add SkewGeneralizedNormal distribution (generalized normal v2)

inhandan:
Adds the skew generalized normal distribution (the "version 2" / skew variant of the generalized normal), parameterized by location `loc`, scale `scale` (> 0) and shape `peak` (any nonzero real). It is the change-of-variables of a standard normal under
y = (-1/peak) * log(1 - peak * (x - loc) / scale), giving a parameter-dependent half-line support whose mode sits at or near the support boundary (the gap shrinks like exp(-peak**2) as |peak| grows).

Includes log_prob / cdf / log_cdf / survival / quantile / sampling, entropy, mean / variance / mode, and a support-respecting default event-space bijector built as a single Chain (valid for either sign of `peak`). Full unit tests validate against the closed-form v2 formulas and the peak -> 0 Normal limit, since no v2 reference exists in scipy (scipy.stats.gennorm is the symmetric v1). Registered in distributions `__init__`, BUILD, and hypothesis_testlib.

### DIFF
--- a/tensorflow_probability/python/distributions/BUILD
+++ b/tensorflow_probability/python/distributions/BUILD
@@ -2110,6 +2110,29 @@ multi_substrate_py_library(
 )
 
 multi_substrate_py_library(
+    name = "skew_generalized_normal",
+    srcs = ["skew_generalized_normal.py"],
+    deps = [
+        ":distribution",
+        # numpy dep,
+        # tensorflow dep,
+        "//tensorflow_probability/python/bijectors:chain",
+        "//tensorflow_probability/python/bijectors:exp",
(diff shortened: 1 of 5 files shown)


inhandan:
Hi there! This is technically a duplicate of a PR I opened some years ago, and recently modernized on the new TF backend. I signed the CLA way back then, and the failed invocation actually points to my having signed already.

Please let me know what's next in the process?

inhandan:
Original PR is here: https://github.com/tensorflow/probability/pulls?q=is%3Apr+is%3Aopen+inhandan

PR #2026, opened 2026-07-06
Regularize hyp2f1's removable singularity at integer c-a-b

### LINKED ISSUE #1861: hypergeometric series unreliable for simple values
The current implementation of `hyp2f1` outputs wrong results for certain basic values, for example:

```python

hyp2f1 = tfp.math.hypergeometric.hyp2f1_small_argument
H = 1
hyp2f1(1.0, 0.5 - H, H + 1.5, 9/10) # NaN
```

This happens under `tensorflow_probability==0.24.0` and `tensorflow==2.17.1` (e.g. under the current image used by Colab)

The correct result is 
```python
from scipy import special
hyp2f1 = special.hyp2f1
hyp2f1(1.0, 0.5 - H, H + 1.5, 9/10) # 0.7836799547024426
```

### LINKED ISSUE #2001: hypergeometric 2F1 undefined for legitimate inputs
In my research on Gaussian Processes I needed hypergeometric function2F1 for small arguments (less than 1 in absolute value).

I have found that the current implementation of `hyp2f1_small_argument` is incomplete:

To reproduce, run this in Colab and compare against Scipy or Mathematica
```python
hyp2f1 = tfp.math.hypergeometric.hyp2f1_small_argument

H = 1
a = tf.constant(1.0)
b = tf.constant(0.5 - H)
c = tf.constant(H + 1.5)
x = tf.constant(0.9901961)
hyp2f1(a,b,c,x) # nan, should be ~ 0.753603 
```

Tested under `tfp==0.25.0` and `tf==2.18.0` on Google Colab.

Here is the problematic code:
https://github.com/tensorflow/probability/blob/65f265c62bb1e2d15ef3e25104afb245a6d52429/tensorflow_probability/python/math/hypergeometric.py#L524-L681

Sumu004:
`hyp2f1_small_argument` returns `nan` near `z=1` whenever `c - a - b` is exactly an integer (same underlying bug as #1861, different repro values).

The near-1 connection formula needs both `lgamma(d)` and `lgamma(-d)` where `d = c-a-b`; when `d` is an integer, one of the two always lands on a Gamma pole, even though `2F1` itself is finite there — a removable singularity in the formula, not the function. I first tried the closed-form DLMF correction directly, verified it at `m=2`, then found it silently diverged at `m=3` — a transcription error a single check wouldn't have caught. Switched instead to nudging `c`/`d` by `epsilon = sqrt(machine eps)` before evaluating the existing (already-correct) formula, since `2F1` is analytic in `c` away from its own poles.

Checked against `mpmath` at 50-digit precision across `m` from -4 to 4 and 300 randomized parameter sets: worst-case relative error `8.6e-7`, within the expected `O(epsilon)` bound. Also matches #1861's repro exactly.

Fixes #2001
Fixes #1861

### DIFF
--- a/tensorflow_probability/python/math/hypergeometric.py
+++ b/tensorflow_probability/python/math/hypergeometric.py
@@ -431,6 +431,7 @@ def _hyp2f1_z_near_one(a, b, c, z):
   """"Compute 2F1(a, b, c, z) when z is near 1."""
   with tf.name_scope('hyp2f1_z_near_one'):
     dtype = dtype_util.common_dtype([a, b, c, z], tf.float32)
+    numpy_dtype = dtype_util.as_numpy_dtype(dtype)
     a = tf.convert_to_tensor(a, dtype=dtype)
     b = tf.convert_to_tensor(b, dtype=dtype)
     c = tf.convert_to_tensor(c, dtype=dtype)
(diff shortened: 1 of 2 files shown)


maciejskorski:
Thanks for this patch!
Note that there is one more issue that could be possibly fixed
https://github.com/tensorflow/probability/issues/1861

Sumu004:
Thanks for the pointer, good catch, #1861 is actually the same underlying bug as #2001: same degenerate parameters (a=1, b=-0.5, c=2.5, so c-a-b=2 exactly), just a different z value (0.9 instead of 0.9901961). Verified the fix already in this PR resolves #1861's exact repro too (0.783679954702441... vs scipy's 0.7836799547024426), and pushed an explicit test case for it. Will note in the PR description that this fixes both issues.

PR #2032, opened 2026-09-18
Handle JAX_MODE for valid PRNG key generation

MaddipatlaChetan24:
Added handling for JAX_MODE to return a valid JAX PRNG key instead of a plain integer seed.


MaddipatlaChetan24:
please check this and let me know !!

